PR #129, opened 2013-03-30
Add Debian package (buildscript and package)

jpec:
You can rebuild liquiprompt.deb by running the create_deb.sh script.

I'm not very aware of Debian packaging standards so you can improve it !

bazinga_logo


dolmen:
Cc: @aborrero

aborrero:
I really think this belongs to distributions. Apart of it, the script is very bad... I would drop it ASAP.

PR #198, opened 2013-06-10
_lp_temp_sensors : a better use of sed

### LINKED ISSUE #189: Temperature support is slow
See PR #187, do not use grep and sed on the same pipe.

jaesivsm:
closes #189


dolmen:
(This is a followup to #187)

PR #200, opened 2013-06-12
Added runtime displaying

mrks:
This patch adds a field that displays the runtime of the last command. Enable the feature with `LP_ENABLE_RUNTIME=1` in your config. Only runtimes over two seconds are displayed - use `LP_RUNTIME_THRESHOLD` to change this value.

Example:

```
[:~] $ sleep 1
[:~] $ sleep 2                # prompt unchanged because the threshold is set to 2 seconds
[:~] 2s $ sleep 1 && sleep 2  # runtime of 2 seconds displayed
[:~] 3s $                     # runtime of chain correctly displayed
```
- As of now, bash only. It would be great if someone could port it to zsh
- The time displayed is the real time elapsed (wall time).
- As this uses the unix timestamp in seconds, an error of up to one second is possible.
- Doesn't use time(1) but the bash internal timestamp, so no additional processes are spawned and the overhead is minimal
- The runtime of chains (see line 3 in the example above) is correctly measured - this is different from e.g. `time sleep 1 && sleep 2`
- Unlike other solutions, this one does not overwrite $SECONDS

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1200,7 +1200,12 @@ _lp_runtime()
     [[ "$LP_ENABLE_RUNTIME" != 1 ]] && return
     if [[ $LP_RUNTIME_SECONDS -ge $LP_RUNTIME_THRESHOLD ]]
     then
-        echo -ne "${LP_COLOR_RUNTIME}${LP_RUNTIME_SECONDS}s${NO_COL}"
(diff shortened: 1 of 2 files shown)


dolmen:
(Just for the record: this PR has been replaced by #205 that has been merged)

Issue #214, opened 2013-07-26
bug in _lp_temp_sensors


jaesivsm:
As I commented #206 (but commenting in closed issue doesn't seem to be the right way to do things), since the merge of #207 into `develop` the temp feature is broken.
#207 should have been merged to the master branch where the bug still exist but not in the develop branch (where it doesn't exist anymore since 2f7cd163cde7637ada27d9cd66f2553ae22b3982).

I think the right way to do things here is to revert 9d96a82 and to consider merging develop into master or to apply 9d96a82 to the master branch in the meantime.

Issue #206, opened 2013-06-30
_lp_temp_sensors() throws a division by zero error


mainerror:
Basically, _lp_temp_sensors() did not assume that parsing the sensors output might fail thus leaving `count` at 0.

As we all know, division by zero is somewhat problematic so to speak, we should check if count is zero before dividing by it.

PR #223, opened 2013-09-17
fixing #214

jaesivsm:
Reverting #206, should fix any bug concerning temp; even #212 although we don't filter on disabled sensors which returns negative values (we just filter out values that aren't positive, so it should be ok).


dolmen:
(Just for the record: fixes #214)

Issue #238, opened 2013-12-02
LP_DWIN_KERNEL_REL_VER and other errors after updating (OS X 10.9)


star-szr:
Related to #208.

I'm seeing this with every prompt after updating.

```
bash: LP_DWIN_KERNEL_REL_VER: command not found
bash: update_terminal_cwd;: command not found
bash: [[: Now drawing from 'AC Power': syntax error: invalid arithmetic operator (error token is "'AC Power'")
```

Getting this on both my machines, one of them doesn't have the AC Power error though.

Before I pulled from the machine showing three errors, I was on 907f9564522717098345a68c04327e8cf993e095. Not sure about my other machine unfortunately.

10.9 and bash 4.2.45(2)-release on the three error machine.
10.9 and bash 4.2.39(2)-release on the two error machine.

PR #246, opened 2013-12-10
fix bash syntax error in Darwin kernel-detection code, and add version for Mavericks

augmentedfourth:
Solves #238.


dolmen:
(Just for the record: this PR has been replaced by #247)

PR #254, opened 2013-12-19
Feature: show full hostname

berserck:
Added possibility to show domain if host name is shown.

I've added the option LP_DOMAIN to show the domain if the host name is shown.
I also edited the test.sh file to test for new feature.

I'm using the command ```hostname -f``` instead of the ```\H``` (for bash) and ```%M```(for zsh)
because I noticed that the ```\H``` is bash is not the same in all systems (zsh is coherent):

In macos (bash 3.2.48(1)-release) and freeBSD(bash 4.2.45(0)-release) :
```
# hostname -f
host.domain
# export PS1=" \\\h=\h \\\H=\H #"
 \h=host \H=host.domain #
```
In linux: SLC6 (bash: 4.1.2(1)-release), debian(bash:4.2.37(1)-release ) and ubunto(bash: 4.2.25(1)-release) ):

```
# hostname -f
host.domain
# export PS1=" \\\h=\h \\\H=\H #"
 \h=host \H=host #
```

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -24,26 +24,26 @@
 # AUTHORS #
 ###########
 
-# Alex Prengère     <alexprengere@gmail.com>      # untracked git files
-# Anthony Gelibert  <anthony.gelibert@me.com>     # small code fix
+# Alex Prengère     <alexprengere@gmail.com>      # Untracked git files
+# Anthony Gelibert  <anthony.gelibert@me.com>     # Several fix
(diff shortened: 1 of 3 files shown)


dolmen:
FQDN feature from #254 is in merged 695d629dd5cf7109e8892075d4cf7fadd8c17d94.
Thanks for being patient!

Note to self: the changes in the testsuite in this PR would be interesting to merge...

PR #277, opened 2014-02-20
Adding option to use FQDN instead of short hostname

thoraxe:
This pull request adds an option (LP_ENABLE_FQDN) which will set whether or not the full hostname is used. This leverages the prompt's hostname option (%H for bash and %M for ZSH).

There are some reports that the output of these values may be incorrect on some systems / with some shell versions.  I do not account for that in this change.


wylieswanson:
I'd love to have this feature!

dolmen:
Rebased and merged in 695d629dd5cf7109e8892075d4cf7fadd8c17d94.
Thanks for being patient!

Issue #281, opened 2014-03-05
Allow setting date and time format


tiziano88:
Right now the prompt only display the system time, but both bash and zsh support printing formatted time:
bash: `\D{format}`
zsh: `%D{format}`
where `format` is passed to the `strftime` function.
It would be good to have a way to override the default behaviour in the config file, so that if a time format string is specified, liquidprompt would prefer that instead of the default system time.

Rycieos:
I'm moving this to v2.1, since v2.0 is getting large and long already, and this mostly depends on #651, which really should wait until after v2.0.

PR #283, opened 2014-03-05
Allow setting date and time format.

tiziano88:
This is done by setting the LP_TIME_FORMAT variable to the desired format string, according to the
format accepted by strftime(3).
Ref #281


Rycieos:
As part of fixing #651, all shell escape sequences needed to be replaced with normal strings. That means that the way this PR handles this change could not work. However, as part of #665, I implemented a feature quite similar. You can read the docs on the [`LP_TIME_FORMAT` config option here](https://liquidprompt.readthedocs.io/en/latest/config.html#LP_TIME_FORMAT).

Thanks for bringing this to our attention!

PR #299, opened 2014-05-15
Enable support for Software Collections

nzwulfin:
Fedora, RHEL, and CentOS have support for parallel installations of versions of software components such as python, perl, mariadb, etc.  One of the common ways to enable the use of a particular collection is to execute a bash shell from the invocation line.  That invocation looks like: 

```
scl enable python27 mariadb55 bash
```

The software collection then sets an environment variable that is a white space separated list of all enabled collections

```
X_SCLS=python27
```

An example of an enabled software collection in a python virtualenv in a git repo:

```
[nzwulfin:~/oso/srtracker] [python27 mariadb55] [flask] master(+4/-1)* ± 
```

The [software collections website ](https://www.softwarecollections.org/en/) has more information about the implementation of the software.


dolmen:
Replaced by #300.

PR #319, opened 2014-10-15
hide sensors errors (if any)

thefab:
example on a a CentOS release 5.6:
$ sensors
No sensors found!
Make sure you loaded all the kernel drivers you need.
Try sensors-detect to find out which these are.

without this commit, this output is displayed before each prompt
(in the default installation/configuration)


dolmen:
Fixed by b816d59a8c6232557b46ff7e5e7dd239715da57b.

Rycieos:

PR #322, opened 2014-10-28
adding option to use \W for path instead of \w

dschep:
The current shorten path doesn't offer a nice way to simply have just the current directoy. This just uses the native \W to achieve that.


dolmen:
Replaced by #322.

PR #336, opened 2014-12-20
Option to show only current directory

alexprengere:
Replacement for #329.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -74,6 +74,7 @@ if test -n "${BASH_VERSION-}" -a -n "$PS1" ; then
     _LP_MARK_SYMBOL='\$'
     _LP_FIRST_INDEX=0
     _LP_PWD_SYMBOL="\\w"
+    _LP_DIR_SYMBOL="\\W"
     _LP_PERCENT='%' # percent must be escaped on zsh
     # Disable the DEBUG trap used by the RUNTIME feature
     # (in case we are reloading LP in the same shell after disabling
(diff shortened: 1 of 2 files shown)


alexprengere:
Any update for this PR?

alexprengere:
Ping

alexprengere:
Ping

dolmen:
Your branch is not rebase on top of the `develop` branch.

I'm also not happy with the option name (LP_CURRENT_DIR_ONLY), but have nothing better for now.
Also some other users use this feature without needing to change the code. See #323.

jongiddy:
Would this be better as a stand-alone variable, separate to the full/shortened path, added similarly to the code in PR #357 ?

dolmen:
I'm not satisfied with the proposed implementation because it has a runtime cost.

Currently a user can already implement the feature himself from his config:
```
LP_ENABLE_SHORTEN_PATH=0
if $_LP_SHELL_bash; then
    LP_PATH_DEFAULT="\\W"
else
    LP_PATH_DEFAULT="%1~"
fi
```
I'm not telling this solution is officially supported. Just that a simple user hack is possible.
So this is the reason why I've not spent much time implementing my ideal implementation of the feature.

pkkolos:
I think there is no need for this PR. You can show only the current directory by setting the variable `LP_PATH_KEEP=-1` as shown in https://github.com/nojhan/liquidprompt/blob/develop/liquidprompt#L616-L619. Admittedly it'd be good if this option is documented.

alexprengere:
I confirm that `LP_ENABLE_SHORTEN_PATH=1` and `LP_PATH_KEEP=-1` will produce the same results as this PR on `bash` and `zsh`.

I have been using using my own fork with this PR for some time now, because keeping more than the current directory took too much space on the prompt. And since I am a `zsh` user, I have the full path in my right prompt anyway :wink:.

But I agree that documenting the `LP_PATH_KEEP=-1` is a better solution than merging this, so if this is alright with you I am thinking of closing the PR.

PR #339, opened 2014-12-24
Add support for more hostname colours

iKevinY:
I noticed that the hostname colours only support 6 different colours, even on terminals where many more colours are available. I used the following command to check all of the available colours on my `xterm-256color` terminal with two different colour schemes (the default OS X Terminal theme with a white background and a darker one that I use on a regular basis):

``` bash
for i in {1..255}; do tput setaf $i; echo -n "$i "; done
```

Light
Dark

On the white background, there is an issue with colours 226–231 being extremely difficult (or impossible) to read. Therefore, I opted to use `tput colors` to detect the number of colours that are available, and taking 75% of those colours. This has a couple benefits:
- The previous behaviour of the hostname colour algorithm chose a random colour in the interval [1, 6]. This implementation behaves exactly the same when the output of `tput colors` is 8.
- 75% of 256 is 192, and choosing a colour in the interval [1, 192] avoids the use of colours 226–231 on `xterm-256color` terminals.

However, I'm not too certain about how well this solution will work on other systems, since I've only tested it on my OS X machine. Is it reasonable to assume that `tput colors` will be available everywhere? If not, I presume it would be a pretty trivial patch (just falling back to `% 6` if `tput colors` is unavailable).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -15,6 +15,8 @@
 	- Git:
 	    * Also show the number of commits behind the remote (Github #269)
 	- Add variable LP_TTYN: the basename of the terminal (Github #357)
+	- Hostname:
+	    * Support more hostname colours on compatiable terminals.
 	Bug fixes:
 	- Git:
 	    * Fix typo in git work directory detection
@@ -28,7 +30,7 @@
 	- Documentation:
 	    * Many small fixes
 
-	Thanks to Matt Fletcher (@MaffooBristol), Kevin Yap (iKevinY), Sean
+	Thanks to Matt Fletcher (@MaffooBristol), Kevin Yap (@iKevinY), Sean
 	Hussey (@seanhussey), François Schmidts (@jaesisvsm), Morgan Knicely
 	(@morganizeit), Daniel Serodio (@dserodio), Jonathan Giddy
 	(@jongiddy)...
--- a/liquidprompt
+++ b/liquidprompt
@@ -252,11 +252,17 @@ _lp_source_config()
     # NO_COL is special: it will be used at runtime, not just during config loading
     NO_COL="${_LP_OPEN_ESC}${ti_sgr0}${_LP_CLOSE_ESC}"
 
-    # compute the hash of the hostname
-    # and get the corresponding number in [1-6] (red,green,yellow,blue,purple or cyan)
-    # FIXME check portability of cksum and add more formats (bold? 256 colors?)
-    local hash=$(( 1 + $(hostname | cksum | cut -d " " -f 1) % 6 ))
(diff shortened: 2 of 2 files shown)


dolmen:
The major issue is not about missing `tput`. It is more about bad terminfo definitions or broken terminal emulators. For example, on my Ubuntu 14.04 under gnome-terminal `tput colors` says `8` instead of `256` because gnome-terminal says it is an `xterm` (instead of `xterm+256color` or `gnome-256color`).

iKevinY:
I see. Maybe instead, we could perform a series of checks (ie. if "256" is in `$TERM` or [if `$COLORTERM` is `gnome-terminal`](http://askubuntu.com/a/126)) to determine if the terminal *should* be capable of supporting 256 colours, and resort to using the output of `tput colors` if the colour count could not be determined in one of those other ways?

dolmen:
If `$COLORTERM` is set we can assume we have 256 colors.

iKevinY:
Alright, would something like this work?

```sh
local COLORS=$([[ -n "$COLORTERM" ]] && echo 256 || echo $(tput colors))
local HASH_MOD=$(($COLORS * 3/4))
# ...
```

dolmen:
```sh
local -i colors
if [[ -n "$COLORTERM" ]]; then
    colors=256
else
    colors=$(tput colors)
fi
```

iKevinY:
Ah, I didn't know about the `-i` flag. Is using the conditional advantageous to the pseudo-ternary operator? I like the compactness of the ternary operator but if the conditional is better in terms of performance or for compatibility reasons, I'll definitely swap it.

iKevinY:
@dolmen Rebased this PR according to your suggestion.

iKevinY:
@dolmen Resolved a merge conflict that cropped up in `CHANGES`.

Rycieos:
In my opinion, having 170 or so possible colors for hostnames is not a benefit, especially because many of those colors are very close to another color. I know I am a minority, but I am color blind, which makes differentiating colors harder than the average person. The colors for a hostname should invoke a feeling in memory to help remember what host you are currently on. Too close of colors would prevent this.

This patch has no way to disable this feature (other than disabling 256 colors in your terminal, which I do not think is acceptable). There is the normal `LP_ENABLE_SSH_COLORS` option, but this changes that option to always 256 colors, which would change the host colors for all users already used to specific colors for their hosts.

I think **256 color support in the color hostnames is worthy goal**, but it needs to be limited to colors that are still distinct. **I would welcome a patch** that implements that.

The new [`lp_terminal_format()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.1/functions/public.html#lp_terminal_format) function has 256 color support, so it would make generating the colors simple. Picking unique colors would be the hard part.

PR #340, opened 2015-01-05
Use absolute path of grep to fix issues when aliasing grep

stonewareslord:
Fixes issue when aliasing `grep`:
selection_073
Should work on Mac and Linux

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -133,9 +133,9 @@ esac
 
 # Get cpu count
 case "$LP_OS" in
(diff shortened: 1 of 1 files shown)


dolmen:
The request is useful, but not the implementation.
The right way is to call `grep` as `\grep` to avoid aliases.

dolmen:
Please fix your patch and `git push -f`.

stonewareslord:
Fixed.

dolmen:
Cherry-picked on `develop` in 724feebf0bb4ff61916f1fbda421794c25415ca3. Thanks!

PR #342, opened 2015-01-05
Oh My Git! support

stonewareslord:
Support [this](https://github.com/arialdomartini/oh-my-git) project.


Rycieos:
This PR was closed because it was targeting the `develop` branch. I deleted that branch so as to not be confusing: it hasn't been current in years.

I don't think that this kind of prompt support is a good idea. With the new [Liquidprompt themes features](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/theme/custom.html), it might be possible to make a theme to make this kind of integration work better, but I don't want it as part of main Liquidprompt.

If you make a theme that integrates it well, I would look at merging that theme in as one of the included themes.

PR #343, opened 2015-01-05
Added RPS1 for ZSH prompts

stonewareslord:
Uses RPS1 of the user uses zsh


dolmen:
This commit does design changes (moving things from PS1 to RPS1). This is a major design change for zsh users that you didn't even bother to document in the commit and in the pull request.
I can't do anything else that reject it.

PR #357, opened 2015-04-14
Add LP_TTYN variable to display terminal number.

jongiddy:
This is a simple addition, which I find particularly useful added at the start of the terminal titlebar, to make it quicker to switch between terminals with similar titles.


dolmen:
Thanks.
I've cherry-picked, reimplemented, and documented the feature which is now on the `develop` branch.

PR #362, opened 2015-04-29
Adds haskell venv support

epkugelmass:
Adds support for displaying the current Haskell virtualenv immediately
after the (optional) python venv.
[The hsenv project on github](https://github.com/tmhedberg/hsenv)


Rycieos:
Hm, I didn't realize this was targeting the `develop` branch. I deleted that branch so as to not be confusing: it hasn't been current in years.

Please rebase on `master` and update this PR to target `master` instead and I will re-open this.

If you need some guidance around the new architecture, see commit f35d9acfddacc1f7f74174b45cf4c4aa9c84beca for a good example of the things you will need to change.

PR #365, opened 2015-05-10
Put 12 o'clock clock face at array index 0

wprl:
I was experiencing an issue on OS X Yosemite where the clock was always displaying an hour ahead.  I think this fixed the issue, since 12 % 12 should equal 0, so I would expect the clock face for 12 o'clock to be at index 0.  This seems to have fixed the issue for me.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1493,9 +1493,9 @@ _lp_time_analog()
     #
 
     local -a plain
-    plain=(🕐 🕑 🕒 🕓 🕔 🕕 🕖 🕗 🕘 🕙 🕚 🕛 )
+    plain=(🕛 🕐 🕑 🕒 🕓 🕔 🕕 🕖 🕗 🕘 🕙 🕚 )
     local -a half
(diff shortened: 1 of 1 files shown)


dolmen:
Related to change made in #273.

anthonygelibert:
That's strange… this change is OK on my Mac (and the explanations are OK) but I remember that, without the changes of my commit, a wrong hour was displayed. So… all my apologies.

dolmen:
Note to self: check if this is an array indexing issue (bash starts at 0, zsh starts at 1)

anthonygelibert:
My apologies… but I've just checked the last develop version and there is still a problem for me.

On ZSH, local time (France) is 0:42 and i can't obtain an analog clock except if I replace the line 
  
          echo -n "${half[hi]} "

by

          echo -n "${half[hi+1]} "

in _lp_time_analog.

I think it's related to your last comment @dolmen.

dolmen:
This seems to be fixed by a976934a70fb23231bb5f907e869efc2feda4662. Closing.

PR #366, opened 2015-05-11
Cleanups

aborrero:
Hi there!

I've been working in an official Debian package for liquidprompt.
I found these cleanups rather important, please let me know your thoughts.


dolmen:
I will review your suggestions, but I will definitely not merge them as is. I agree that dist/archlinux-aur is just awful.

But I will never (at least until a 2.0 version) move the main `liquidprompt` script out of the root directory. That would break the environments of user who get liquidprompt directly from Git.


Could you give me some pointers to some similar projects (pure shell code) that I could use as best practice demonstrator? Basically, I won't just trust what you think as a good project organization if don't give me some references.

aborrero:
Ok! Then, please just merge the first patch. That can cause us problems in Debian, given we also distribute source code.

dolmen:
> Ok! Then, please just merge the first patch.

It seems I must write it more clearly: **NO!**

> That can cause us problems in Debian, given we also distribute source code.

Tell me what problems it causes and I will consider an alternate solution than yours.

dolmen:
I see in the commits on your aborrero/pkg-liquidprompt some references to critics on your package. Where does this discussion takes place?

aborrero:
What problem do you see with this commit?

https://github.com/aborrero/liquidprompt/commit/821c2b22f742feeb943ea1b5c16105a6b89a3b1a

That's the commit i'm referring to as 'first'. You mentioned your concerns regarding the top-level dirs cleanups, and I'm fine with them.

But what about deleting binary data from the your sources?
Our tarball at Debian contains:
 * the source code from the source tree root.
 * the source code contained at dist/archlinux-aur/liquidprompt-git-20130314-3.src.tar.gz
 * the source code contained at dist/archlinux-aur/liquidprompt-git-20130314-1.src.tar.gz
 * the source code contained at dist/debian/liquidprompt.deb

So, the tarball contains 4 times the same source code, 3 times in binary format.

dolmen:
> What problem do you see with this commit?
>
> aborrero@821c2b2
>
> That's the commit i'm referring to as 'first'.

The problem is that you have a strange definition of 'first'. That commit is the second commit on your branch.

>But what about deleting binary data from the your sources?
>Our tarball at Debian contains:
>
>*    the source code from the source tree root.
>*    the source code contained at dist/archlinux-aur/liquidprompt-git-20130314-3.src.tar.gz
>*    the source code contained at dist/archlinux-aur/liquidprompt-git-20130314-1.src.tar.gz
>*    the source code contained at dist/debian/liquidprompt.deb
>
>So, the tarball contains 4 times the same source code, 3 times in binary format.

No.
* The two tarballs in dit/archlinux-aur do not contain the full liquidprompt source. They contain only the metadata for Arch Linux. Just a quick look at the file sizes would have given you a hint.
* dist/debian/liquidprompt.deb only contains two (outdated) files
```
dpkg-deb -c liquidprompt.deb 
drwxr-xr-x jpec/wheel        0 2013-03-30 17:49 ./
drwxr-xr-x jpec/wheel        0 2013-03-30 18:27 ./etc/
-rw-r--r-- jpec/wheel     3183 2013-03-30 18:27 ./etc/liquidpromptrc
drwxr-xr-x jpec/wheel        0 2013-03-30 17:49 ./usr/
drwxr-xr-x jpec/wheel        0 2013-03-30 18:27 ./usr/bin/
-rwxr-xr-x jpec/wheel    48073 2013-03-30 18:27 ./usr/bin/liquidprompt
```

So your commit does mass deletion of unrelated files. A good reason for me to not merge this commit either.
I've instead committed c32ab5cd4ab5f18566bc870ce559297ad04866fc, 4e9c41338b929bec755abedad3b839851d9cbf64 and cff331632eeb8c1d9c9538a8ba080a99959332a7.

PR #367, opened 2015-05-15
Add install instructions for Homebrew

wprl:


dolmen:
The [homebrew formula for liquidprompt](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/liquidprompt.rb) is not maintained by the liquidprompt core team. It can not be officially endorsed by me. The only supported way to install and upgrade liquidprompt is by using git.

Issue #304, opened 2014-06-04
tmux session on localhost is misinterpreted as telnet session


wimstefan:
And therefore liquidprompt thinks it is a remote connection and shows the hostname in WARNING RED :)

This is the output from `who`:
```
swimmer  pts/12       2014-06-03 12:53 (tmux(15944).%0)
swimmer  pts/13       2014-06-03 12:53 (tmux(15944).%1)
swimmer  pts/14       2014-06-03 12:53 (tmux(15944).%2)
swimmer  pts/15       2014-06-03 12:53 (tmux(15944).%3)
swimmer  pts/16       2014-06-03 12:53 (tmux(15944).%4)
```

flasheater:
I have a similiar issue. The host name is red inside Guake terminal. This is not the case in Terminal.

I confirmed by setting LP_COLOR_TELNET to some other color, that it is in fact the telnet color. 

This may be a Guake issue.

dolmen:
@wimstefan @flasheater @XDex Please try branch `fix/connection-source-tmux` where I have committed 0d7d3fdcf1a892149c0ffbbc580248deefe0be70.

dolmen:
I still need user feedback on the proposed fix...

XDex:
Just pulled the changes from upstream fix/connection-source-tmux - that fixed the issue on my system..

mathieumd:
The branch `fix/connection-source-tmux` fixes the issue on my system too (Tmux 2.0 into Konsole 15.08.0). Before/After:

```
[mathieu@hostname:/opt/liquidprompt] master ± git checkout fix/connection-source-tmux
Basculement sur la branche 'fix/connection-source-tmux'
Votre branche est à jour avec 'origin/fix/connection-source-tmux'.
[mathieu@hostname:/opt/liquidprompt] fix/connection-source-tmux ± . ~/.bashrc
[mathieu:/opt/liquidprompt] fix/connection-source-tmux ±
```

Syllo:
The fix worked for me as well.
Using: 
Tmux 2.0 and who (GNU coreutils) 8.24 on Konsole 15.04

hcgraf:
fixed for me with merge 2703ee03c83595e95fa60fb68ff6d77c042a7ca0

dolmen:
@hcgraf @Syllo @mathieumd @XDex Thanks for your feedback!

daverigby:
[...]

PR #368, opened 2015-05-19
fixed detection of ssh and tmux local sessions

chrigl:
At least new versions of tmux (running v2.0), the detection of ssh and
local sessions did not work proper.

Since some update of tmux, only `SSH_CONNECTION` will be kept when
starting tmux during an SSH session. So I just added this one to
`_lp_connection`.

For local sessions, liquidprompt thought to run via telnet. I think this
is related to the tmux update as well. However, tmux shows this for who
am i:
```
   guest    pts/4        2015-05-19 20:28 (tmux(1938).%3)
```
Which breaks the local detection.
Since `sess_parent` is already known in `_lp_connection`, I use this to
check "this" for being either tmux or screen.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -493,13 +493,13 @@ fi
 
 _lp_connection()
 {
-    if [[ -n "$SSH_CLIENT$SSH2_CLIENT$SSH_TTY" ]] ; then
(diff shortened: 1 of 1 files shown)


dolmen:
This is a fix for #304.

Could you instead test (and review) my branch `fix/connection-source-tmux`?

dolmen:
Everybody on #304 tells me this is fixed. So closing.

Issue #370, opened 2015-06-05
screen corruption with tmux on Debian Jessie


brianmay:
Not sure what is going on here. Is fine without tmux, however if I have tmux enabled, if I try to edit a previous command line I get all sorts of wonderful effects. e.g. the cursor is lined up on the place where I want to insert text, however the new text appears in a different place on the command line, so when I push enter, the command executed is different from the command that was shown on screen.

brianmay:
Actually, not just tmux

brianmay:
I noticed under xterm, the prompt appears as:

```
:~] % [brian:~] %
```

Instead of 

```
[brian:~] %
```

If I push ctrl-R (bound to history-incremental-search-backward), the cursor jumps back to the ':' character, and my search results overwrite the prompt:

```
:~] % [brianecho 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 21 23 24 25 65 66 77 788 99 00 --- hello
```

gnome-terminal instead of xterm seems fine. tmux inside a gnome-terminal breaks also.

dolmen:
Does this happen with the `develop` branch?

dolmen:
Tell me also your tmux version (`tmux -V`) and your terminal (`echo $TERM`).

brianmay:
tmux 1.9. When running under xterm I get TERM=xterm, under tmux I get TERM=screen. Will try the develop branch now.

brianmay:
Under develop, xterm support seems a lot better. Thanks

tmux still has a problem. Under tmux the prompt appears as

```
134[brian:~] %
```

I am not sure where this 134 is coming from. Always seems to be 134. I think that pushes the prompt 3 characters further to the right then expected, which is affecting editing.

brianmay:
Think the problem is somehow related to the value of ``LP_SCREEN_TITLE_CLOSE``:

```zsh
LP_SCREEN_TITLE_CLOSE=${LP_SCREEN_TITLE_CLOSE:-"\033\134"}
```

brianmay:
I think that might be a mistake, I can't see any reference to why \134 is used[1], it should be:

```
LP_SCREEN_TITLE_CLOSE=${LP_SCREEN_TITLE_CLOSE:-"\033"}
```

[1] e.g. not mentioned here:
http://aperiodic.net/screen/title_examples

pkkolos:
[...]

PR #371, opened 2015-06-06
Fix issue #370

pkkolos:
Change screen/tmux title close escape sequence to \033\\
Simplify the code of the title escape sequences


dolmen:
Fixes issue #370.
Merged.

PR #376, opened 2015-06-19
Added support for tab colors based on CD in iTerm

michalzaptk:
Based on colors specified in the TAB COLOR function in the theme file, users can specify tab color changes in iTerm based on the current directory (includes all subdirectories of specified folder)

### DIFF
--- a/liquid.theme
+++ b/liquid.theme
@@ -116,6 +116,32 @@ LP_COLOR_VIRTUALENV="$CYAN"
 # Runtime
 LP_COLOR_RUNTIME="$YELLOW"
 
+# Define folder name and color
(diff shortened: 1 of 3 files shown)


dolmen:
Tab color is an obscure feature of a particular terminal program. I don't want to add this feature to LiquidPrompt. Especially as I don't have a Mac, so I will never be able to properly ensure that the feature still work in the future.

But I will consider adding a more general hook function that would be called when LiquidPrompt detects that the current directory changed.

dolmen:
Not that in zsh the hook is already available:
```zsh
chpwd_functions=(${chpwd_functions[@]} TAB_COLOR)
```

Issue #377, opened 2015-06-24
tmux session on localhost is misinterpreted as telnet session


XDex:
Hi!
I'm having the same issue as #304 on Debian Testing.
Relevant output:
```
⏚ [andrei@andrei-lt-deb:~] $ echo $(who am i | sed -n 's/.*(\(.*\))/\1/p')
9574).%0
⏚ [andrei@andrei-lt-deb:~] $ tmux -V
tmux 2.0
⏚ [andrei@andrei-lt-deb:~] $ who am i
andrei   pts/1        2015-06-24 17:21 (tmux(9574).%0)
```

dolmen:
Do you use the `develop` branch?

XDex:
Nope, should I?

dolmen:
Yes, you should as it includes many fixes, and some are related to tmux. See [the CHANGES on develop](https://github.com/nojhan/liquidprompt/blob/develop/CHANGES) since the last release.

I'm only considering bug reports against the `develop` branch.

dolmen:
Note for example that #370 (which is somewhat related to tmux) has been fixed on the `develop` branch.

XDex:
Just switched to develop. The issue persists..

XDex:
Fixed in PR #383

dolmen:
As I rejected the "fix" in #383, the issue is still open.

dolmen:
Closing because duplicate of #304.

PR #383, opened 2015-07-02
Fix issue #377

XDex:
Corrected detection of local tmux sessions


dolmen:
The problem is in the regexp used when parsing `who am i` output. This is not the right approach to fix the issue.

XDex:
Even if you fix the $sess_src regexp, you'll still have to add an additional if statement condition. My PR fixes the issue with minimal code changes..

dolmen:
I don't need minimal code change. I need a good quality fix that will resist to time and not break later in a different way.

PR #388, opened 2015-08-20
Better proxy detection

xstefanox:
In Unix systems, the proxy environment variables can be uppercase or lowercase, but currently the script checks for _$http_proxy_.
Also, there could be the need to highlight the configuration of an FTP proxy, so I added the check for that variables too.

With this patch, the script will set _$LP_PROXY_ if it finds that at least one of the following variables are not empty:
- _$http_proxy_
- _$HTTP_PROXY_
- _$ftp_proxy_
- _$FTP_PROXY_

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1557,7 +1557,7 @@ _lp_set_prompt()
     LP_TIME="$(_lp_sr "$(_lp_time)")"
 
     # in main prompt: no space
-    if [[ "$LP_ENABLE_PROXY,$http_proxy" = 1,?* ]] ; then
+    if [[ "$LP_ENABLE_PROXY,$http_proxy" = 1,?* || "$LP_ENABLE_PROXY,$HTTP_PROXY" = 1,?* || "$LP_ENABLE_PROXY,$ftp_proxy" = 1,?* || "$LP_ENABLE_PROXY,$FTP_PROXY" = 1,?* ]] ; then
         LP_PROXY="$LP_COLOR_PROXY$LP_MARK_PROXY$NO_COL"
     else
         LP_PROXY=


dolmen:
Yes, environment variables can be upper case or lower case.

But proxy variables are lower case. Point me to something that tells the contrary.

dolmen:
The [curl manual](http://curl.haxx.se/docs/manpage.html#ENVIRONMENT) says:

> The environment variables can be specified in lower case or upper case. The lower case version has precedence. `http_proxy` is an exception as it is only available in lower case.

So there is no reason to support `HTTP_PROXY` in upper case.

xstefanox:
As you can see in the [ArchLinux wiki](https://wiki.archlinux.org/index.php/proxy_settings#Environment_variables), the better solution to correctly configure the proxy on a Unix system seems to be setting both upper and lower case environment variables, because some programs support the first format and other support the latter.

There have been various cases during the development of some proxy-aware software that have been adapted to support both, for example [this](https://github.com/cloudfoundry/cli/issues/213) and [this](https://mail.gnome.org/archives/commits-list/2009-December/msg02597.html), so I think that also Liquid Prompt should check for both formats.

Moreover, following the example found on the ArchLinux wiki, for completeness we should check https and rsync proxy variables too. I could add this to the patch.

dolmen:
No I disagree. Liquidprompt must only support the case of the variable names that is really used by applications. And not give to the user the illusion that he has properly set his environment.

Your references are irrelevant:
- in cloudfoundry/cli#213, the software was failing to support `http_proxy` in lower case. Liquidprompt has always supported it.
- in the gnome-terminal patch you reference, it just shows that gnome-terminal exports both. I don't consider gnome-terminal to be a relevant application regarding the USAGE of the proxy environment variables.

xstefanox:
It is not in dispute that Liquidprompt has always supported the lowercase form, which is correct, but that it is lacking the support for the uppercase form.
The facts are:
1. On Unix systems there is no a standard way of configuring proxy using environment variables.
2. Some programs require the lowercase form, other the uppercase form. Some developers have been more forward-looking and decided to write software compatible with both.

In regards to your objections:
- In [cloudfoundry/cli#213](https://github.com/cloudfoundry/cli/issues/213), the application was not failing: it was supporting only one of the two configuration variables. They decided to support both lowercase and uppercase forms of the proxy environment variables, because both are commonly used.
- Considering Gnome Terminal non-relevant is your personal opinion, not a matter of fact. I think that the behavior of the terminal application provided in the default installation of the most popular linux distributions (Ubuntu) should be taken into account.

Two more examples of popular softwares supporting both forms:
1. Ruby, in its built-in Net::HTTP module: (https://bugs.ruby-lang.org/issues/6546)
2. Composer: (https://getcomposer.org/doc/03-cli.md#http-proxy-or-http-proxy)

dolmen:
>The facts are:
>1. On Unix systems there is no a standard way of configuring proxy using environment variables.

There is no specification, but there is a defacto standard which is the lowercase `http_proxy` as shown in cURL and wget manuals.

>2. Some programs require the lowercase form, other the uppercase form.

I'm still waiting for you to provide a proof of at least a single application of significant traction that *only* supports the uppercase form of `HTTP_PROXY` when making HTTP requests.

>  Some developers have been more forward-looking and decided to write software compatible with both.

I repeat my position written above: Liquidprompt must only support the case of the variable names that is really used by applications. And not give to the user the illusion that he has properly set his environment.

dolmen:
If it wasn't enough, here is a good reason why HTTP client libraries MUST completely ignore `HTTP_PROXY` and only care about `http_proxy`: the 15 years old [httpoxy](https://httpoxy.org/) security vulnerability when such libraries are called from a CGI environment that has not been sanitized.

PR #390, opened 2015-09-02
Fix Fossil tag regexp to work on OS X

nox:


dolmen:
You forgot to give example of what is incorrecly matched.
Remember that I have to check that your fix will not break on other platforms...

nox:
The I modifier doesn't exist in OS X's sed.

```
± sed s/foo/bar/I
sed: 1: "s/foo/bar/I": bad flag in substitute command: 'I'
```

nox:
And the \w class only exist in extended regular expressions, which flag differs between OS X and GNU.

dolmen:
@nox Do you intend to fix this pull request (use of `_LP_SED_EXTENDED`) and update it (git rebase; git push -f)?

dolmen:
@nox Could you check that my fix in 5c3c6a713a33ddcae08cd970b3bd4996258cc9b3 fixes your issue?

nox:
It does, thank you.

milouse:
Your modification introduces a crash on linux platform. I try to fix it in 6e41ef7e1a2a693795666eb34635c20cb75c2942. Could you both give it a try?

dolmen:
@milouse I've just cherry-picked your patch in the develop branch.

PR #391, opened 2015-09-06
Correct _lp_load_color color scale

anthonygelibert:
Like the CPU load is divided by CPUNUM, I think the value is always between 0..100, isn't it?

So, why use a custom scale of 200 to colorize the output of _lp_load_color()?

Maybe, I made a mistake and don't understand something (in this case, drop this PR)…

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1398,7 +1398,7 @@ _lp_load_color()
     local -i load=${tmp:-0}/$_lp_CPUNUM
 
     if (( load > LP_LOAD_THRESHOLD )); then
-        local ret="$(_lp_color_map $load 200)${LP_MARK_LOAD}"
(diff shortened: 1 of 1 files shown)


saiarcot895:
The CPU load can exceed 100%, if you happen to have more processes running (and wanting all of the CPU) than number of cores you have.

dolmen:
@saiarcot895 We already have handling for multiple CPU: https://github.com/nojhan/liquidprompt/blob/master/liquidprompt#L1354

@anthonygelibert I cherry-picked your commit into the `develop` branch. Thanks!

dolmen:
In fact, loadavg can exceed 1 even on a single core. I should not have merged this.
See #455, #499

PR #392, opened 2015-09-20
Revert "LP_TTYN: simplify/optimize the implementation"

elementalvoid:
This reverts commit b67434a61eec3f90bde983d083935e4ca854266d.

The commit message of the reverted commit was about optimizing the TTY detection. Instead of relying on the shell's builtin functionality the commit ran the 'tty' command and cached the results.

Unfortunately this breaks when a tty hasn't been allocated yet. This is what I was seeing both on a local terminals and (some) remote ssh connections:
```
basename: extra operand ‘tty’
Try 'basename --help' for more information.
```
With the commit reverted all is well and this error never appears.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -69,6 +69,7 @@ if test -n "$BASH_VERSION" -a -n "$PS1" ; then
     _LP_USER_SYMBOL="\u"
     _LP_HOST_SYMBOL="\h"
     _LP_TIME_SYMBOL="\t"
+    _LP_TTYN_SYMBOL="\l"
     _LP_MARK_SYMBOL='\$'
     _LP_FIRST_INDEX=0
     _LP_PWD_SYMBOL="\\w"
@@ -86,6 +87,7 @@ elif test -n "$ZSH_VERSION" ; then
     _LP_USER_SYMBOL="%n"
     _LP_HOST_SYMBOL="%m"
     _LP_TIME_SYMBOL="%*"
+    _LP_TTYN_SYMBOL="%l"
     _LP_MARK_SYMBOL='%(!.#.%%)'
     _LP_FIRST_INDEX=1
     _LP_PERCENT='%%'
@@ -452,10 +454,6 @@ fi
 # Default value for LP_PERM when LP_ENABLE_PERM is 0
 LP_PERM=:   # without color
 
-# Same as bash '\l', but inlined as a constant as the value will not change
-# during the shell's life
-LP_TTYN="$(basename $(tty))"
-
 
 # Escape the given strings
 # Must be used for all strings that may comes from remote sources,
@@ -521,6 +519,8 @@ fi
 #################
 # Where are we? #
 #################
+
+LP_TTYN="${_LP_TTYN_SYMBOL}"
 
 _lp_connection()
 {


dolmen:
If you have an issue using LiquidPrompt with sessions not attached to a terminal, it just shows that you are not using LiquidPrompt properly. LiquidPrompt is only for interactive shell sessions, and should be loaded from your `~/.bashrc` / `~/.zshrc` only in that case.
Check that your shell rc matches the [recommandation from the manual](../blob/master/README.md#test-drive-and-installation).

elementalvoid:
It would appear that I'm using it correctly; I'll provide a log below. This may be a Debian specific thing. However, the issue persists even if I follow verbatim the recommendation.

This happens on Bash versions: 4.1.5, 4.2.37, and 4.3.30.

Interactive shells will source liquidprompt as suggested in the docs and non-interactive shells do not. However, as seen below, only interactive shells exhibit this behavior. 

```
[evoid:~/.homesick/repos/liquidprompt] develop ± cat ~/.bashrc.d/liquidprompt
echo '------ DEBUG -------'
echo "\$- = $-"
echo '------ DEBUG -------'

[[ $- = *i* ]] && source ~/.homesick/repos/liquidprompt/liquidprompt
```
Note that the `~/.bashrc.d/liquidprompt` file is sourced during a loop at shell startup.

Plain old interactive bash shell.
```
[evoid:~/.homesick/repos/liquidprompt] develop ± bash
------ DEBUG -------
$- = himBHc
------ DEBUG -------
basename: extra operand ‘tty’
Try 'basename --help' for more information.
```
A non-interactive shell.
```
[evoid:~/.homesick/repos/liquidprompt] develop ± bash -c hostname
evoid-lap
```
An interactive shell over SSH.
```
[evoid:~/.homesick/repos/liquidprompt] develop ± ssh localhost
------ DEBUG -------
$- = himBH
------ DEBUG -------
basename: extra operand ‘tty’
Try 'basename --help' for more information.
```
And a non-interactive shell over SSH.
```
[evoid:~/.homesick/repos/liquidprompt] develop ± ssh localhost hostname
------ DEBUG -------
$- = hBc
------ DEBUG -------
evoid-lap                                                 
```

dolmen:
Thanks for your feedback.
Could you tell me the output of the "tty" command in all broken cases?

dolmen:
I've just pushed f4368670bf258257fece5611a9aad17e88f10b5a on `develop`. Could you check if that fixes your issue?

elementalvoid:
The output of `tty` in all cases is "not a tty". So, while your recent push makes the error silent, the LP_TTYN variable now contains that phrase and it's useless inside of the PS1.

```
[evoid:~/.homesick/repos/liquidprompt] develop ± cat ~/.bashrc.d/liquidprompt
echo '------ DEBUG -------'
echo "\$- = $-"
echo -n "tty -> "
tty
echo '------ DEBUG -------'

[[ $- = *i* ]] && source ~/.homesick/repos/liquidprompt/liquidprompt
```

```
[evoid:~/.homesick/repos/liquidprompt] develop ± bash
------ DEBUG -------
$- = himBH
tty -> not a tty
------ DEBUG -------
[evoid:~/.homesick/repos/liquidprompt] develop ± echo $LP_TTYN 
not a tty
```

Frankly, I'm confused as to what the optimization in the original commit is trying to achieve. It feels like a code cleanliness thing rather than a performance optimization. Is that the case? If so, there's an easier cleanup that could be done.

If not, and it's a performance oriented change then I don't think that it makes any appreciable difference to performance. However, I have no empirical evidence to back that up. When set statically via the tty command at shell invocation the shell forks and execs the tty command; this is slow, relatively speaking, but only happens once per shell. Usually this is not a big deal. However, when using the shell's builtin escape sequence there are no forks and thus no startup penalties. Instead it relies on the, probably highly optimized, shell's compiled binary to parse the escape in the PS line. It has to parse this line whether liquidprompt uses a variable or uses and escape. I would venture to guess that any optimization attempted in liquidprompt in this regard is outweighed by the performance of the shell itself. I fail to see the benefit.

dolmen:
> The output of tty in all cases is "not a tty". So, while your recent push makes the error silent, 

What is your broken operating system?
On a normal operating system, the errors of the `tty` command go to stderr, not stdout so the message should not be in `LP_TTYN`.

> the LP_TTYN variable now contains that phrase and it's useless inside of the PS1.

Yes LP_TTYN is useless with this message. But basically LiquidPrompt is useless if the shell is not attached to a terminal. I still can't figure out how you are running LiquidPrompt in a session not attached to a terminal. This is where you should help me.

> However, when using the shell's builtin escape sequence there are no forks and thus no startup penalties. Instead it relies on the, probably highly optimized, shell's compiled binary to parse the escape in the PS line. It has to parse this line whether liquidprompt uses a variable or uses and escape.

LiquidPrompt is implemented by using a shell hook that changes `PS1` every time `PS1` is displayed. This implies that the time taken by the shell to parse the new value of `PS1` matters. With the optimization the shell will see a bare string (that is just copied) instead of having to do a lookup to get the mapping for the `\l`/`%l` symbol. Yes, this is a micro-optimization.

> I would venture to guess that any optimization attempted in liquidprompt in this regard is outweighed by the performance of the shell itself. I fail to see the benefit.

I fail to see the benefit of reverting the commit because I don't fully understand how you can be in a situation where you are running LiquidPrompt in a session not attached to a terminal.

elementalvoid:
I'll be the first to admit that Debian is broken in more than one way regardless of tty's output decisions. The tty command comes from the [coreutils](https://packages.debian.org/jessie/coreutils) package.

I created a brand new user and ran a completely empty (aside from including liquidprompt) bashrc. In this shell I was unable to reproduce the issue. So, something is up with my personal shell config. I suspect it's a timing issue on shell startup.

For what it's worth, I don't even use the LP_TTYN variable in the PS lines so your commit solves my immediate needs. If you're content with the current fix then by all means close this pull request.

However, I've discovered that b67434a actually changed the behavior of liquidprompt, at least under Bash. I tested this on Debian and CentOS 6. When the prompt contains `\l` Bash replaces it with `tty`. However liquidprompt, using `basename $(tty)`, actually replaces it with `0`. Again, I'm not using the LP_TTYN variable in my prompt so it doesn't affect me, but it may affect others.

elementalvoid:
> LiquidPrompt is implemented by using a shell hook that changes PS1 every time PS1 is displayed.

This is my point exactly. The real time taken to display the prompt is the execution of the `_lp_set_prompt` function, not the parsing and evaluation of the PS1 variable by the shell. The PS1 is parsed every time Bash/ZSh displays the prompt. This is true whether or not the PS1 variable contains a variable whose contents are static string set at shell startup (i.e. `LP_TTYN="$(basename -- $(tty) 2>/dev/null)"`) or whether the PS1 variable contains a variable with a shell escape (i.e. `LP_TTYN='\l'`).

When `PS1='${LP_TTYN}'` Bash still has to parse and evaluate the line regardless of the contents of LP_TTYN. In either case it has to evaluate, interpolate, perform parameter expansion, perform glob expansion, etc. on the contents of the PS1 variable. Then once that's complete and either `PS1='\l'` or `PS1='1'` the shell performs a final parse and evaluation in order to display the prompt. Given that the PS1 contains things like color escape sequences, username escape sequences, and hostname escape sequences Bash will _always_ perform the final parse/eval step; there's no way around it.

Out of curiosity I devised a simple performance test. This is a CPU bound test.
```
# Make sure we're running at tip top speed
[mklich:~] $ echo performance | sudo tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor

# Test with Bash escape
[mklich:~] $ export LP_TTYN='\l'; time -p bash --rcfile <(echo "PS1='${LP_TTYN} '; exec 2>/dev/null") -i < <(yes true | head -n 10000000)
real 648.76
user 590.95
sys 58.41

# Test with static string
[mklich:~] $ export LP_TTYN='3'; time -p bash --rcfile <(echo "PS1='${LP_TTYN} '; exec 2>/dev/null") -i < <(yes true | head -n 10000000)
real 626.90
user 570.57
sys 56.89

# Caclulate the benefit
[mklich:~] $ echo '(648.76-626.90)/10000000' | bc -l
.00000218600000000000
```

So, running on a Core i7-3687U the Bash escape syntax consumes, on average, an additional 2.186 microseconds per prompt display over the static string version.

As I stated in my last comment, I'm not truly vested in the direction for this variable, but it would seem to me that it's a micro-optimization that's not worth it. It changes the actual functionality of liquidprompt (from showing 'tty' to showing the tty number), it breaks on some shell configurations, and it saves microseconds per prompt display. Admittedly, my configuration issue is not your problem but the old method (Bash escape sequence) worked on my shell, your shell, and everybody else's shell, and it's not susceptible to the timing issues that I have (which others may suffer from as well).

dolmen:
> However, I've discovered that b67434a actually changed the behavior of liquidprompt, at least under Bash. I tested this on Debian and CentOS 6. When the prompt contains \l Bash replaces it with tty. However liquidprompt, using basename $(tty), actually replaces it with 0.

See #357: the LP_TTYN feature is called "variable to display the terminal number". And the bash manual says for `%l`: "the basename of the shell's terminal device name". So my implementation is consistent.

Thanks for the benchmark. I reduced it a bit. But I'm not convinced it is really reliable for the measure of PS1 evaluation.
```
$ LP_TTYN='3' time -p bash --rcfile <(echo "PS1='${LP_TTYN}$ '; exec 2>/dev/null") -i < <(yes : | head -n 10000000)
$ LP_TTYN='\l' time -p bash --rcfile <(echo "PS1='${LP_TTYN}$ '; exec 2>/dev/null") -i < <(yes : | head -n 10000000)
```

elementalvoid:
For what it is worth, true is identical to colon. See [the builtin](http://git.savannah.gnu.org/cgit/bash.git/tree/builtins/colon.def).

Issue #217, opened 2013-08-15
Mercurial "outgoing" check is slow


patramsey:
This seems like a similar question to https://github.com/nojhan/liquidprompt/issues/19 but for Mercurial. 

Setting "LP_ENABLE_HG=0" fixes the problem. The repo that this is happening on is approximately 700MB. Curious if anyone had any performance tips around this issue.

PR #402, opened 2015-10-30
using phases to count outgoing revisions in mercurial repos

hcgraf:
This change counts the commits in the "draft" phase and shows that as the number of outgoing commits. It's much cheaper than actually contacting the remote.
For very old versions of Mercurial (not supporting phases) it will always result in zero (same as before this change)

I also removed the corresponding line in "Known Limitations".

To read on Phases, see https://www.mercurial-scm.org/wiki/Phases

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,7 +342,6 @@ Set the variable to a null string (`""`) if you do not want color.
 Liquid Prompt is distributed under the [GNU Affero General Public License
 version 3](LICENSE).
 
-* Does not display the number of commits to be pushed in Mercurial repositories.
 * Browsing very large Subversion repositories may dramatically slow down
   the display of Liquid Prompt (use `LP_DISABLED_VCS_PATH` to avoid that).
 * Subversion repositories cannot display commits to be pushed because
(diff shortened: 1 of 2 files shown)


dolmen:
Related to #217.

@anthonygelibert Could you review this? I'm tempted to merge it...

hcgraf:
@anthonygelibert did you have a look at my changes? Or maybe anybody else?

Maybe some short remarks about the differences of this approach to the one for git:
* Git compares to the default remote tracking branch of the current branch, 
* Mercurial "draft" revisions are (usually) converted to "public" by pushing them to ANY other repository ("usually" because repositories can be configured as "non-publishing" which changes this behaviour, e.g. if you want to share some preliminary ideas)
  * in a centralized workflow, it should be exactly the same behaviour
  * in a customized ("advanced") distributed workflow, the number can be seen as a reminder about what has not yet been published

I'm open to questions or suggestions!

Rycieos:
This looks interesting, but I'm guess I'm not convinced that these phases correlate 100% to how remote tracking branches are used in other VCSs. v2.0.0 and forward are trying to use a generic interface over all VCS providers so that themes can design a display only once. See [`_lp_vcs_commits_off_remote()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.1/functions/data/vcs.html#lp_vcs_commits_off_remote). Though Mercurial is throwing the most wrenches in that plan with its unique design.

Anyway, I would like to see some more feedback or discussion on this before I merge it. Slightly less accurate data is probably better than no data, so I am leaning towards merging even if it is not perfect, but I would like to hear if there even is interest any more.

Rycieos:
This PR was closed because it was targeting the `develop` branch. I deleted that branch so as to not be confusing: it hasn't been current in years.

Please rebase on `master` and update this PR to target `master` instead, address my above questions, and I will re-open this. 

PR #403, opened 2015-11-04
also display the currently active bookmark (if any) in mercurial

hcgraf:
In Mercurial, "bookmarks" is the concept which closely resembles "branches" in git.
(real mercurial "branches" are part of a commits metadata, afaik there is nothing like that in git)

Many developers use bookmarks instead of named branches, so the display of the current branch is not very useful for them.

This PR checks if there is an active bookmark in the current repository ("active" means it would be moved with a commit).
If so, the bookmark name will be shown together with the branch name as `branch`,`bookmark`.
If there is no currently active bookmark, only `branch` (no trailing comma) will be shown.

Tested with bash & zsh.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -905,6 +905,13 @@ _lp_hg_branch_color()
     branch="$(_lp_hg_branch)"
     if [[ -n "$branch" ]]; then
 
+        # extract currently active bookmark, if any
+        local bookmark
+        bookmark="$(hg bookmarks | sed -n 's/^ \* \(.*[^ ]\) *[0-9]\+:[0-9a-f]*$/\1/p')"
+        if [[ -n "$bookmark" ]]; then
+            bookmark=",$(_lp_escape "$bookmark")"
+        fi
+
         local has_untracked
         has_untracked=
         if hg status -u 2>/dev/null | \grep -q '^\?' >/dev/null ; then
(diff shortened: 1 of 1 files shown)


dolmen:
As I'm not a Mercurial user, I'm asking feedback from users of liquidprompt with Mercurial:
@rubenorta @orieux @ahmedtd @livet0ski @jaesivsm @Guid75 @cykl

nedbat:
I tried this on my Mac, and it did not find bookmarks.  It worked if I changed the sed line:
```
diff --git a/liquidprompt b/liquidprompt
index f6dcab9..0167f55 100755
--- a/liquidprompt
+++ b/liquidprompt
@@ -907,7 +907,7 @@ _lp_hg_branch_color()
 
         # extract currently active bookmark, if any
         local bookmark
-        bookmark="$(hg bookmarks | sed -n 's/^ \* \(.*[^ ]\) *[0-9]\+:[0-9a-f]*$/\1/p')"
+        bookmark="$(hg bookmarks | sed -n 's/^ \* \([^ ][^ ]*\) *[0-9][0-9]*:[0-9a-f]*$/\1/p')"
         if [[ -n "$bookmark" ]]; then
             bookmark=",$(_lp_escape "$bookmark")"
         fi
```

hcgraf:
Thanks for the feedback!

So, mac-sed doesn't know '\+' for '1 for more repetitions'?
And bookmarks may actually contain spaces, you are missing that case...

Does it work if you only replace '\+' by '*'?

Am 14. November 2015 13:55:40 MEZ, schrieb Ned Batchelder <notifications@github.com>:

nedbat:
I tried it with a bookmark with spaces, and you are right, my line wasn't good.  This one works:
```
        bookmark="$(hg bookmarks | sed -n 's/^ \* \(.*[^ ]\) * [0-9][0-9]*:[0-9a-f]*$/\1/p')"
```

hcgraf:
OK, I changed the PR to avoid GNU extensions in the regex.
Thanks again for the feedback!

milouse:
You should use the command `hg --color=never bookmarks -ql .` instead of sed to retrieve current bookmark name.

Rycieos:
#627 was merged instead. But thanks for the suggestion!

[v2.0.0-beta.1](https://github.com/nojhan/liquidprompt/releases/tag/v2.0.0-beta.1) now contains [`_lp_hg_bookmark()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.1/functions/data/vcs.html#lp_hg_bookmark), and the use of that data function is in [`_lp_vcs_details_color()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.1/functions/theme.html#lp_vcs_details_color), which will display in the default theme.

PR #355, opened 2015-03-26
Added runtime to zsh prompts

stonewareslord:
Adds runtime for ZSH prompts: 
selection_012

PR #404, opened 2015-11-04
ZSH runtime (precmd/preexec corrections to previous PR #355)

hcgraf:
This is just a rebase of the previous PR #355 by @stonewareslord on top of develop.

Additionally, I made it actually work by using precmd/preexec *_functions arrays instead of just as single function (which would be overwritten by prompt_off/prompt_on)

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,7 +353,6 @@ version 3](LICENSE).
 * The analog clock requires a Unicode-aware terminal and at least a
   sufficiently complete font on your system. The [Symbola](http://users.teilar.gr/~g1951d/)
   font, designed by Georges Douros, is known to work well.
-* Displaying the runtime currently only works with Bash.
 
 
 ## Authors
--- a/liquidprompt
+++ b/liquidprompt
@@ -46,6 +46,7 @@
 # Rolf Morel        <rolfmorel@gmail.com>         # "Shorten path" refactoring and fixes
 # Thomas Debesse    <thomas.debesse@gmail.com>    # Fix columns use.
 # Yann 'Ze' Richard <ze@nbox.org>                 # Do not fail on missing commands.
+# Austen Adler      <stonewareslord@gmail.com>    # ZSH runtime
 
 # See the README.md file for a summary of features.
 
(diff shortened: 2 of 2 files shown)


dolmen:
Great!
When did this `precmd_functions` feature appeared in zsh? I just want to check users that have a zsh a bit old will not encounter issues...

hcgraf:
Apparently, it was introduced by this [Commit](http://sourceforge.net/p/zsh/code/ci/aba936b17cca86ee0742341f23f0f96795a23275) in 2006. 
The corresponding release is probably [4.3.3](http://sourceforge.net/p/zsh/code/ci/zsh-4.3.3/)

Debian Stable includes [5.0.7](https://packages.debian.org/stable/shells/zsh)
CentOS6 [4.3.11](http://rpmfind.net/linux/RPM/centos/6.7/x86_64/Packages/ZByName.html) - I have access to that, and everything works fine.
CentOS5 [4.2.6](http://rpmfind.net/linux/RPM/centos/5.11/x86_64/ZByName.html) - That one would be unsupported.

Is there anyone with such an old OS, who could test if liquidprompt WITHOUT this PR is supported anyway?

hcgraf:
oh, [this Commit](http://sourceforge.net/p/zsh/code/ci/0d101890d0faeaefa863099042e3510f5021fbc3)
introduces `add_zsh_hook` which makes things even nicer/cleaner… 
But would probably require version 4.3.4.

something like
```
autoload -Uz add-zsh-hook
add-zsh-hook precmd _lp_main_precmd
```

hcgraf:
implemented my previous comment.
I think this is much cleaner, and we are talking about compatibility to ZSH versions which are approx. 8 vs 8.5 years old…

dolmen:
zsh 4.3.4 is old enough.

Please update your commit message to say that add-zsh-hook requires zsh 4.3.4 (released in 2006):
git commit --amend && git push -f

I'll then merge it.

hcgraf:
done. Thanks for the feedback!

dolmen:
Merged! Thanks!

PR #406, opened 2015-11-04
LP_HOSTNAME_ALWAYS=-1 to never show hostname

augmentedfourth:
On OS X, I've been having problems determining connection status, and often a local connection shows up as if it were Telnet (color: WARN_RED).

Instead of adding an LP_HOSTNAME_NEVER option, or changing the option name to LP_HOSTNAME and breaking existing configs, just added a value "2" that will prevent showing hostname at all.

I realize this is a hack, because I'd rather find out the terminal status properly, but I wasn't having much luck with that so this option will hopefully suffice for now.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -548,46 +548,52 @@ _lp_connection()
 # The connection is not expected to change from inside the shell, so we
 # build this just once
 LP_HOST=""
-[[ -r /etc/debian_chroot ]] && LP_HOST="($(< /etc/debian_chroot))"
 
-# Which host symbol should we use?
-if [[ "${LP_ENABLE_FQDN}" -eq "1" ]] ; then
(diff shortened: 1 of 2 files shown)


augmentedfourth:
New change looks way more extensive than it is, since I wrapped a section in an "if" wrapper and had to increase the indentation level for everything in the block.

dolmen:
Seems ok, except I would prefer to use the value `-1` instead of `2` for this case.
Could you fix that and also the commit message of the first commit (see how I changed the new title for this PR)?

augmentedfourth:
Value changed, commit message updated, and commits squashed.

augmentedfourth:
Rebased over feature/FQDN merge

dolmen:
Thanks! Merged!

Issue #410, opened 2015-11-27
Debian sid desktop: "No support for device type: thermal" message


sebrem:
Hi,

as of recent I have

    No support for device type: thermal
    ⌂131% [sebrem:~] $

showing in every new prompt. Maybe it's more a bug in Debian/acpi but I didn't find it yet. Any ideas? I run debian unstable with latest kernel

    [sebrem:~] $ uname -a
    Linux neocortex 4.3.0-trunk-amd64 #1 SMP Debian 4.3-1~exp1 (2015-11-04) x86_64 GNU/Linux
    No support for device type: thermal
    [sebrem:~] $ cat /etc/issue
    Debian GNU/Linux stretch/sid \n \l
    [sebrem:~] $ apt show liquidprompt
    Package: liquidprompt
    Version: 1.9-4
    [...]

I think this started like 3 or 4 weeks ago when a new acpi version was uploaded to Debian but I didn't find any hint there.

dolmen:
I'm sorry. The Debian fork of Liquidprompt is not supported here. Contact @aborrero who made this package.

You should instead try the `develop` branch in this repo. Your issue is probably already fixed. If this is not the case, open another issue mentioning the last commit sha1 on the develop branch.

aborrero:
I'm sorry. Is not fixed in liquidprompt in the develop branch.

Last commit is 8f5ef7de2d3b5eb8672c3b3f12ed0c06679081f3

aborrero:
BTW, you insist in calling the debian package 'a fork', while the difference is just the name of 3 functions.
I don't understand your position of 'not supported here', it seems you are against distro packaging. That isn't how the linux world works.

I know where is the issue, but with your apparently non-collaborative behaviour makes me thinking of fixing this issue just in the debian package.

Let's not be so negative. I would really like to thank you for pinging me in this bug report. Thanks, and thanks for maintaining liquidprompt, it's really appreciated :-)

dolmen:
@aborrero My position is that I only reply to tickets against the latest commit of the `develop` branch. A packaged version made from a released version of liquidprompt is frozen from many months ago so tracking changes is too much effort for me while an interested user can make the effort of reproducing the issue from the develop branch.

dolmen:
> I'm sorry. Is not fixed in liquidprompt in the develop branch.

What is the output of the following commands?

    acpi -t ; echo $?

aborrero:
[...]

PR #414, opened 2015-12-02
liquidprompt: redirect acpi -t stderr to /dev/null

aborrero:
Let's prevent acpi from reporting missing support for devices.
We are only interested in data. If there is no data, ignore it.

This is related to issue #410

Signed-off-by: Arturo Borrero Gonzalez <arturo.borrero.glez@gmail.com>


dolmen:
This is not the right fix for issue #410. See [my comment](../issues/410#issuecomment-162003427)

PR #416, opened 2015-12-16
Force echo to enable "\." expansion for title escape sequences

pkkolos:
`LP_TITLE_OPEN` & `LP_TITLE_CLOSE` need the "\\." expansion, so explicitly enable it.

A screenshot:
set_title

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1491,9 +1491,9 @@ _lp_title()
     (( LP_ENABLE_TITLE )) || return
 
     # Get the current computed prompt as pure text
-    echo -nE "${_LP_OPEN_ESC}${LP_TITLE_OPEN}"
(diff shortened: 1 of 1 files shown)


dolmen:
The issue only had consequences in zsh.

I fixed it with a2f7a3af79ff01cf8fb4fef9a179d6d47caaaac4.

pkkolos:
Commit a2f7a3a indeed fixed the problem in zsh, although introduce a minor one in bash under tmux as shown below:
bash_tmux

Changing the escape sequence `LP_TITLE_CLOSE="$(printf '\033\\')"` to `LP_TITLE_CLOSE="$(printf '\033\\\')"` fixes the problem.

pkkolos:
Commits a2f7a3a & 229920a indeed fixed the problem.

Issue #393, opened 2015-09-21
Multiple batteries


towo:
```
17:57:45 ⌁47% towo@loki:~ 1 % acpi -V
Battery 0: Discharging, 46%, 00:55:17 remaining
Battery 0: design capacity 2809 mAh, last full capacity 2758 mAh = 98%
Battery 1: Unknown, 79%
Battery 1: design capacity 3863 mAh, last full capacity 3898 mAh = 100%
```

Pretty much says it all, I'd say.

dolmen:
I'm sorry but it doesn't say much to me. What is the behaviour that you expect?

towo:
Well, it doesn't accurately summarize what's going on with two batteries.

It just takes the one currently discharging and goes with that, and after that one's empty, it doesn't switch over to the other one.

So it'd either need two battery counters for both batteries, or a combined one giving total estimate (as GNOME does.)

dolmen:
So for the combined indicator, what formula do you expect to be applied to compute the value?

towo:
Good question, but I'd probably go by just adding up the battery percentages as per the mAh ratings.

dolmen:
Patch welcome!
The code that must be adapted is the [`_lp_battery` function](blob/master/liquidprompt#L1108).

towo:
Currently trying out a patch, if it works out, I'll have a PR by the evening.

towo:
ping.

PR #419, opened 2016-01-12
Feature/393 multiple batteries

towo:
Patch to add feature for reading out the state of multiple batteries. Uses a combined percentage of last full charge capacities to deduce the current system overall battery level.

Essentially: `(battery0_percentage * battery0_capacity + battery1_percentage * battery1_capacity + [...])/(battery0_capacity + battery1_capacity + ...)`

Might include some small rounding errors due to using `bc` at scale 2 to get the floating point parts of the calculation.

Rebased from master to develop, for full disclosure.

This would resolve nojhan/liquidprompt#393

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1255,19 +1255,19 @@ case "$LP_OS" in
         # If we found a "defect" state, i.e. a battery reading was empty
         if [[ ${bat_states[(i)defect]} -le $#bat_states ]]; then
             return 4;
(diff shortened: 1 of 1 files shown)


towo:
So, this is finished on my side, by the way.

seraphyn:
See `https://github.com/nojhan/liquidprompt/issues/493`

Issue #420, opened 2016-01-13
use $XDG_CONFIG_HOME and $XDG_CONFIG_DIRS according to XDG basedir spec


smcv:
liquidprompt has `~/.config/liquidpromptrc` as a supported config file, which is great. However, it does not follow the rest of the [freedesktop.org (XDG) basedirs specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), which is what popularized `~/.config`. Part of the value of that spec is that it gives a small number of environment variables which can be overridden once, to divert all basedir-compatible programs' configuration, cached files or other internal data (as appropriate) to a different location, with configuration and internal data using a `$PATH`-like search path rather than just a single directory.

Current implementation:

```
    elif [[ -z "$XDG_HOME_DIR" ]]
    then
        configfile="$HOME/.config/liquidpromptrc"
    else
        configfile="$XDG_HOME_DIR/liquidpromptrc"
    fi
```

`$XDG_HOME_DIR` is not part of the basedir spec and it is unlikely that users will have set it. The correct implementation would be to look in `${XDG_CONFIG_HOME:-"$HOME/.config"}` as top priority, and then in each `:`-separated element of `${XDG_CONFIG_DIRS:-/etc/xdg}` in decreasing priority order (like `$PATH`). You could either go through those in decreasing priority order and source the first file found, or go through them in reverse (increasing priority) order and source all of them (so that higher priorities override lower, like the existing support for the personal config file overriding `/etc/liquidpromptrc`).

Sample code for sourcing the first one is something like this:

```bash
if [[ -f "$HOME/.liquidpromptrc" ]]; then
    configfile="$HOME/.liquidpromptrc"
else
    local _lp_IFS="$IFS"
    local IFS=":"
    local path
    for path in "${XDG_CONFIG_HOME:-"$HOME/.config"}" $XDG_CONFIG_DIRS:-/etc/xdg}; do
        if [[ -f "$path/liquidpromptrc" ]]; then
            configfile="$path/liquidpromptrc"
            break
       fi
    done
    IFS="$_lp_IFS"
fi
if [[ -n "$configfile" ]]; then
    . "$configfile"
fi
```

dolmen:
Related to #415.

smcv:
> go through them in reverse (increasing priority) order and source all of them

[...]

PR #422, opened 2016-01-18
Source $XDG_CONFIG_DIRS/liquidpromptrc, lowest-priority first (#420)

smcv:
~/.liquidpromptrc is still the highest-priority, and
/etc/liquidpromptrc is still the lowest.

---

Refs #420. This implements the more powerful of the two options: sourcing all the possible places in the search path where we might have a liquidpromptrc. This is potentially useful if you're doing something like putting a host-specific entry in XDG_CONFIG_DIRS ahead of a generic entry.


jcrben:
Under the XDG standard, I guess I would expect liquidpromptrc to be either inside its own folder or inside the zsh or bash folder.

dolmen:
@aborrero Does Debian have guidelines for this?

smcv:
Closing this PR: there doesn't seem to be a way to tell GitHub that I now want it to be merged into a different branch :-(

I think you'd probably get more mergeable contributions and fewer confused contributors if `master` was the focus of development like it usually is (e.g. in `git` and `linux`).

aborrero:
El 10 feb. 2016 5:49 a. m., "Olivier Mengué" <notifications@github.com>
escribió:
>
> @aborrero Does Debian have guidelines for this?
>

I guess following XDG standards should be fine.
I can't analyze the problem right now (I'm in a conference).

smcv:
> Does Debian have guidelines for this?

<https://wiki.debian.org/UpstreamGuide#User_home_directories> recommends the XDG basedir spec.

smcv:
(See replacement PR, #427 - I couldn't find a way to make github change the branch into which I was asking for a merge.)

aborrero:
Please go ahead. Thanks.

PR #423, opened 2016-01-21
 _lp_are_vcs_enabled() requires sh_word_split

Hotschke:
otherwise $LP_DISABLED_VCS_PATH has no effect


liquidprompt
```diff
@@ -812,6 +812,7 @@ _lp_jobcount_color()
 _lp_are_vcs_enabled()
 {
     [[ -z "$LP_DISABLED_VCS_PATH" ]] && return 0
+    setopt local_options && setopt sh_word_split
```
dolmen: Isn't it zsh-specific?

dolmen:
Thanks! Merged!

PR #426, opened 2016-02-09
avoid “No support for device type: thermal” messages from “acpi”

illwieckz:
Hi, this modification suppress the annoying “No support for device type: thermal” message displayed at every prompt when there is no acpi thermal stuff.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1380,7 +1380,7 @@ _lp_temp_sensors()
 _lp_temp_acpi()
 {
     local i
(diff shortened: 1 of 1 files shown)


dolmen:
As your pull request is based on the `master` branch instead of `develop` (did you read [CONTRIBUTING.md](../blob/develop/CONTRIBUTING.md)), you are probably not aware of the existing fix. This issue is already fixed in the `develop` branch (see #410), unless proven otherwise.
Yeah, I know. I should make a release.

PR #427, opened 2016-02-10
Load exactly one configuration file, searching XDG_CONFIG_DIRS

smcv:
~/.liquidpromptrc is still the highest-priority, and
/etc/liquidpromptrc is still the lowest.

---

Rebased on `develop`, which already looks in the correct `$XDG_CONFIG_HOME` (not `$XDG_HOME_DIR` like `master` did), but does not do the full path-search behaviour described in the basedir spec.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -418,14 +418,15 @@ _lp_source_config()
     if [[ -f "$HOME/.liquidpromptrc" ]]; then
         configfile="$HOME/.liquidpromptrc"
     else
+        local first
         local search
         # trailing ":" is so that ${search#*:} always removes something
         search="${XDG_CONFIG_HOME:-"$HOME/.config"}:${XDG_CONFIG_DIRS:-/etc/xdg}:"
         while [[ -n "$search" ]]; do
-            path="${search%%:*}"
+            first="${search%%:*}"
             search="${search#*:}"
-            if [[ -f "$path/liquidpromptrc" ]]; then
-                configfile="$path/liquidpromptrc"
+            if [[ -f "$first/liquidpromptrc" ]]; then
+                configfile="$first/liquidpromptrc"
                 break
             fi
         done


dolmen:
To preserve the existing behavior (only one file loaded), move `$HOME/.liquidpromptrc` as the first path in the `search` variable.

smcv:
> To preserve the existing behavior (only one file loaded)

At the moment up to two files are loaded: the one from `/etc`, and a single user-specific file (in that order, so the per-user file can override the per-system one). I had expected that you'd prefer one of the two possible extremes, both of which seem internally consistent:

* only load up to one file, the first (most specific) that exists from among { `~/.liquidpromptrc`; `$XDG_CONFIG_HOME/liquidpromptrc`; (`$d/liquidpromptrc` for each `$d` in `$XDG_DATA_DIRS`); `/etc/liquidpromptrc` }; or
* load any number of files from that list, in reverse order so that more specific files can override less specific

The current behaviour is somewhere between those two: it will always load both the file from `/etc` and the user-specific file (assuming both exist), but it will never load both `~/.liquidpromptrc` and `~/.config/liquidpromptrc`. That seems a bit inconsistent to me; what's your preferred resolution for that inconsistency? Load all, load up to one, or load up to one but as a special case additionally load `/etc/liquidpromptrc` separately?

`$XDG_CONFIG_DIRS` can contain a mixture of per-user and per-system paths. The default is just `/etc/xdg`, which is obviously entirely per-system; but users who split their dotfiles (perhaps to keep some but not all of them in git) can do things like `XDG_CONFIG_HOME="$HOME/.config.$(hostname)"; `XDG_CONFIG_DIRS="$HOME/.config:/etc/xdg`.

dolmen:
The only sane solution is to load just one config as proposed in your first point.
Because config files in /etc may have been written for a liquidprompt installed system-wide, while the version of liquidprompt used is not that one and may differ in the interpretation of the config file.

smcv:
> The only sane solution is to load just one config as proposed in your first point.

So to clarify: liquidprompt currently loads both `/etc/liquidpromptrc` and `~/.liquidpromptrc` if present, and you would like me to change that behaviour? Not the resolution I'd expected, but OK.

smcv:
> The only sane solution is to load just one config as proposed in your first point.

Branch rewritten to do that.

> Because config files in /etc may have been written for a liquidprompt installed system-wide, while the version of liquidprompt used is not that one

That does make sense.

smcv:
Please also cherry-pick 0a196a9 to fix the zsh breakage mentioned on #420.

dolmen:
Merged a month ago. Closing.

PR #431, opened 2016-03-20
_lp_temp_acpi: redirect error message to /dev/null

starox:
Not all systems support acpi thermal device.
In these case acpi command outputs 'No support for device type: thermal'
which is annoying.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1487,7 +1487,7 @@ if (( LP_ENABLE_TEMP )); then
     {
         local -i i
         # Only the integer part is retained
-        for i in $(LANG=C acpi -t |
+        for i in $(LANG=C acpi -t 2>/dev/null |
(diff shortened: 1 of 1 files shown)


dolmen:
We are already running `_lp_temp_acpi` with `2>/dev/null` during initial detection. See line 1511.
If there is *no support for device type: thermal*, `_lp_temp_acpi` should fail, and so not be considered ok during detection in `_lp_temp_detect`.
So I don't know how this issue could happen, and so how your patch really fixes an issue.

I need a copy of the full raw output of `acpi -t` on your system.

starox:
You are right, this is already fixed. I apologize to take your time.
When I had the problem I was on the master branch. I switched to the develop branch but didn't quit my shell to reload liquidprompt. So I believed that the bug was still there.
A quick test today show that everything is working right on the develop branch.

Have a good day,

dolmen:
No worries. You are confirming exactly what I suspected. :)

PR #432, opened 2016-03-26
Fix RPS1 variables not expanding

alexprengere:
Hello,
I updated to the latest version of liquidprompt (develop), and it broke my zsh right prompt.
I have some variables that are computed using the `precmd` in my `zshrc`, then these variables are displayed in the right prompt of zsh, using the `RPS1`.

A recent commit in `develop` set the `nopromptsubst`, this prevents the variables from expanding in the right prompt, leaving things like `$VARIABLES` instead. I did not fully understand why this was added in the first place, but removing it fixes the right prompt and I did not notice other problems occurring after.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1876,7 +1876,7 @@ prompt_on()
         [[ -n "$_LP_ZSH_HOOK" ]] && add-zsh-hook -d precmd $_LP_ZSH_HOOK
         # Set options that affect PS1 evaluation
         # Disable parameter/command expansion; enable percent expansion
-        setopt promptpercent nopromptbang nopromptsubst
+        setopt promptpercent nopromptbang
         # 'time' doesn't seem to work on shell functions: no time output
         #if (( LP_DEBUG_TIME )); then
         #    _lp_main_precmd() {


dolmen:
`promptsubst` enabled is a security issue for liquidprompt. Try this:
```sh
set -o promptsubst
mkdir '<<${PATH}>>'
cd '<<${PATH}>>'
cd ..
rmdir '<<${PATH}>>'
mkdir '<<$(echo abc)>>'
cd '<<$(echo abc)>>'
cd ..
rmdir '<<$(echo abc)>>'
```

Use `promptsubst` at your own risk. But I don't want to continue to expose Liquid Prompt users to this.

I will have to document that `nopromptsubst` is set as a breaking change though.

dolmen:
Note: my demo above will not show the issue if you have `LP_PATH_KEEP=-1` (as I see you have in #433). But there are other areas of Liquid Prompt than the path that are at risk. Think about a malicious VCS branch name.

alexprengere:
I understand the rationale, but as a zsh user, this is still a big change (for example this breaks [zsh vi mode](http://dougblack.io/words/zsh-vi-mode.html)). For me, liquidprompt has worked for years without this option activated. I did not dive into the code, but could this option be configured through the `liquidpromptrc`?

dolmen:
> could this option be configured through the liquidpromptrc?

No because this is a security issue.

You can still enable `promptsubst` from your `.zshrc` once liquidprompt is loaded, but at your own risks. Just add this line:
```
set -o promptsubst
```

alexprengere:
This works perfectly, thanks!

PR #434, opened 2016-03-31
Fix runtime counter

kuehnelth:


dolmen:
Thanks for the patch. I finally choose to implement it in a different way in d485ed191fb6b896290a32848c4fefd0342e2046.

PR #435, opened 2016-04-02
Add ruby/gemset support through rbenv or rvm

### LINKED ISSUE #249: Show Ruby version
Something like this could work:

```
if [ -e ~/.rvm/bin/rvm-prompt ]; then
  RUBY_ENV=$(~/.rvm/bin/rvm-prompt i v g)
else
  if which rbenv &> /dev/null; then
    RUBY_ENV=$(rbenv version | sed -e 's/ (set.*$//' -e 's/^ruby-//')
  fi
fi
```

milouse:
Fix #249

### DIFF
--- a/liquid.ps1
+++ b/liquid.ps1
@@ -14,6 +14,7 @@
 # LP_VENV Python virtual environment
 # LP_PROXY HTTP proxy
 # LP_VCS the content of the current repository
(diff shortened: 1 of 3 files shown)


nojhan:
Merged on develop, thanks!

milouse:
De nada :)

PR #438, opened 2016-04-19
Fix PATH variable for zsh.

zauguin:
zsh interprets `path` as an array version of `PATH`, so 36914a089aee00ec1f41c96cb227638cbd0ec53c changed the PATH and broke almost everything for zsh users.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -419,6 +419,7 @@ _lp_source_config()
         configfile="$HOME/.liquidpromptrc"
     else
         local search
+        local path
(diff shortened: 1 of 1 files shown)


zauguin:
Fixed by 7b2bb7815cf6493e18bb8fc5e5585832b2b0ac6f

dolmen:
Thanks for the patch, even if yours was not applied!

dolmen:
This isuse was related to #427.

PR #440, opened 2016-05-02
Adding overridable variable for the separator between user and host

tstone2077:
because I like having a space between user and host for the prompt.
For example:
[ user @ hostname : /pwd ] master $


nojhan:
Not merged, but the feature has been added to develop with LP_MARK_X11.

PR #442, opened 2016-05-12
Add prefix and postfix not shown in title

nojhan:
When using LP_ENABLE_TITLE=1, the prompt will be dispalyed as is in the window's title. This could be a problem if you want to customize the prompt with fancy stuff, like a large horizontal line on a new line.

This feature add the LP_PS1_PRETITLE and LP_PS1_POSTTITLE variables, added to PS1 only after the LP_TITLE has been generated.

For example, if you want to add a line separator across all you terminal:

LP_PS1_PRETITLE="$(printf '%*s\n' $(tput cols) '' | tr ' ' '-')"

If you want to customize the typing field:

LP_PS1_POSTTITLE="${WHITE}〉 "

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,27 @@ it to add a keyword to each of your different terminals:
     mycode [:~/code/liquidprompt] develop ±
 
 
+### Adding a Prefix/Postfix not shown in the title
+
+When using `LP_ENABLE_TITLE=1`, the prompt will be dispalyed as is in the window's title.
+To prefix or postfix the prompt only in the command line and not in the title,
+use `LP_PS1_PRETITLE` and `LP_PS1_POSTTITLE`.
+
+For example, if you want to add a line separator across all you terminal:
+
+    LP_PS1_PRETITLE="$(printf '%*s\n' $(tput cols) '' | tr ' ' '-')"
+
+If you want to customize the typing field:
+
+    LP_PS1_POSTTITLE="${WHITE}〉 "
+
+Note: in this example, you most probably want to reset the color of the output.
+You should thus add a call to a reset command before anything is run.
+In bash, you should add the following debug trap:
+
+    trap 'tput sgr0' DEBUG
+
+
 ### Rearranging the Prompt
 
 You can sort what you want to see by sourcing your favorite template file
--- a/liquidprompt
+++ b/liquidprompt
@@ -310,6 +310,8 @@ _lp_source_config()
     LP_PS1=${LP_PS1:-""}
     LP_PS1_PREFIX=${LP_PS1_PREFIX:-""}
     LP_PS1_POSTFIX=${LP_PS1_POSTFIX:-""}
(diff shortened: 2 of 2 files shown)


dolmen:
> When using LP_ENABLE_TITLE=1, the prompt will be dispalyed as is in the window's title.

The logic is the opposite (but still gives the same result): the full prompt content is exposed as the terminal title. At least that's how I understood the intent of the code since I took over maintenance.

I think that the current design for the term title is not good: it should have its own design separate from PS1. But I have not much interest in working on that (or even thinking about a better desing) because I don't use the feature. Besides that we should just echo the title instead of escaping it and putting in PS1 so the shell echoes it.

Rycieos:
I've been working on a fix or replacement for the current title functionality, and have some thoughts:

We can't change the API of `_lp_title()`, as it is the one function used in the default liquid.ps1. It's probably the one "private" function that just can't change. There are probably a sizable number of users using it. This means a few things:
* It must return it's output through stdout.
* It must accept color and shell escape codes, and remove them.
* It must wrap the whole returned string with shell escape codes.

I had wanted to remove the title from PS1 for two reasons: First is that trying to read or debug your PS1 is much harder when there is a ton of extra junk in it. Second, the shell reprints PS1 if the terminal is redrawn, which can happen with `^L` or on resize. But that is false; the terminal won't redraw PS1 unless it is waiting on the command line. I would only want the title to not be overridden when a command is running (since I want to make the title the running command in the future). And second, `^L` to clear screen also only works when on the command line.

The true reason that I think the title should stay as part of PS1 is that if we echo `\u@\h` directly to the terminal (surrounded in title escapes), the terminal will set the title as exactly that. If we want to put the username in the title, we have two options:
1. Discover that username and save it for use (which Liquidprompt isn't doing right now).
2. Embed the title string in PS1 and let the shell expand it for us.

2\. is obviously easier.

So how to solve the original problem? `_lp_title()` already can accept any string, it just needs to be configured to do so. Users actually have that power right now with the "old" themes, since they control what they pass into `_lp_title()`. And I don't see a reason to change that.

So I guess the answer is, right now I'm planning on changing nothing, unless someone suggests a better default for the title than the whole of PS1. I'm open to other suggestions of course.

Rycieos:
Liquidprompt v2.0, with its theme support, includes better tools for setting the title independently from the prompt as a whole. This should be done either by [a theme](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/theme/custom.html) (see below for an example) or in a [default theme template file](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/theme/default.html#templates).

https://github.com/nojhan/liquidprompt/blob/2daf0e453039950d79dcf0eec5a04931e363a92a/themes/powerline/powerline.theme#L55-L69

Either of these options is both controlled by the user and offers more control than this patch would provide, so I'm closing this PR. If anyone has something specific they want to do with the terminal title that is not possible currently, please open an enhancement request.

PR #443, opened 2016-05-31
Correct SUDO detection on Mac OS X

anthonygelibert:
Correct SUDO detection on Mac OS X where true isn't a provided as a /bin command.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -556,7 +556,7 @@ if (( EUID != 0 )); then  # if user is not root
         #   sudo -K   # revoke your credentials
         _lp_sudo_check()
         {
-            if sudo -n /bin/true 2>/dev/null; then
+            if sudo -n true 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


dolmen:
So where is the `true` binary stored?
```
$ which true
```

anthonygelibert:
Excuse-me… I was pretty hurry, I thought it was only a "shell built-in command" ("which true" in my ZSH).

After your answer, I checked and it's in /usr/bin/true.

So, maybe a:
```sudo env true```

Will be better?

dolmen:
The point of this line is to run *quickly* a dummy command that has a 0 exit code.
`env true` would fork one more time, so this misses the point.
I'm merging your patch as is.

dolmen:
Cherry-picked as 3d3061c846673fc6593595510907fbfc1da978d3. Thanks!

PR #444, opened 2016-06-01
OS X: missing battery charge level

augmentedfourth:
_lp_battery refactor in 4636e0e caused battery percentages to no longer
be displayed in OS X. Adding the echo statements it removed back in.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1291,12 +1291,26 @@ case "$LP_OS" in
             return 4
             ;;
             discharging)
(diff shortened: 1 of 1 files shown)


augmentedfourth:
I wish I'd been able to get this into 1.10, but I only just noticed the problem when pushing the release to Homebrew today...

dolmen:
Thanks for the report. I choose to implement the fix in a different way.

PR #456, opened 2016-06-30
Make the maximum load a configuration option

mgeb:
This is a feature request to add a configuration option that allows the user to define what is considered the maximum load when choosing colors from the colormap.

The pull request adds a config option named LP_LOAD_SCALE_MAX that is passed to _lp_color_map() as second argument (scale) when the color for the load indicator is determined. LP_LOAD_SCALE_MAX defaults to 100 which is the default scale used by _lp_color_map(), so the status quo is not changed.

While it is not strictly necessary, it'd probably make sense to merge this together with my other pull request #455, which fixes colors for loads >= scale.

The second commit in this request is documentation only. Feel free to not merge it if you want to document this yourself.


nojhan:
Merged on develop, thanks.

PR #459, opened 2016-07-13
Display virtualenv parent dir name instead of its own.

romantolkachyov:
I'm always create virtualenv with name `venv` inside project. Also I used virtualenvs `--prompt` argument to specify name for my venv (because it `venv` in every project). But it doesn't work with liquidprompt because it rewrites PS1.

So I've added an option LP_VIRTUALENV_PARENT (default to 0). If enabled, liquidprompt will display venv parent dir name instead of fixed in my case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,14 @@ To erase your new formatting, just bind `LP_PS1` to a null string:
     export LP_PS1=""
 
 
+### Display virtualenv parent dir name
+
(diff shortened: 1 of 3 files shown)


dolmen:
This is a very obscure feature that you are proposing, and the implementation adds cost at runtime.
I will not merge it.

PR #460, opened 2016-07-14
fixed pmset regex for OS X Battery status

tsia:


liquidprompt
```diff
@@ -1285,7 +1285,7 @@ case "$LP_OS" in
     {
         (( LP_ENABLE_BATT )) || return 4
         local percent batt_status
-        eval "$(pmset -g batt | sed -n 's/^ -InternalBattery[^	 ]*[	 ]\([0-9]*[0-9]\)%; \([^;]*\).*$/percent=\1 batt_status='\'\\2\'/p)"
+        eval "$(pmset -g batt | sed -n 's/^ -InternalBattery[^	 ]*[	 ]*\([0-9]*[0-9]\)%; \([^;]*\).*$/percent=\1 batt_status='\'\\2\'/p)"
```
dolmen: Is this a case where there is is more than one space?
In this case `[	 ]*` should be replaced with `[	 ][	 ]*`.

tsia:
there are also tab characters. but i will close this request because it seems that it doesn't work reliably for me anymore.

i sometimes get no battery display and sometimes there are values between 85% and 113% quickly changing.

Issue #461, opened 2016-07-19
Hostname hash cksum not portable


Rycieos:
Shell: bash
Operating system: SunOS 5.11 11.2 sun4v sparc sun4v
Liquid Prompt version (tag, commit): 1.11, 488bb5975db9fa8be0b795019f0b6d99ec792f50

Offending line 292:
`local hash=$(( 1 + $(hostname | cksum | cut -d " " -f 1) % 6 ))`

Error message:

```
~/.liquidprompt/liquidprompt: line 292: 1 + 1903320147    12 % 6 : syntax error in expression (error token is "12 % 6 ")
```

Problem:
On SunOS, cksum is returning the two values with a `<TAB>` character between them, where the above line is looking for spaces. This makes cut return the whole line, which bash does not know how to math on.

Expected value:

```
hostname | cksum
1903320147<SPACE>12
```

Actual value:

```
hostname | cksum
1903320147<TAB>12
```

Note that although bash seems to print the above with multiple spaces in between, cut will not cut with the delimiter set to spaces.

I was able to fix this problem by removing the `-d " "` in the line above. Since cut's default delimiter is a `<TAB>`, it works as expected.

mvndrstl:
While cut does not support picking multiple delim chars, awk supports it, as well as having all types of whitespace as the default.

The following works on all of my machines:
`hostname | cksum | awk '{print $1}'`

dolmen:
awk is a no-no in Liquid Prompt, because it works universally, but universally slowly for a simple task such as parsing a single line.
```
$ ls -l /usr/bin/cut /bin/sed /usr/bin/gawk
-rwxr-xr-x 1 root root  73352 févr. 13  2014 /bin/sed
-rwxr-xr-x 1 root root  43680 mars  10 20:10 /usr/bin/cut
-rwxr-xr-x 1 root root 441512 juil.  2  2013 /usr/bin/gawk
```

`sed 's/[^0-9].*$//'` is less worse, so you can fix your PR to use that.
Please also tell in the commit message the reason of the fix ("On SunOS, cksum use tab as delimiter instead of space as on other platforms").

Rycieos:
Updated pull request with `sed` instead of `awk`. I figured since it was part of the init, speed didn't matter as much, but the speedup of `sed` is worth it.

Rycieos:
Closing since the issue was fixed in proposed pull request.

PR #462, opened 2016-07-19
Fix computing hostname hash on SunOS

Rycieos:
Fix for #461 

As noted in the issue, `cut -d " " -f 1` does not work universally, since `cksum` does not have the same delimiting character on all OS's. `awk '{print $1}'` works universally, since it splits on any type of white-space.


Rycieos:
Closing since the pull request was merged.

PR #464, opened 2016-07-27
Avoid recursion when launching new shell

ncs-msmith:


dolmen:
Your change in interesting, but is not in the right place.
The right fix is to change how `OLD_PROMPT_COMMAND` is initialized: if `PROMPT_COMMAND` contains `_lp_set_prompt` we should warn about an incorrect configuration and keep OLD_PROMPT_COMMAND empty.

Issue #349, opened 2015-03-03
Show path relative to repo


felixSchl:
It would be nice if I could display the path in my prompt relative to the git repository I am working on. This removes all the unnecessary parts of the path and leaves me to focus on the repo itself. Is this functionality already available, and if not is it something worthwhile looking into?

Furthermore, would it be possible to know if we are inside a submodule? So we could show the housing git repo's name somewhere as well? This would be a recursive op, so the displayed level of parent repos would need to be controlled.

dolmen:
> It would be nice if I could display the path in my prompt relative to the git repository I am working on. This removes all the unnecessary parts of the path and leaves me to focus on the repo itself. Is this functionality already available, and if not is it something worthwhile looking into?

No and yes.

> Furthermore, would it be possible to know if we are inside a submodule?
Yes it is probably possible. Is it sane to look into this when displaying the prompt? No. Too much runtime cost.

paulvt:
Or just color the part of the prompt that is relative to the base of the VCS.  At one time I used some code that printed it like this: ~/projects/some-repo[master]/sub/dir.  However, given the different approach of liquidprompt (only append things when needed) and features such as part shortening, this might not work.

rhopfer:
Sorry for the advertisement, but maybe you are interested how I do this in [my prompt](https://github.com/rhopfer/bash-prompt). I use an environment variable PROMPT_BASES to store special paths which should be reduced to a name. It should be easy to adopt this to your prompt.

nojhan:
I like the idea, but it conflicts with features related to the path management (shortening, for instance). I think the idea of @paulvt feels more ergonomic: just highlight the part of the path that is under to the current repository root (the part that is shown). The prompt of @rhopfer makes me realize that we can also color the slashes.

felixSchl:
I'd be happy to put in the hours to support both. I know I have not properly addressed the issue of path shortening (I listed it in the TODOs of my PR), but I would assume it should simply shorten the shortened (relative to repo) path. Either way, I have been using this at work for the last week and it's been really nice to use. I made a quite asciinema demo to demonstrate it: https://asciinema.org/a/21be5t92ib26uddrs9ehuaovr.

Rycieos:
This was (partially) implemented as part of [v2.0.0-rc.1](https://github.com/nojhan/liquidprompt/releases/tag/v2.0.0-rc.1).

[...]

PR #465, opened 2016-09-06
WIP - shorten directory relative to repo root

felixSchl:
demo: https://asciinema.org/a/21be5t92ib26uddrs9ehuaovr

I reported this as a feature request early last year: https://github.com/nojhan/liquidprompt/issues/349.
I have since brushed up on my shell skills and feel comfortable delivering the said feature set.

What's left to do?
- [x] show relative path when in **git** repo
- [x] show relative path when in **svn** repo
- [x] show relative path when in **hg** repo
- [x] show relative path when in **bzr** repo
- [x] show relative path when in **fossil** repo
- [ ] configure colouring of repo name (currently uses `$LP_COLOR_PATH_ROOT`) - **input required**
- [ ] toggle feature ON/OFF via option - **input required**
- [ ] how should this interact with the ordinary directory shortening? - **input required**

The way it's implemented it _should_ be easy to add support for the supported VCS systems, provided they have a way to get the top-level directory... otherwise we might half to walk the file system, but I hope not.

I'd say supporting submodules as mentioned in the original issue would be a separate PR, but I personally lost interest in that feature.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -666,6 +666,32 @@ _lp_get_home_tilde_collapsed()
     echo "${PWD/#$HOME/$tilde}"
 }
 
+# Shorten the path of the current working directory by making it relative
+# to the VCS root, if in a VCS tracked directory
+# * Show directory name (as repo "name")
+# * Show the relative path from the root to "$PWD" in a different color
+_lp_shorten_vcs_path ()
+{
+    local root
+    case "$LP_VCS_TYPE" in
+        git*)   root="$(git rev-parse --show-toplevel)" ;;
+        hg)     root="$(hg root)" ;;
+        svn)    LANG=C root="$(svn info | grep 'Working Copy Root Path' | cut -d' ' -f5- | xargs)" ;;
+        bzr)    root="$(bzr root)" ;;
+        fossil) root="$(fossil status | grep local-root | cut -d' ' -f2- | xargs)" ;;
+    esac
+
+    # trim leading and trailing whitespace, as well as a potentially trailing slash
+    if [[ -n "$root" ]]; then
+        local name="${root##*/}"
+        local relpath="${${PWD#$root}%/}"
+        LP_PWD="${LP_COLOR_PATH_ROOT}$(_lp_escape "$name")$NO_COL"
+        LP_PWD="${LP_PWD}${LP_COLOR_PATH}$(_lp_escape "$relpath")$NO_COL"
+    else
+        _lp_shorten_path
+    fi
+}
+
 # Shorten the path of the current working directory
 # * Show only the current directory
 # * Show as much of the cwd path as possible, if shortened display a
@@ -1766,8 +1792,6 @@ _lp_set_prompt()
             fi
         fi
 
-        _lp_shorten_path   # set LP_PWD
-
         if _lp_are_vcs_enabled; then
             LP_VCS="$(_lp_git_branch_color)"
             LP_VCS_TYPE="git"
@@ -1814,6 +1838,12 @@ _lp_set_prompt()
 
         # end of the prompt line: double spaces
         LP_MARK="$(_lp_sr "$(_lp_smart_mark $LP_VCS_TYPE)")"
+
+        if [[ -n "$LP_VCS" ]]; then
+            _lp_shorten_vcs_path
+        else
+            _lp_shorten_path
+        fi
 
         LP_OLD_PWD="LP:$PWD"
 


felixSchl:
Alright, detecting the root is implemented for all supported VCSs (afaict). I've also had to fix up the awk script to run on OSX, but I could imagine it being broken elsewhere since the `+` character was not escaped in the regex to mean literal `+`. I also took the liberty to span it across more lines to make it readable, but if you're precious of you sloc count, I can collapse them again.

dolmen:
I've cherry-picked your second commit ("Fix up awk script and make readable") and it was good and independent of the rest of this PR. Please rebase your branch on master and push -f.

liquidprompt
```diff
@@ -666,6 +666,31 @@ _lp_get_home_tilde_collapsed()
...
+# to the VCS root, if in a VCS tracked directory
+# * Show directory name (as repo "name")
+# * Show the relative path from the root to "$PWD" in a different color
+_lp_shorten_vcs_path () {
+    local root x
+    root="$(
+        case "$LP_VCS_TYPE" in
+            git*)   git rev-parse --show-toplevel; ;;
+            hg)     hg root; ;;
+            svn)    svn info | grep 'Working Copy Root Path' | cut -d' ' -f5-; ;;
```
dolmen: Add LANG=C

liquidprompt
```diff
@@ -666,6 +666,31 @@ _lp_get_home_tilde_collapsed()
     echo "${PWD/#$HOME/$tilde}"
 }
 
+# Shorten the path of the current working directory by making it relative
+# to the VCS root, if in a VCS tracked directory
+# * Show directory name (as repo "name")
+# * Show the relative path from the root to "$PWD" in a different color
+_lp_shorten_vcs_path () {
+    local root x
+    root="$(
```
dolmen: Move the `root` assignment inside each case.

liquidprompt
```diff
@@ -666,6 +666,31 @@ _lp_get_home_tilde_collapsed()
...
+            git*)   git rev-parse --show-toplevel; ;;
+            hg)     hg root; ;;
+            svn)    svn info | grep 'Working Copy Root Path' | cut -d' ' -f5-; ;;
+            bzr)    bzr root; ;;
+            fossil) fossil status | grep local-root | cut -d' ' -f2-; ;;
+        esac
+    )" || return
+    # trim leading and trailing whitespace, as well as a potentially trailing slash
+    root="${${${root##*[[:space:]]}%%[[:space:]]*}%/}"
+    if [ ! -z "$root" ]; then
```
dolmen: `if [[ -n "$root" ]]; then`

felixSchl:
Alright, done.

liquidprompt
```diff
@@ -1802,6 +1825,9 @@ _lp_set_prompt()
         # end of the prompt line: double spaces
         LP_MARK="$(_lp_sr "$(_lp_smart_mark $LP_VCS_TYPE)")"
 
+        _lp_shorten_path   # set LP_PWD
+        _lp_shorten_vcs_path
```
dolmen: You are calling both `lp_shorten_path` and `_lp_shorten_vcs_path`. This is inefficient as only the result of one of them will matter.

1. Instead of calling both, `_lp_shorten_vcs_path` should delegate to `_lp_shorten_path` in case it does not fill `LP_PWD`  
2. The call to `_lp_shorten_vcs_path` should be in a `if [[ -n "$LP_VCS" ]]` block

dolmen (changes requested):
This implementation (even with my remarks applied) will break the _`bash with LP_ENABLE_SHORTEN_PATH=0 and PROMPT_DIRTRIM set`_ case.

felixSchl:
> This implementation (even with my remarks applied) will break the bash with `LP_ENABLE_SHORTEN_PATH=0` and `PROMPT_DIRTRIM` set case.

From the comments above it seems you want to do either trimming or relative dirs? What do you think about combining the to, i.e. generalising the code to trim and then apply it as a last transform on whatever is present?

dolmen:
>>This implementation (even with my remarks applied) will break the bash with LP_ENABLE_SHORTEN_PATH=0 and PROMPT_DIRTRIM set case.
>From the comments above it seems you want to do either trimming or relative dirs?

No. I'm just telling the this particular user configuration has been [so optimised](../blob/e681e64ecdd10fdd0ef3344b26ed64bb84b26d7e/liquidprompt#L746) in liquidprompt that `_lp_shorten_path` is replaced by an implementation never update LP_PWD. If LP_PWD is changed by something else, nobody will restore it when exiting a VCS controlled directory.

One major concern is stability. I like your relative dirs idea (so once we have a good implementation I want it to be enabled by default), but I want existing users who upgrade to still be able to switch it off and go back to the previous behaviour.

> What do you think about combining the two, i.e. generalising the code to trim and then apply it as a last transform on whatever is present?

This is interesting, but this is a much more tough project. It would require a more major rework of Liquidprompt and to be much more careful to not break things.

liquidprompt
```diff
@@ -670,24 +670,25 @@ _lp_get_home_tilde_collapsed()
...
+        hg)     root="$(hg root)" ;;
+        svn)    LANG=C root="$(svn info | grep 'Working Copy Root Path' | cut -d' ' -f5- | xargs)" ;;
+        bzr)    root="$(bzr root)" ;;
+        fossil) root="$(fossil status | grep local-root | cut -d' ' -f2- | xargs)" ;;
+    esac
+
     # trim leading and trailing whitespace, as well as a potentially trailing slash
-    root="${${${root##*[[:space:]]}%%[[:space:]]*}%/}"
-    if [ ! -z "$root" ]; then
+    if [ -n "$root" ]; then
```
dolmen: Use double brackets: `[[ ... ]]` instead of `[ ... ]`

felixSchl:
> It would require a more major rework of Liquidprompt and to be much more careful to not break things.

What is holding me back a little is the lack of automated tests that I can run my changes through. Manually verifying is quite mundane for so many scenarios. I saw that many functions change global variables and are expected to be called in certain places under certain conditions. For example, I would have expected the shortening function to take as input the path to shorten and then return the new, shortened path, which would then be set at call site. I tried to make the added vcs shorten function to follow the existing pattern.

Would you accept a PR to set up CI with say travis?

felixSchl:
Btw, I am trying to reduce the number of added commits as we go along, but we can certainly squash later on once we're happy to remove commits like the one that just came in ("Use double-brackets").

dolmen:
:+1: for Travis. But we have no testsuite. The test.sh is very old and does not work with the current code. Open another issue to discuss that.

dolmen:
> Btw, I am trying to reduce the number of added commits as we go along, but we can certainly squash later on once we're happy to remove commits like the one that just came in ("Use double-brackets").

You can squash now if you want and push -f.

Rycieos:

Part of v2.0 was to support multiple path shortening methods and to highlight the VCS root and last directory. This gets close to what you were trying to do here, but none of the shortening methods show the path relative to the repo root (though all path formats will never shorten the repo root if `LP_PATH_VCS_ROOT` is enabled). If you want to rebase and refactor to add a shortening method that would do that, I would be happy to merge such a feature.

Docs that would be helpful:
[`_lp_path_format()`](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/functions/data.html#lp_path_format)
[`LP_PATH_METHOD`](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/config.html#LP_PATH_METHOD)
[`LP_PATH_VCS_ROOT`](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/config.html#LP_PATH_VCS_ROOT)

And the `_lp_path_format()` source:
https://github.com/nojhan/liquidprompt/blob/2daf0e453039950d79dcf0eec5a04931e363a92a/liquidprompt#L861-L867

And feel free to ask for help!

Issue #469, opened 2016-09-17
Set up test suite and CI


felixSchl:
I am trying to land #465 but verifying it won't break existing behavior is mundane and probably incomprehensive. It would be great to set up a test suite that has the ability to test various pieces of functionality automatically. I have used https://github.com/sstephenson/bats in my personal projects with pretty good success but I am happy to hear about alternatives. Writing tests in retrospective is no fun, so I'd propose (and volunteer my time) to set up enough of a testing framework that tests can be added easily, so they can accumulate over time. What are your thoughts on this? In terms of CI, I am using travis for my personal open source projects. Whatever we choose, we can have it install all supported shells and vcs and so on and run the tests in a matrix.

nojhan:
There is an old first attempt of automated test in the `test.sh` file,
maybe useful to gather some ideas.

dolmen:
:+1: for bats: I like that it can output TAP. The only drawback: I see nothing about a port on zsh.
:+1: for Tavis-CI: I'm already using it for all my Perl projects (I am using it extensively for unit testing of my other shell prompt project, [angel-PS1](/dolmen/angel-PS1); see its [.travis.yaml](/dolmen/angel-PS1/blob/release/.travis.yml))

So go ahead!

For the level of testing I think we should concentrate on this level of functional test:

- setup:
    - setup an environment
    - define a liquidprompt configuration
    - run liquidprompt
- for each test:
    - run a command that changes the environment (for example: chdir)
    - run _lp_set_prompt
    - check the values of the `LP_*` and/or `PS1` variables

Testing the changes of the LP_* variables seems to me the right level of testing because those are publicly documented, so they are a kind of API we are exposing to the world. Any test that relies on deeper details of implementation (such as values of the private `_LP_*` variables or the `_lp_*` functions) will be a pain to maintain.

felixSchl:
@dolmen I think we are thinking very much alike. Unfortunately bats does not run on zsh, but we can easily enough run zsh from a test case. We still get plenty of benefits from using bats.

https://github.com/nojhan/liquidprompt/pull/470/files#diff-e83ee8c6737007c4e99566f73a1f8d65R12

I think we can spruce it up quite a bit still, but it should do for now. Once we have written say 3 or 4 tests a pattern will likely emerge that we can then abstract.

nojhan:
See PR #470

Rycieos:
[...]

PR #470, opened 2016-09-18
WIP - Set up bats based test bed

felixSchl:
Complementary PR for #469 

I will remove tabs in favour spaces for consistency once I am done or closing in on being done. I find it quite pleasant to work with hard tabs in bash.


felixSchl:
I think the it's at a stage where some initial feedback would be helpful. To explain what I have done so far:

The test bed itself is shell agnostic, but depending on how you execute bats, a different shell will be used to interact with liquidprompt:

```
TEST_SHELL=bash bats/tests
TEST_SHELL=zsh bats/tests
```

Most of this logic sits in [support.bash](https://github.com/felixSchl/liquidprompt/blob/f4339f713e46b1ec4ae47a80bc6cbdfdc3ff603d/tests/support.bash).

I think that's the right way to do it because liquidprompt aims to work out of the box on each. We can specialize tests by checking `$TEST_SHELL`, if need be.

I have also [implemented a function that allows fetching symbols from the underlying shell](https://github.com/felixSchl/liquidprompt/blob/f4339f713e46b1ec4ae47a80bc6cbdfdc3ff603d/tests/support.bash#L122), so that they can be compared in the host shell. This allows to check for a symbol in PS1 irrespective of the host shell. You could probably do more with `eval` here in terms of implementation, but I I think that would add too much complexity.

_What I do **not** like this set up_, is that currently the code interacting with liquidprompt and the code evaluating the results is [separate](https://github.com/felixSchl/liquidprompt/blob/f4339f713e46b1ec4ae47a80bc6cbdfdc3ff603d/tests/liquidprompt.bats#L31-L129), creating a disjoint. What would be nice if we could open a interactive shell in the background and send commands to it using fifo or something.

_What I **do** like about this set up_, is that's very close to being an actual pristine environment (afaict) for every single test case, for every single `run` call even.

dolmen:
Personally I also prefer hard tabs for shell scripting. But my experience as a maintainer when working with multiple casual contributors is that when using tabs you have to spend too much time rejecting patches that have spaces instead of tabs. And using a different indent style in the testsuite from the main code would make the problem even worse.

So the testsuite of liquidprompt should use the same indent style as the liquidprompt source: 4-spaces.

I'll review your PR later.

felixSchl:
Cool. There's currently a lot of noise again, trying to get things to go on travis.yml. I agree with respect to consistency and I will switch it over very soon, before the history would get too messy but one thing should be mentioned: it seems that shell was supposed to be written with hard tabs because there's special support for heredocs, allowing me to indent them rather than glueing them to the left of the file with `<<-EOF` which is really handy. I guess I could still use tabs leading up to the block and then drop them, but a bit of a hassle to work with (in vim at least).

felixSchl:
I drastically simplified the approach and removed plenty of code. Rather than using heredoc, the code a single test lives now in a file in `tests/cases/*`. There's a bats file that would literally just do `@test 'lp: git' { run_shell tests/cases/git.sh }`. The file extension is `.sh` just to help file editors pick it up.

Some notes:

* Travis is set up for my fork, and can be seen here: https://travis-ci.org/felixSchl/liquidprompt
* Tests are run w/o colour atm, and the entire PS1 can be checked for equality this way. We can invest more time into finding a way test colour also.
* There are mocks living in `tests/support/mocked-liquidprompt.sh` - some are defined prior to liquidprompt start, others after. The ones defined prior are required for feature detection or where liquidprompt caches stuff right from the get go, such as the connection type. (note: i have had to make a slight change to liquidprompt: https://github.com/felixSchl/liquidprompt/blob/f88f3fa368b716b00924c19020614cb2507efb24/liquidprompt#L587 to enable this).
* The subshell is launched using `env -i` and only select variables are assigned for it, providing further isolation
* `eval_prompt` calls `_lp_set_prompt` but also rewires the assertion functions so the assertions following are specific to the changes made since the last `eval_prompt` call or since the beginning of the script.
* For the git tests, I am using real git, which means the tests runs a bit slower but we know it works against real git. we could provide mocks here, too, if you think that's better.

I thought we could also generate the bats file on the fly, but maybe a bit too early for that.

Not entirely sure where from here, I think we could just start adding tests

felixSchl:
I've gone ahead and added some path shorten tests, and I found this interesting. Probably this thread is not the place to discuss the ins and outs, but briefly:

```
for t in \
    {0..8}:'/tmp … path' \
    9:'/tmp … /path' \
    10:'/tmp … g/path'
do
    IFS=: read -r ns p <<< "$t"
    for n in $ns; do
        LP_PATH_LENGTH="$n"
        source "$SCRIPT_DIR/tests/support/mocked-liquidprompt.sh"
        eval_prompt "LP_PATH_LENGTH=$n -> $p"
        assert_path "$p"
    done
done
```

will pass on bash, but cause a failure on zsh:

```
   ...check: LP_PATH_LENGTH=9 -> /tmp … /path
   Expected PS1 to equal "0.64 4d [\u:/tmp … /path] 2s \$ " - got: "0.64 4d [\u:/tmp … path] 2s \$ "
   ...check: LP_PATH_LENGTH=10 -> /tmp … g/path
   Expected PS1 to equal "0.64 4d [\u:/tmp … g/path] 2s \$ " - got: "0.64 4d [\u:/tmp … path] 2s \$ "
```

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+  - TEST_CONNECTION=lcl TEST_SHELL=zsh
+
+before_install:
+  - |
+    if [[ "$TRAVIS_OS_NAME" == osx ]]; then
+      brew install bats zsh
+    else
+      sudo add-apt-repository ppa:duggan/bats --yes
+      sudo apt-get update -qq
+      sudo apt-get install -qq -y bats zsh
```
dolmen: Use `addons.apt.packages` instead.
https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-Container-Based-Infrastructure

.gitignore
```diff
@@ -0,0 +1,2 @@
+*.un~
+*.sw[a-z]
```
dolmen: `.sw?` is for Vim files. Adding a `.gitignore` for those files is out of scope for this PR.
I recommend you to instead have a global gitignore (see the core.excludesfile option of git).

What about `*.un~`?

test.sh
```diff
@@ -6,9 +6,9 @@ if [ -z "$BASH_VERSION$ZSH_VERSION" ]; then
     for sh in bash zsh
     do
 	if [ -x /bin/$sh ]; then
-	    /bin/$sh "$0"
+		/bin/$sh "$0"
 	elif [ -x /usr/bin/$sh ]; then
-	    /usr/bin/$sh "$0"
+		/usr/bin/$sh "$0"
 	fi
```
dolmen: Just use bash/zsh from $PATH: `$sh "$0"`

tests/support.bash
```diff
@@ -0,0 +1,76 @@
...
+esac
+
+function setup
+{
+    export TMP_DIR="$BATS_TMPDIR/liquidprompt-tests"
+    rm -rf "$TMP_DIR"
+    mkdir -p "$TMP_DIR"
+
+    # on OSX, we are handed a deep path to some
+    # /var/private/tmp/ directory
```
dolmen: So, why is it a problem?

liquidprompt
```diff
@@ -584,6 +584,7 @@ fi
 # Where are we? #
 #################
 
+if ! type _lp_connection &> /dev/null; then
```
dolmen: This breaks my usual developement workflow: this check stops from updating liquidprompt's code by just reloading it in the same shell.
We should find a better way.

tests/support/assert.sh
```diff
@@ -0,0 +1,80 @@
...
+function eval_prompt
+{
+    if [[ -n "$1" ]]; then
+        echo "...check: $1"
+    fi
+
+    export _LP_PS1 _LP_PS1_PLAIN
+    _lp_set_prompt
+    _LP_PS1="$PS1"
+    _LP_PS1_PLAIN="$(strip_colors <<< "$_LP_PS1")"
```
dolmen: Don't use `_LP_` prefix (and neither `LP_`) for variables in the testsuite but not in liquidprompt.

tests/support/mocked-liquidprompt.sh
```diff
@@ -0,0 +1,53 @@
...
+        for i in eval "echo {1..$DETACHED_JOB_COUNT_SCREEN}"; do
+            echo 'Detach)'
+        done
+    else
+        "$_screen" "$@"
+    fi
+}
+
+
+source "$SCRIPT_DIR/liquidprompt"
```
dolmen: A fixed liquidprompt configuration file should be used.
Currently this testsuite fails on my environment because I have a `~/.liquidpromptrc` with non default settings.

tests/support/mocked-liquidprompt.sh
```diff
@@ -0,0 +1,53 @@
...
+
+DETACHED_SESSION_COUNT_TMUX=2
+DETACHED_SESSION_COUNT_SCREEN=2
+DETACHED_SESSION_COUNT=4
+
+_tmux="$(which tmux)"
+function tmux
+{
+    if [[ "$1" == list-sessions ]]; then
+        for i in eval "echo {1..$DETACHED_JOB_COUNT_TMUX}"; do
```
dolmen: - `i` must be local
- The loop doesn't work:
```
$ for i in eval "echo {1..$x}"; do
> echo $i
> done
eval
echo {1..4}
```

tests/support/mocked-liquidprompt.sh
```diff
@@ -0,0 +1,53 @@
...
+    else
+        "$_tmux" "$@"
+    fi
+}
+
+_screen="$(which screen)"
+function screen
+{
+    if [[ "$1" == -ls ]]; then
+        for i in eval "echo {1..$DETACHED_JOB_COUNT_SCREEN}"; do
```
dolmen: - `i` must be local
- The loop doesn't work

tests/support/mocked-liquidprompt.sh
felixSchl: I think I can achieve this by re-setting "$HOME" during the test run

liquidprompt
felixSchl: we could add a check like `! type _mock_lp_connection` and then make `_lp_connection` call that. But that would introduce test specific code...

tests/support/mocked-liquidprompt.sh
felixSchl: thanks, fixed in later commits (also the one below)

test.sh
felixSchl: Sorry, this was not mine, i just accidentally changed indentation. I reverted my change.

.gitignore
felixSchl: Removed from repo

tests/support.bash
felixSchl: Because path shortening would yield different results based on platform. I will add to the comment

.travis.yml
felixSchl: I've done this, but it appears to be a beta feature on travis and travis is not happy: https://travis-ci.org/felixSchl/liquidprompt/jobs/162850367

nojhan:
This PR is old but a test suite would be very interesting, can I consider merging it or is there something else to change since two years ago?

Rycieos:
I'm closing this PR as the tests are not compatible with the reworked code. 

The point of the rework was to separate data gathering from data display (data vs theme). This had the side effect of making testing the data sources much simpler, which I have done (for most of the data sources) with the new test suite.

As I see it, this PR adds testing for:
* Git
* Path shortening
* Prompt tags (PS1_PREFIX)
* Digital time
* Analog time

None of which (besides Git and somewhat analog time) are data sources. I would be fine with testing on the theme layer, but it would have to be done by mocking the new data source layer. And with the split of data from theme, most of the theme functions are so simple, any test built for them would be exactly duplicating the function they are testing, which defeats the point. (Ex, `_lp_username_color()`. But `_lp_hostname_color()` should probably be tested.)

As for Git (and other VCS providers) I would like to see testing on them, but they can be a bit tricky to get right (as this PR shows). If you fixup the Git tests for the new test suite, I would merge them, but writing VCS tests is not the top priority for me right now.

So thanks for the work you put into this, but unfortunately the state of the project when you wrote this just wasn't good enough to make testing work well.

If you wanted to write some tests for the new test suite, I would be happy to merge them.

Issue #463, opened 2016-07-27
Opening new interactive shell results in recursion, segfault 11


ncs-msmith:
Shell: bash
Operating system: OSX 10.11.6
Liquid Prompt version (tag, commit): 488bb5975db9fa8be0b795019f0b6d99ec792f50

Opening a new interactive shell from a shell with liquidprompt enabled results in recursion, and fails with segfault 11. Unsetting PROMPT_COMMAND and launching an interactive shell works fine.

```
[foobar:~] $ export | grep PROMPT
declare -x PROMPT_COMMAND="_lp_set_prompt"
[foobar:~] $ bash -l
Segmentation fault: 11
[foobar:~] 139 $ export PROMPT_COMMAND=""
[foobar:~] 139 $ bash -l
[foobar:~] $ echo "yay no more error"
yay no more error
[foobar:~] $ exit
logout
[foobar:~] 139 $
```

Some of the output with `set -x` enabled in liquidprompt:

```
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ _lp_set_prompt
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ local -i err=0
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ ((  err != 0  ))
[...]

Issue #450, opened 2016-06-15
_lp_set_prompt error on su


logological:
Operating system: openSUSE 13.2
Liquid Prompt version (tag, commit): dcc18d2454a57e0fb52722530d36b63a59a3efb6

I'm trying out Liquid Prompt for the first time.  The tool works fine from my user account, but whenever I `su` to another account (such as root), the prompt is prepended with an error message about `_lp_set_prompt`:

```
[miller:~] $ su
Password: 
If '_lp_set_prompt' is not a typo you can use command-not-found to lookup the package that contains it, like this:
    cnf _lp_set_prompt
desktop-152:/home/miller # echo

If '_lp_set_prompt' is not a typo you can use command-not-found to lookup the package that contains it, like this:
    cnf _lp_set_prompt
desktop-152:/home/miller # exit
exit
1& [miller:~] 38s $
```

The other accounts don't have any bash dot files so I don't think anything there should be interfering with Liquid Prompt.  Though perhaps openSUSE's `/etc/bash.bashrc` has some incompatibility?

dolmen:
I suspect that the `PROMPT_COMMAND` is exported in your environment. Liquid Prompt doesn't do that, but something else in openSUSE (or your own .bashrc) may do that.

Could you give me the output of this command from a `miller` shell:
```
export | grep PROMPT_COMMAND
```

logological:
Your suspicion seems to be correct:

```
$ export | grep PROMPT_COMMAND
declare -x PROMPT_COMMAND="_lp_set_prompt"
$ fgrep PROMPT_COMMAND .bashrc
export PROMPT_COMMAND="history -a"
```

This was added to my `.bashrc` in hopes of preserving my history across sessions but maybe [the advice to do this](http://unix.stackexchange.com/questions/1288/preserve-bash-history-in-multiple-terminal-windows) wasn't so sound.

dolmen:
At least, remove the `export` keyword.
As `.bashrc` is loaded in every shells (including non-interactive ones) you should set your `PROMPT_COMMAND` in a `[[ $- = *i* ]]` block like Liquid Prompt.

logological:
Yes, that solved the problem.  Thanks for the explanation and advice; feel free to close this issue.

PR #471, opened 2016-09-23
Warn and discard old prompt_command if duplicated

ncs-msmith:
This should prevent errors like #450 and #463 when user's bash environment gets too fancy with $PROMPT_COMMAND.


dolmen:
This has been fixed differently with bea53c056332b0ccb2ffdea333250e5bd0258464 and 18fd3fba035f080abc9d9e64b214dc48c81a6b3d. Thanks anyway!

PR #475, opened 2016-10-07
_lp_color_map has output even if value exceeds scale

jaesivsm:
Notably with the load, no color is displayed by `_lp_color_map` when the value exceeds the provided scale. This PR fixes that by echoing the last item of the color map.


jaesivsm:
just realized this a a duplicate of #455 (which is prettier), closing

PR #477, opened 2016-10-27
Only calculate hash for ssh host color if option is enabled

Gerschtli:
With this change, the checksum of the hostname is only calculated if enabled with the option LP_ENABLE_SSH_COLORS.


Gerschtli:
I am sorry. It seems like I have started off with the wrong branch.. Can somebody tell me, if I need to revert the last rebase to develop?

dolmen:
Replaced by #478.

PR #478, opened 2016-10-27
Only calculate hash for ssh host color if option is enabled

Gerschtli:
Only calculate hash for ssh host color if option LP_ENABLE_SSH_COLORS is enabled.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -286,11 +286,13 @@ _lp_source_config()
     # NO_COL is special: it will be used at runtime, not just during config loading
     NO_COL="${_LP_OPEN_ESC}${ti_sgr0}${_LP_CLOSE_ESC}"
 
-    # compute the hash of the hostname
-    # and get the corresponding number in [1-6] (red,green,yellow,blue,purple or cyan)
-    # FIXME check portability of cksum and add more formats (bold? 256 colors?)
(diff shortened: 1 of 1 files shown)


dolmen:
You're not telling the problem you are trying to fix.
AFAIK there is no problem with SSH_COLORS.

Gerschtli:
It has a performance improvement, because the calculation is useless if the option is not enabled.
The second reason for this is, that I am getting an error in Termux on Android 6.0.1 with ZSH, originated in this calculation:
_lp_source_config:70: number truncated after 9 digits: 2977580043 % 6
Am 31.10.2016 10:27 nachm. schrieb Olivier Mengué <notifications@github.com>:You're not telling the problem you are trying to fix.
AFAIK there is no problem with SSH_COLORS.

dolmen:
Interesting. Could you give me the output of `hostname | cksum` on that host?

chipsenkbeil:
I'm getting this on termux as well. Was there a different solution to this problem?

Gerschtli:
Not afaik

Gerschtli:
Ping

Gerschtli:
Since this change does not seem to be wanted, I close this PR. But feel free to reopen it, if neccessary.

PR #485, opened 2016-11-13
Source liquidprompt only for bash and interactive shell

lhost:


dolmen:
`example.bashrc` already has this line that does the job:
```[[ $- = *i* ]] || return```

Your patch does not improves guarding against non-interactive shells.

Besides that it introduce a bug: there is no reason why `~/.config/liquidprompt/liquidprompt` should be sourced.

**Comment by lhost** [comment]
OK, I should not change path where liquidprompt is installed. But 

```
test  -z "$BASH_VERSION" \
  || test -z "$TERM" -o "x$TERM" = xdumb \
```
is an important part, because it prevents other incompatible shells to load liquidprompt. Yes, these shells uses ~/.bashrc too.

dolmen:
No, the right way to test if the shell is interactive is with `[[ $- = *i* ]]`. Please read the bash manual about `$-`.
And the part about `x$TERM = xdumb` is already the first line of `liquidprompt` itself.

> [...] is an important part, because it prevents other incompatible shells to load liquidprompt.

No other shell than bash should parse `~/.bashrc`. And AFAIK, none does. Unless you have a broken setup.

PR #488, opened 2016-11-18
Updated default stash char to avoid double-click select of plus sign.

### LINKED ISSUE #487: default stash character (+) gets included in double-click highlight of branch
Shell: bash
Operating system: OSX
Liquid Prompt version (tag, commit): 9c1c8a3

stash

I propose changing the default to a character that is not automatically highlighted in this way, such as `:`

stash2

dogeared:
Fixes #487


dolmen:
I rejected #487.

PR #496, opened 2016-12-06
Add support for AWS_PROFILE and AWS_DEFAULT_PROFILE environment variables

sagepe:
This is useful when working with multiple AWS accounts over the API or CLI. This simple addition will ensure `AWS_DEFAULT_PROFILE` is displayed if set.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -228,6 +228,20 @@ Features
 
    See also: :attr:`LP_MARK_DISABLED`.
 
+.. attribute:: LP_ENABLE_AWS_PROFILE
+   :type: bool
+   :value: 1
+
+   Display the current value of :envvar:`AWS_PROFILE` or 
+   :envvar:`AWS_DEFAULT_PROFILE`. These variables are used to switch between
+   configuration profiles by the `AWS CLI`_.
+
+   .. _`AWS CLI`: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html
+
+   See also: :attr:`LP_COLOR_AWS_PROFILE`.
+
+   .. versionadded:: 2.1
+
 .. attribute:: LP_ENABLE_BATT
    :type: bool
    :value: 1
@@ -889,6 +903,16 @@ Valid preset color variables are:
 
    See also: :attr:`LP_ENABLE_BATT`, :attr:`LP_ENABLE_LOAD`, and
    :attr:`LP_ENABLE_TEMP`.
+
+.. attribute:: LP_COLOR_AWS_PROFILE
+   :type: string
+   :value: $YELLOW
+
+   Color used to display the current active AWS Profile.
+
+   See also: :attr:`LP_ENABLE_AWS_PROFILE`.
+
+   .. versionadded:: 2.1
 
 .. attribute:: LP_COLOR_CHANGES
    :type: string
--- a/docs/functions/theme.rst
+++ b/docs/functions/theme.rst
@@ -90,6 +90,12 @@ specific text and formatting may change.
 
    .. versionadded:: 2.0
 
+.. function:: _lp_aws_profile_color() -> var:lp_aws_profile_color
+
+   Returns :func:`_lp_aws_profile` with color from :attr:`LP_COLOR_AWS_PROFILE`.
+
+   .. versionadded:: 2.1
+
 .. function:: _lp_battery_color() -> var:lp_battery_color
 
    Returns data from :func:`_lp_battery`, colored with:
--- a/liquidprompt
+++ b/liquidprompt
@@ -216,6 +216,7 @@ __lp_source_config() {
     LP_ENABLE_RUNTIME_BELL=${LP_ENABLE_RUNTIME_BELL:-0}
     LP_ENABLE_VIRTUALENV=${LP_ENABLE_VIRTUALENV:-1}
     LP_ENABLE_SCLS=${LP_ENABLE_SCLS:-1}
+    LP_ENABLE_AWS_PROFILE=${LP_ENABLE_AWS_PROFILE:-1}
     LP_ENABLE_VCS_ROOT=${LP_ENABLE_VCS_ROOT:-0}
     LP_ENABLE_TITLE=${LP_ENABLE_TITLE:-0}
     LP_ENABLE_SCREEN_TITLE=${LP_ENABLE_SCREEN_TITLE:-0}
@@ -292,6 +293,7 @@ __lp_source_config() {
     LP_COLOR_VIRTUALENV=${LP_COLOR_VIRTUALENV:-$CYAN}
     LP_COLOR_DIRSTACK=${LP_COLOR_DIRSTACK:-$BOLD_YELLOW}
     LP_COLOR_KUBECONTEXT=${LP_COLOR_KUBECONTEXT:-$CYAN}
+    LP_COLOR_AWS_PROFILE=${LP_COLOR_AWS_PROFILE:-$YELLOW}
 
     if [[ -z "${LP_COLORMAP-}" ]]; then
         LP_COLORMAP=(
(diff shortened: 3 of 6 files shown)


dolmen:
I'm sorry but LiquidPrompt implementation is not flexible enough to allow to support any additional development environment variable without adding a runtime cost.

sagepe:
Thanks for taking the time to review and comment, @dolmen. It's something I need and use so had added to my fork and thought I'd at least send over a PR in case it was of interest.

Rycieos:
[Release Candidate v2.0.0-rc.1](https://github.com/nojhan/liquidprompt/releases/tag/v2.0.0-rc.1) is now out, which means that the rework is complete, which means the runtime cost of new features is greatly reduced! I would merge this, but first this has merge conflicts that need to be fixed.

sagepe:
How's this?

docs/config.rst
```diff
@@ -194,6 +194,15 @@ Features
 
    See also: :attr:`LP_MARK_DISABLED`.
 
+.. attribute:: LP_ENABLE_AWS_PROFILE
+   :type: bool
+   :value: 1
+
+   Display the current value of the ``AWS_PROFILE`` or ``AWS_DEFAULT_PROFILE``
```
Rycieos, line 201: Prefix these with `:envvar:`, like this:
```suggestion
   Display the current value of the :envvar:`AWS_PROFILE` or
   :envvar:`AWS_DEFAULT_PROFILE`.
```

It won't display any differently, but it will index and search better.

docs/config.rst
```diff
@@ -194,6 +194,15 @@ Features
...
+   environment variables if they are set.
```
Rycieos, line 202: This would be the place to add link(s) to the tools that this is displaying.

I'm assuming this is for the AWS CLI, so maybe his link? https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
```suggestion
   environment variables if they are set. These variables are used to
   configure the `AWS CLI`_.
   
   .. _`AWS CLI`: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html
```

docs/functions/theme.rst
```diff
@@ -90,6 +90,12 @@ specific text and formatting may change.
 
    .. versionadded:: 2.0
 
+.. function:: _lp_aws_profile() -> var:lp_aws_profile
+
+   Returns :func:`_lp_analog_time` with color from :attr:`LP_COLOR_AWS_PROFILE`.
```
Rycieos, line 95:
```suggestion
   Returns :func:`_lp_aws_profile` with color from :attr:`LP_COLOR_AWS_PROFILE`.
```

liquidprompt
```diff
@@ -1289,6 +1291,28 @@ _lp_dirstack_color() {
...
+    (( LP_ENABLE_AWS_PROFILE )) || return 2
+
+    local ret
+
+    if [[ -n "${AWS_PROFILE-}" ]]; then
+        __lp_escape "${AWS_PROFILE##*/}"
+        lp_aws_profile=$ret
+    elif [[ -n "${AWS_DEFAULT_PROFILE-}" ]]; then
+        __lp_escape "${AWS_PROFILE##*/}"
+        lp_aws_profile=$ret
```
Rycieos, line 1304: This whole block could be simplified a bit to reduce duplication:
```suggestion
    local aws_profile="${AWS_PROFILE-${AWS_DEFAULT_PROFILE-}}"
    if [[ -n $aws_profile ]]; then
        __lp_escape "${aws_profile##*/}"
        lp_aws_profile=$ret
```

liquidprompt
```diff
@@ -1289,6 +1291,28 @@ _lp_dirstack_color() {
     lp_dirstack_color="${LP_COLOR_DIRSTACK}${LP_MARK_DIRSTACK}${lp_dirstack}${NO_COL}"
 }
 
+_lp_aws_profile() {
+    (( LP_ENABLE_AWS_PROFILE )) || return 2
+
+    local ret
+
+    if [[ -n "${AWS_PROFILE-}" ]]; then
+        __lp_escape "${AWS_PROFILE##*/}"
```
Rycieos, line 1300: What is the reasoning behind removing everything before the final slash? Is it just copying what the Python env data function does? I'd like to see a comment explaining why.

liquidprompt
```diff
@@ -1289,6 +1291,28 @@ _lp_dirstack_color() {
...
+        lp_aws_profile=$ret
+    else
+        return 1
+    fi
+}
+
+_lp_aws_profile_color() {
+    _lp_aws_profile || return "$?"
+
+    lp_aws_profile_color="[${LP_COLOR_AWS_PROFILE}AWS:${lp_aws_profile}${NO_COL}]"
```
Rycieos, line 1313: This `AWS:` string really should be turned into a "mark" configuration option, probably named `LP_MARK_AWS_PROFILE`.

Rycieos (changes requested):
A few bugs, a few small nitpicks, and a few documentation clarifications.

This in general looks good, other than those fixes.

FYI, the merge window for v2.0 is closed (now just waiting on bug reports), so I can't merge this until v2.0 is fully released. Which means your `versionadded` lines need to say `2.1`. And there were a few places in the documentation that you missed adding a `.. versionadded::` block, double check those.

liquidprompt
sagepe: Hah, I think it was probably copied from there, all those years ago. It serves no purpose here as it does for the virtualenv so I'll remove that.

liquidprompt
```diff
@@ -2730,7 +2758,7 @@ _lp_default_theme_prompt_template() {
         # add user, host and permissions colon
         PS1+="${LP_BRACKET_OPEN}${LP_USER}${LP_HOST}${LP_PERM}"
 
-        PS1+="${LP_PWD}${LP_DIRSTACK}${LP_BRACKET_CLOSE}${LP_SCLS}${LP_VENV}${LP_PROXY}"
+        PS1+="${LP_PWD}${LP_DIRSTACK}${LP_BRACKET_CLOSE}${LP_SCLS}${LP_AWS_PROFILE}${LP_VENV}${LP_PROXY}"
```
Rycieos, line 2761: Here you have `LP_AWS_PROFILE` before `LP_VENV`, but in `docs/theme/default.rst` you have it after. The order needs to be the same, as specified in that documentation:

> All of the available template sections are listed below. Their order is the default order if the user does not configure a different template.

I don't care which one comes first personally, just make them the same.

You also added it to the `liquid.ps1` file. You don't need to do that, see the comment in that file:
> This file is not updated with new template sections.

But if you do, make sure it is the same order as here.

docs/functions/theme.rst
```diff
@@ -90,6 +90,12 @@ specific text and formatting may change.
 
    .. versionadded:: 2.0
 
+.. function:: _lp_aws_profile() -> var:lp_aws_profile
```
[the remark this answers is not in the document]
sagepe: Yes, quite right. Fixed.

liquidprompt
```diff
@@ -1289,6 +1291,28 @@ _lp_dirstack_color() {
...
+_lp_aws_profile() {
+    (( LP_ENABLE_AWS_PROFILE )) || return 2
+
+    local ret
+
+    if [[ -n "${AWS_PROFILE-}" ]]; then
+        __lp_escape "${AWS_PROFILE##*/}"
+        lp_aws_profile=$ret
+    elif [[ -n "${AWS_DEFAULT_PROFILE-}" ]]; then
+        __lp_escape "${AWS_PROFILE##*/}"
```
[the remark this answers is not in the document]
sagepe: Not needed following the simplification you suggest below.

docs/config.rst
sagepe: Done.

docs/config.rst
sagepe: I've added a slightly different link that lands on the AWS documentation about named profiles, this seems more targetted.

docs/functions/theme.rst
sagepe: Apologies for the careless cut-and-paste error - I should have spotted this!

liquidprompt
sagepe: Good call, done.

liquidprompt
sagepe: You are right, it should but I have decided to remove this altogether; I added it as an afterthought and returning to it I think it is unnecessary and clutters the prompt.

If there was a desire for a marker, one could be added in a separate commit at a later stage.

liquidprompt
sagepe: I've removed this from `liquid.ps1` and standardised the order elsewhere.

sagepe:
First, apologies for the radio silence for the last month.

I think I've responded to all your feedback and noted bits in the relevant conversations above - I hope this it now in better shape!

I've rebased against master and squashed everything together into a single commit.

Rycieos (approved):
> apologies for the radio silence for the last month.

No worries! This PR sat for 4 years, so another month is a drop in the bucket. Sorry it took someone so long to get to it.

Thanks for the fixups! This looks really good, I have no more issues with it.

Issue #399, opened 2015-10-16
Add an option to display the branch but not the status information.


pjvandehaar:
Could you add an option to show the current branch without any status information? This is simply for (eg network) drives with slow IO and big git repos.

Like this:

``` bash
_lp_git_branch_color()
{
    local branch="$(_lp_git_branch)"
    [[ -n "${branch}" ]] && echo -ne "${LP_COLOR_UP}${branch}$NO_COL"
}
```

Alternatively, could this be displayed under paths in `LP_DISABLED_VCS_PATH`?

dolmen:
Note that you already have the solution for yourself: in your `~/.bashrc`/`zshrc` you can redefine the `_lp_git_branch_color` function as you suggest just after loading LiquidPrompt.

Rycieos:
Why would this be a better solution than using `LP_DISABLED_VCS_PATH`? The proposed option would disable it everywhere, where the vcs path would allow turning it off for only the slow network directories.

I would instead propose making `LP_DISABLED_VCS_PATH` disable all status information while keeping branch name on.

Rycieos:
I just merged #635, which adds a separate theme `alternate_vcs` that only modifies the default theme's VCS display. Check out the implementation of that new theme in `themes/alternate_vcs/` as an example. The new theming engine in v2 makes it much simpler than before to make the modification you want to only display a branch name. You could probably make a theme that includes only this:

```bash
_lp_my_theme_theme_prompt() {
    _lp_default_theme_prompt_data

    if _lp_find_vcs && _lp_vcs_branch; then
        LP_VCS=" $lp_vcs_branch"
    else
        LP_VCS=
    fi

    _lp_default_theme_prompt_template
}
```
Of course [`_lp_vcs_branch()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.2/functions/data/vcs.html#lp_vcs_branch) could return nothing if no branch is checked out, so I would recommend looking at how the default theme does branch detection (in `_lp_vcs_details_color()`) and maybe copy that.

As for

> I would instead propose making LP_DISABLED_VCS_PATH disable all status information while keeping branch name on.

I did not implement this. It would need another feature toggle, and would make more work for theme creators. I also think that older VCSs like SVN and Bazaar would still be too slow in large repos to just print a branch name. And like I mentioned above, I think you would need multiple calls still in case there was no current branch. I'm still open to the feature, just need to see more interest and discussion on it before I would implement it.

PR #505, opened 2017-02-18
Add an option to display the current git branch only.

laurentbardin:
This is a first draft fix for #399. It has a few drawbacks and could be enhanced if needed:

1. if we wanted to implement the mechanism for every supported VCS, it would add one variable each time;
2. we could use one variable for every VCS, something like `LP_ENABLE_VCS_BRANCH_ONLY`, but then we lose fine-grained control over it.

I don't think that second point really matters, as the problem mentioned in the issue (slow I/O and big repo) is bound to happen with every VCS at some point.

As for the patch itself, I'm really not fluent with bash so feel free to point any mistake. It looks messy, but it's mostly changes in indentation.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -876,7 +876,7 @@ _lp_git_head_status() {
     fi
 }
 
-if [[ LP_ENABLE_GIT_BRANCH_ONLY ]]; then
+if (( LP_ENABLE_GIT_BRANCH_ONLY )); then
     # Only display the current branch, in green by default.
     _lp_git_branch_color()
     {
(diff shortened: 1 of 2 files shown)


laurentbardin:
Fixed a conflict introduced by #508.

dolmen:
This implementation is not efficient. This is the whole function definition that should be overriden if `LP_ENABLE_GIT_BRANCH_ONLY` is set.

Rycieos:
Closing as this needs work, a rebase, should now be done with themes, and there have been no updates for years.

Issue #506, opened 2017-02-21
Show git tag in detached state


cburschka:
Shell: bash/zsh
Operating system: Linux / MacOS
Liquid Prompt version (tag, commit): 47126c7 

Currently _lp_git_branch() shows a branch if possible, and otherwise falls back on a short commit hash.

It would be nice to see a tag, if one is available. `git describe --tags --exact-match` does exactly that - though the error needs to be suppressed if no tag exists. If that command returns an error, fall back on the commit hash.

Rycieos:
[v2.0.0-beta.1](https://github.com/nojhan/liquidprompt/releases/tag/v2.0.0-beta.1) has a new function supporting this: [`_lp_vcs_tag()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.1/functions/data/vcs.html#lp_vcs_tag). See the link for documentation.

The default theme now falls back to a tag if a branch is not found. See [`_lp_vcs_details_color()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.1/functions/theme.html#lp_vcs_details_color) for details.

The details of the implementation are here:
https://github.com/nojhan/liquidprompt/blob/6e7f4c77044bea7ab92970e820997601926bcbea/liquidprompt#L1211-L1224

Thanks to @cburschka and @LlinksRechts for the implementations which helped me get this working.

PR #507, opened 2017-02-21
Display git tag, if one is available. (#506)

cburschka:
For #506

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -854,6 +854,7 @@ _lp_git_branch()
     if branch="$(\git symbolic-ref -q HEAD)"; then
         _lp_escape "${branch#refs/heads/}"
     else
+        \git describe --tags --exact-match 2>/dev/null && return
         # In detached head state, use commit instead
         # No escape needed
         \git rev-parse --short -q HEAD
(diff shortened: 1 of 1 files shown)


Rycieos:
One issue I see is that there is no way to differentiate between a branch name and a tag name. Since there is no limitation in git to prevent having a tag and a branch named the same thing, this could create confusion with which one is currently checked out.

Git does tell you that you are in a detached head state when checking it out, so as long as one remembers their context they can avoid confusion, but I don't think we should depend on that.

I would suggest adding a "tag: " before the tag name. This is how git does it in `git log`, and it makes it obvious it isn't a branch. Then this would be a feature I would want.

LlinksRechts:
addressed in pr #533

nojhan:
The "tag" text should be configurable with something like LP_MARK_VCS_TAG, and defaults to a shorter unicode character, like u1F516: 🔖.

pichi:
Check out #567 which address also detached tags.

Rycieos:
As commented on https://github.com/nojhan/liquidprompt/issues/506#issuecomment-741997599, this has been added in [v2.0.0-beta.1](https://github.com/nojhan/liquidprompt/releases/tag/v2.0.0-beta.1).

PR #512, opened 2017-03-07
Allow liquidpromptrc in ~/.config/liquidprompt/

j-degreef:
This feature allows to put liquidpromptrc in the ~/.config/liquidprompt/ directory. We then can have all configuration files (rc, .theme, .PS1) in one unique directory called "liquidprompt" under ~/.config/.

This is my first pull request in Github, I hope I've made everything correctly.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -430,6 +430,9 @@ _lp_source_config()
             if [[ -f "$first/liquidpromptrc" ]]; then
                 configfile="$first/liquidpromptrc"
                 break
+            elif [[ -f "$first/liquidprompt/liquidpromptrc" ]]; then
+                configfile="$first/liquidprompt/liquidpromptrc"
+                break
             fi
         done
     fi


j-degreef:
So, no one is interested by my little PR ?
In CONTRIBUTING.md it's said that we can post a "ping comment" if we don't get a review but I must say that I don't know what that actually means.

Rycieos:
I'll leave my thoughts:

I think this is a good idea, having every type of config under .config/<program-name>/ makes good organizing sense. Liquidprompt is one of the only programs I use that doesn't conform to this standard.

This PR does it well, it makes it the backup option if the current way isn't used.

Point for thought: does the RC file still need to be called "liquidpromptrc" when it is in the directory? Is there a way to name it something else, like "config", or do we need to keep to our standard?

Only other thing would be making sure this is documented in the README.

j-degreef:
@Rycieos 
Thanks for your reply.

I think it's better to keep the linuxpromptrc name for backward compatibility and to not confuse the users. It's just a name but t says what it is : a rc file for linuxprompt.
I'm not sure if there is any standard to name a rc file in XDG_CONFIG_DIRS/ProgDir/
, i3 one is named config, Ranger one is named rc.conf, NeoVim init.vim and so on...

I'll have a look at how to adapt the README and attach it to this PR.

dolmen:
My only concern is that this change do not simplify either maintainer's or user's life: it *adds* a *third* location for the configuration file. This will not help users to find where they put their config 6 months before.

Rycieos:
@dolmen does bring up a good point, which I would agree with. Adding confusing locations is never good.

The other option would be to change the current setup from just .config/ to .config/liquidprompt/, which would break things. It would be ideal though.

j-degreef:
@Dolmen @Rycieos 
If it was up to me I would go for the solution with everything in .config/liquidprompt/. At the top the rc file it's already said to put .theme and .PS1 in this directory (even if they coud be sourced frpom somewhere else).
I'm using this solution since a few weeks after changing the code and I've not seen that something is broken but I assume I've not tested all possible configurations especially using SSH, tmux.
By the way this could be a slow transition as my code handles all other locations for the rc file.

Rycieos:
By broken, I meant that changing the only location without fair warning would cause people updating to no longer be loading their config, and not know why.

While I still think that @j-degreef's proposed location is better than what we have now, I think I will agree with @dolmen and say that this might cause too much confusion.

However, I would very much like to see this change, and it would be a perfect candidate for if liquidprompt ever gets any sort of a rewrite.

j-degreef:
What about merging it as a hidden feature ?
Only people that read the code, as I did, will understand that they can put everything under .config/liquidprompt/.
Plus that will avoid me to report this change every time liquidprompt is updated. Well, ok, that not a good argument ;) 

Concerning a rewrite, I've been thinking about that recently. Currently everything is more or less mixed, colors, prompt building and data collection. For instance it's not currently possible to put brackets around the time without changing the code. Or change bracket color. Or use colors out of the provided list (I have created a  set_color function in my liquid.theme that goes around this).
We can put ${LP_TIME} in our .PS1 file but disable it by LP_ENABLE_TIME=1 in liquidpromptrc which is quite confusing.
As I see it, a rewrite would completely split the 3 main parts : 
1. .theme that handles colors and special characters
2. prompt definitions (.PS1) that handles the prompt itself (what to include, what to exclude and in which order)
3. the bin (liquidprompt) that handles data collection and then build the prompt according to the .theme and the prompt definition (.PS1)

This is my 0.50¢ contribution and don't take it wrong, liquidprompt is already amazing ;)

Rycieos:
While this might not be the right place to discuss a rewrite, I know that there are performance problems that are often talked about. A rewrite would have to take performance into consideration.

I will say that I have thought about a rewrite before, but there are few problems that would require a rewrite, and being such an old project, many portability bugs have been fixed, which could be undone with a rewrite.

dolmen:
About a liquidprompt rewrite, see my own project [angel-PS1](/dolmen/angel-PS1) that I started months before taking over maintenance of LiquidPrompt: it has a more general scope: not just a shell prompt, but a shell prompt *engine*. I even have work in (slow) progress for a plugin to read LiquidPrompt config and implement it. The [`devel` branch](/dolmen/angel-PS1/tree/devel) is the one that has the angel-PS1 source (`release` is the result of a build).

dolmen:
@j-degreef 
> What about merging it as a hidden feature ?

A feature hidden is still a feature and people will rely on it. Once it is in we have to support it forever to allow smooth unattended upgrades (such as when liquidprompt is packaged by your OS distribution).

> Only people that read the code, as I did, will understand that they can put everything under .config/liquidprompt/.

Many many people read LiquidPrompt's code: the amount of pull requests show it. So nothing is hidden.

>Plus that will avoid me to report this change every time liquidprompt is updated. Well, ok, that not a good argument ;)

The best thing is to maintain your own fork. This is very easy with Git: just use `git rebase` when I push new commits. Some other users do that. Tell me if you need more help to setup this.

j-degreef:
@dolmen 
You're right.
I had a look at angel-PS I like the idea of separated modules for each functionality.
If I have a problem rebasing my tree, I'll let you know.
Thanks for your help.

Rycieos:
I'm going to have to agree and reject this feature. We already have 4 possible config locations:
`~/.liquidpromptrc`
`$XDG_CONFIG_HOME/liquidpromptrc`
`$XDG_CONFIG_DIRS/liquidpromptrc`
`/etc/liquidpromptrc`

Adding another would simply be too confusing. And as previously mentioned, there is no way to remove a previous location without breaking support for users.

PR #523, opened 2017-05-31
Strip non-numeric suffix from minor bash version (#522)

### LINKED ISSUE #522: Bash version detection fails for non-numeric minor versions (e.g. bash 3.05b.*)
Shell: bash 3.05b.*
Operating system: any
Liquid Prompt version (tag, commit): master, 9c80396021a8106bfaeade9a1ea51b85152e951d

The initial bash version detection fails for BASH_VERSIONs with a minor version containing a non-numeric suffix, e.g. `BASH_VERSION="3.05b.0(1)-release"`.
Such a BASH_VERSION results in a bash error:
```
-bash: 05b: value too great for base (error token is "05b")
```

Stripping a non-numeric suffix from `bminor` in `liquidprompt` resolves the issue.

flopp:
This fixes #522 by removing any non-numeric suffix from `bminor` before the actual version check.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -55,7 +55,7 @@ test -z "$TERM" -o "x$TERM" = xdumb && return
 
 # Check for recent enough version of bash.
 if test -n "${BASH_VERSION-}" -a -n "$PS1" ; then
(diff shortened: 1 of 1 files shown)


nojhan:
Merged in develop, thanks!

PR #533, opened 2017-09-29
showing tag instead of commit hash in detached state if applicable

LlinksRechts:
as suggested in issue #506


Rycieos:
Why open a new PR? What was wrong with the old one, this new one has the exact same changes proposed.

LlinksRechts:
apparently I overlooked that there already was one, sorry about that

PR #536, opened 2017-12-22
Fixed battery reporting

overhacked:
The sed script was having some issues with the way `pmset -g batt` output looks on macOS 10.12 Sierra and later. I rewrote the sed commands to an awk script, which is a bit more resistant to any future format changes.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1314,7 +1314,7 @@ case "$LP_OS" in
             *)  # "charging", "AC attached"
                 echo -nE "$percent"
                 # under => 2, above => 3
-                return $(( 1 + ( percent > LP_BATTERY_THRESHOLD ) ))
+                return $(( 2 + ( percent > LP_BATTERY_THRESHOLD ) ))
             ;;
         esac
     }
--- a/liquidprompt
+++ b/liquidprompt
@@ -1301,7 +1301,7 @@ case "$LP_OS" in
     {
         (( LP_ENABLE_BATT )) || return 4
         local percent batt_status
-        eval "$(pmset -g batt | sed -n 's/^ -InternalBattery[^	 ]*[	 ]\([0-9]*[0-9]\)%; \([^;]*\).*$/percent=\1 batt_status='\'\\2\'/p)"
+        eval "$(pmset -g batt | awk '/InternalBattery/{gsub(/[^0-9]/,"",$3);gsub(/[;]/,"",$4);printf "percent=%d batt_status=%s", $3, $4}')"
         case "$batt_status" in
             charged | "")
             return 4


augmentedfourth:
#476 was requested over a year ago and never got pulled. Good luck with this one.

overhacked:
This change is more robust than the change in #476, because it doesn't depend on the specific formatting of numbers and whitespace of the `InternalBattery` line, only that the third and fourth fields (space-delimited) are percent and status. This is the same on older versions, so this should be a backwards-compatible change.

augmentedfourth:
If you're going to fix this, the same off-by-one error also exists in line 1295 (on the Linux code for this same battery check).

Rycieos:
@overhacked I disagree that this change is more robust.

[dolmen has said before](https://github.com/nojhan/liquidprompt/issues/461#issuecomment-233940647) that using `awk` is frowned upon since it is much slower than `sed` or other alternatives. The change from `sed` to `awk` here isn't needed.

Your point that "it doesn't depend on the specific formatting of numbers and whitespace of the InternalBattery line" is false, since your usage of `$3` and `$4` is directly dependent on whitespace. According to [this StackOverflow answer](https://stackoverflow.com/a/30406868) for how `awk` parses whitespace to separate fields:
> any run of spaces and/or tabs and/or newlines is treated as a field separator with leading and trailing runs ignored.

The other PR has solved this in the same way, using the POSIX `[[:space:]]`, which [according to the manual](https://www.gnu.org/software/sed/manual/sed.txt):
> in the 'C' locale, this is tab, newline, vertical tab, form feed, carriage return, and space.

The real issue *is* exactly how many fields there are, since that is what broke before. We don't have a record of what `pmset` output before, but now it looks like this:
```
Now drawing from 'AC Power'
 -InternalBattery-0 (id=2228323)    100%; charged; 0:00 remaining present: true
```
Based on what the old `sed` command was, my guess is that the line used to look like:
```
-InternalBattery-0    100%; charged; 0:00 remaining present: true
```
and they added the ID field, which changed what number the percent and status fields were. If they do that again, your `awk` script would break. The `sed` script in the other PR would not break, since it is looking for the percent sign (%).

Of course there are a multitude of things that Apple could change in the output, and they are hard to prepare for. A simple one (that could already be possible) would be for them to output a line for each physical battery:
```
Now drawing from 'AC Power'
 -InternalBattery-0 (id=2228323)    100%; charged; 0:00 remaining present: true
 -InternalBattery-1 (id=2228324)    100%; charged; 0:00 remaining present: true
```

The `sed` script still works:
```
percent=100 batt_status='charged'
percent=100 batt_status='charged'
```
But will only show the stats for the last listed battery.

The `awk` script:
```
percent=100 batt_status=chargedpercent=100 batt_status=charged
```
Still kinda works, but will show the percent of the first battery and the status of the last battery.

I would of course love for this feature to become more robust, but that is difficult, and I'm not it's worth anyone's time, since we don't know what changes will or could be coming.

Issue #543, opened 2018-03-16
Disabling pink error code


atsaloli:
Shell: bash
Operating system: RHEL 6
Liquid Prompt version (tag, commit): 9c80396 (2017-04-04)

Is there any way to disable the pink reporting of the exit status code of the last command when non-zero?

I looked in the config file and in the `liquid.ps1` file but I didn't see anything... did I miss it, or where is it?

Thanks! Love the product.  =)

Rycieos:
You would have to use a [theme](https://github.com/nojhan/liquidprompt#themes) to change default colors. If you copy the default theme file, the line you are looking for is the [LP_COLOR_ERR](https://github.com/nojhan/liquidprompt/blob/9c80396021a8106bfaeade9a1ea51b85152e951d/liquid.theme#L73).

augmentedfourth:
@atsaloli If you want to get rid of the whole thing, you can comment out lines 1702-1707 in the main liquidprompt script. Even better, put a conditional around those lines and add a config parameter to turn it on/off, then send in a pull request. 😉

atsaloli:
Thanks, @augmentedfourth! It's a small world! =)  

And thanks @Rycieos!  =)

augmentedfourth:
@nojhan Was that added in another pull request? #544 was closed without merging due to @Rycieos stating that we shouldn't be adding new parameters like `LP_ENABLE_ERR`. The official solution was to take `${LP_ERR}` out of the theme file rather than add the new `.liquidpromptrc` option.

nojhan:
I was unaware of this discussion, I added the feature without a PR (I'm not even sure I can do a PR toward my own repo?). To avoid such mistake, this kind of thought should definitely be written down as comment in the code, we cannot expect contributors to read all closed PRs.
Additionally, if the config flags number is a problem, maybe it would be better to remove them entirely in favour of a cleaner LP_PS1_FILE system?

Rycieos:
@nojhan you can do a PR against your own repo; the benefit being other contributors can read and give suggestions before merging.

I agree that it would be better to have some sort of modular configuration system, where instead of checking configuration variables, each construction function was called only if it was wanted.

Rycieos:
[...]

PR #544, opened 2018-03-24
set up LP_ENABLE_ERR to switch non-zero exit reporting on/off

augmentedfourth:
Addresses #543.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -326,6 +326,7 @@ _lp_source_config()
     LP_ENABLE_HG=${LP_ENABLE_HG:-1}
     LP_ENABLE_BZR=${LP_ENABLE_BZR:-1}
     LP_ENABLE_TIME=${LP_ENABLE_TIME:-0}
+    LP_ENABLE_ERR=${LP_ENABLE_ERR:-1}
     LP_ENABLE_RUNTIME=${LP_ENABLE_RUNTIME:-1}
     LP_ENABLE_VIRTUALENV=${LP_ENABLE_VIRTUALENV:-1}
     LP_ENABLE_SCLS=${LP_ENABLE_SCLS:-1}
@@ -1701,7 +1702,9 @@ _lp_set_prompt()
     # As this get the last returned code, it should be called first
     local -i err=$?
     if (( err != 0 )); then
-        LP_ERR=" $LP_COLOR_ERR$err$NO_COL"
+        if (( LP_ENABLE_ERR )); then
+            LP_ERR=" $LP_COLOR_ERR$err$NO_COL"
+        fi
     else
         LP_ERR=''     # Hidden
     fi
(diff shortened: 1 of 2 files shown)


augmentedfourth:
I know that this doesn't explicitly set `LP_ERR=''` if `err != 0` and `LP_ENABLE_ERR` is disabled, but if `LP_ENABLE_ERR` is disabled, `LP_ERR` will never be populated so it won't need to be hidden or returned to an empty string.

Rycieos:
Is there no way to put both of the checks inside the same `if`? That would make it cleaner and possibly faster.

I'm not sure that @dolmen will want to merge this; he has been quite adamant about not adding new config options, especially since this specific case can be done easily through a .ps1 file.

augmentedfourth:
A .ps1 file can remove it entirely? As far as I could tell, that can only change the color.

I tried getting the conditionals merged into a single statement, but since the preferred conditional style throughout has been the double-parentheses I couldn't get it to work. I can definitely make it work with square brackets, but was trying to maintain consistency.

augmentedfourth:
Ah, got it. @atsaloli, is this solution OK for you?

atsaloli:
Works beautifully!  Thank you!  :+1: :heart: :fireworks: :unicorn: :shell:

atsaloli:
Actually, my prompt is broken now in that it no longer shows the path of the working directory.

atsaloli:
And sorry I took a while to reply... I've been traveling.  =)   (That's also why I did not dig into it myself like you suggested.)

augmentedfourth:
If all you've changed from the default is to remove the error message, line 44 of `liquid.ps1` should say:

`LP_PS1="${LP_PS1}${LP_RUNTIME}${LP_MARK_PREFIX}${LP_MARK}${LP_PS1_POSTFIX}"`

Has anything else changed from the release version of `liquid.ps1`?

atsaloli:
I applied the diff in this pull request, which updated two files:  liquidprompt, and liquidpromptrc-dist

It's been working fine ever since, just the path froze in the prompt originally right after the change (i.e., wouldn't update even if I changed dirs).  It works fine now, I've been using it nearly every day. I've been telling my friends about it too.  =)  Good stuff!

augmentedfourth:
As @Rycieos noted, this pull request won't be merged because it adds a new option to `.liquidpromptrc` and that's become frowned upon (I wasn't aware of that when I submitted this). You're free to keep this diff applied to your local copy, but you'll need to rebase the change if you want to get new official updates. Apparently the "official" answer is to edit your `.ps1` theme file to not include `${LP_ERR}` in the prompt definition, unless that actually doesn't address your concern somehow.

atsaloli:
Well! OK, thank you.  =)

atsaloli:
I set up a custom ps1 theme file, and edited it to remove LP_ERR and reverted the liquidprompt bash code back to master.  Works great, thanks!

Issue #516, opened 2017-04-19
Bash 4.4.11 Command time and return status


izissise:
Shell: bash 4.4.11
Operating system: Debian
Liquid Prompt version (tag, commit): 93df0169499c4ca8563add267e95dcd343d95f12

Hello,
When using liquid prompt with bash 4.4.11, the time taken by a command doesn't work properly, it is always displayed and it shows the time since the shell was launched.
Also the return code of command doesn't shows even for non-zero return status.

nolanl:
I see the same 2 problems, Also Debian, bash version 4.4.12.

vabada:
I have the same problem! Any way we can help to fix it?

Rycieos:
We will probably need some example output of it not working as intended. Running a few commands with the wrong output would help. As it is right now we don't have enough info to start figuring out what is wrong.

Relevant code sections:
https://github.com/nojhan/liquidprompt/blob/9c80396021a8106bfaeade9a1ea51b85152e951d/liquidprompt#L1408-L1420

https://github.com/nojhan/liquidprompt/blob/9c80396021a8106bfaeade9a1ea51b85152e951d/liquidprompt#L1437-L1459

@nolanl and @vabada, we could use your shell and shell version info, as well as confirming that your liquidprompt is up to date.

nolanl:
@Rycieos Thanks for taking a look.

```
[:~/bin/upstream/liquidprompt] master ± git pull
Current branch master is up to date.

[:~] $ echo $BASH_VERSION
4.4.12(1)-release

[:~] $ false
[:~] $ true
[:~] $ 

[:~] $ grep RUNTIME .config/liquidpromptrc 
# Show runtime of the previous command if over LP_RUNTIME_THRESHOLD
LP_ENABLE_RUNTIME=1
LP_RUNTIME_THRESHOLD=10
[:~] $ sleep 11
[:~] 40s $ 
[:~] 51s $ 
[:~] 51s $ 
[:~] 52s $ 
[:~] 52s $ 
[:~] 52s $ true
[:~] 55s $ false
[:~] 56s $ 
```

vabada:
Hi!
So my shell is: `GNU bash, version 4.4.12(1)-release (x86_64-pc-linux-gnu)`
And some code output:
~~~
[vabada:~] $ cd .liquidprompt/
[vabada:~/.liquidprompt] master 5s ± git pull
Already up-to-date.
[vabada:~/.liquidprompt] master 7s ± 
[...]

Issue #481, opened 2016-11-02
Default prompts for execution time and plugged in status not correct on macOS Sierra


mraible:
Shell: bash/zsh
Operating system: macOS Sierra
Liquid Prompt version (tag, commit): 9c1c8a378846c23e0a39be2aadd11531c2ecf196

I have two MacBook Pros, one with OS X El Capitan (version 10.11.6) and one with macOS Sierra (10.12.1).

I have the same Liquid Prompt version on both machines and both are configured the same in my `~/.bash_profile`. I haven't changed anything in the liquidprompt source tree. Everything works great on El Capitan. However, on macOS Sierra, the prompts are wrong.

macOS Sierra seems to add the time between commands rather than just showing how long it took to execute the last one. Also, it doesn't show that I'm plugged in. For example:

```bash
⌂69% [mraible:~/dev/liquidprompt] master 16s ± cd ~/dev/stormpath-sdk-java/
⌂66% [mraible:~/dev/stormpath-sdk-java] master 1m25s ± cd api/
⌂63% [mraible:~/dev/stormpath-sdk-java/api] master 1m27s ± mvn test
```

dolmen:
Could you tell me the bash versions on each machines?

asnr:
I'm having a similar problem with liquidprompt showing the time since the shell started rather than the time spent executing the last command, but I'm running El Capitan. I installed liquidprompt with homebrew:

```
$ brew info liquidprompt
liquidprompt: stable 1.11, HEAD
Adaptive prompt for bash and zsh shells
https://github.com/nojhan/liquidprompt
/usr/local/Cellar/liquidprompt/1.11 (6 files, 124.3K) *
  Built from source on 2016-08-08 at 20:05:37
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/liquidprompt.rb
```

I'm running bash version

```sh
$ echo $BASH_VERSION
3.2.57(1)-release
```

on iTerm2 build 3.0.12 on El Capitan version 10.11.6 (15G1108).

jasmas:
This is also not working for me on Sierra 10.12.5 (16F73) using either the built-in 3.2.57(1)-release version of bash or 4.4.12(1)-release version installed through homebrew.

jasmas:
I just tested #476 and it appears to fix the plugged-in status for me.

augmentedfourth:
[...]

Issue #436, opened 2016-04-04
[bash] Return value or runtime isn't shown if used with other tools that ose $PROMPT_COMMAND 


schorsch3000:
I use [direnv](http://direnv.net/), it also hooks as $PROMPT_COMMAND.

accoourding to the order i source direnf and liquidpromt i'll get a $PROMPT_COMMAND as:
`_direnv_hook;_lp_set_prompt`
this will have a working rumtime in my prompt, but the prompt doesn't show my the returncode of the command i've used. this is due to _direnv_hook overwrites $? before lp can get it.

the other way arround:
`_lp_set_prompt;_direnv_hook`
gives me the right return value, but a false runtime.
i'll get the time since the last prompt, nut the runtime.

my hacky solution to this problem is:

```
function _CATCH_RET_VAL(){
    export _CATCHED_RET_VAL=$?
}
function _THROW_RET_VAL(){
    return $_CATCHED_RET_VAL
}
export PROMPT_COMMAND="_CATCH_RET_VAL;_direnv_hook;_THROW_RET_VAL;_lp_set_prompt"

```
So, this is not something i needed to be solved, it'S more a notice that this problem exists :)

dolmen:
Good hack!
But note that you don't have to use `export`. Here is your code, fixed and simplified:
```
function _throw_ret_val(){
    return $_ret_val
}
PROMPT_COMMAND='_ret_val=$?;_direnv_hook;_throw_ret_val;_lp_set_prompt'
```

dolmen:
Note that Liquidprompt has native support for wrapping an existing $PROMPT_COMMAND. It is supposed to work, but I admit I have never used it myself.

So have you tried to setup direnv first so that $PROMPT_COMMAND is set when liquidprompt initialize?

schorsch3000:
your shorter and less verboste fix works fine, thx.
as described, i've tried both, with different results.

Issue #352, opened 2015-03-18
Execute "history -a; history -n;" at every command (previously in my PROMPT_COMMAND)


durdn:
To make sure I can re-use commands across multiple shells I had for a very long time this line in my `.bashrc`:

    export PROMPT_COMMAND="history -a; history -n;"

Liquidprompt is explicit in the documentation that the value will be overwritten but I still would like to execute those to commands at every prompt. Any suggestions on how to accomplish that?

waldyrious:
@durdn can you comment on [this answer](http://superuser.com/a/734410/112593) that says that -n should be avoided?

durdn:
@waldyrious I have used the version with `-n` for a while without much trouble but if there's an even more accurate solution I'll take it! I can go with `history -a; history -c; history -r`. The question stands, how easy is it to add this (i.e. my own custom post or pre-command) to liquid prompt?

dolmen:
@waldyrious

Just set your `PROMPT_COMMAND` *before* loading `liquidprompt`. Everything is supposed to work (despites being documented as not working, but just for safety). Check that your own `PROMPT_COMMAND` is properly saved into `LP_OLD_PROMPT_COMMAND`.

By the way, do not use `export` because `.bashrc` is already loaded in all sub shells, so you do not need the inheritance of this variable.

durdn:
Sorry I am late to comment on the issue.

I solved it by re-defining `PROMPT_COMMAND` like the following after sourcing `.liquidprompt`:

```
source $HOME/.liquidprompt
#make sure the history is updated at every command
shopt -s histappend
PROMPT_COMMAND="${PROMPT_COMMAND:+$PROMPT_COMMAND$'\n'}history -a; history -c; history -r"
```

dolmen:
@durdn Be aware that the `prompt_off`, `prompt_on`, and `prompt_OFF` commands of liquidprompt will not work with your config.

kynan:
@durdn Setting `PROMPT_COMMAND` *before* sourcing liquidprompt works fine for me: history commands are executed as expected.

PR #546, opened 2018-04-16
Explain unofficial work around for PROMPT_COMMAND

Rycieos:
Since there have been so many issues with people setting `PROMPT_COMMAND`
when they shouldn't, add the official explanation of the unofficial
work around to be able to set it and still get the prompt to work.

List of issues where this was at least part of the solution:
#276
#352
#436
#450
#463
#481
#516


nojhan:
Merged on develop, thanks!

Rycieos:
This is now in master as 2995687f372c7724791c8576e834155bbd0c6913 and in [v1.12-beta.1](https://github.com/nojhan/liquidprompt/releases/tag/v1.12-beta.1).

PR #556, opened 2018-08-28
_lp_connection: correctly detect local connection if X11 display doesn't start with colon

mpasternacki:
FreeBSD with slim display manager sets `DISPLAY=unix:0.0`, which is
currently detected as telnet session and displayed in ugly red.

```
[japhy:~] % who am i
japhy            pts/3        28 Aug 16:36 (unix:0.0)
[japhy:~] % echo $DISPLAY
unix:0.0
[japhy:~] % uname -a
FreeBSD sophon.local 12.0-CURRENT FreeBSD 12.0-CURRENT #0 r333486: Fri May 11 12:32:05 UTC 2018     maciej@hatter:/usr/obj/usr/src/amd64.amd64/sys/GENERIC-NODEBUG  amd64
```

It's quite likely that this renders the previous condition (`x"$whoami" = *'(:'*`) redundant, but I didn't want to remove it without confirmation.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -595,7 +595,7 @@ _lp_connection()
         # TODO check on *BSD
         local whoami="$(LC_ALL=C who am i)"
         local sess_parent="$(ps -o comm= -p $PPID 2> /dev/null)"
-        if [[ x"$whoami" != *'('* || x"$whoami" = *'(:'* || x"$whoami" = *'(tmux'* ]]; then
+        if [[ x"$whoami" != *'('* || x"$whoami" = *'(:'* || ( -n "${DISPLAY-}" && x"$whoami" = *"(${DISPLAY-})" ) || x"$whoami" = *'(tmux'* ]]; then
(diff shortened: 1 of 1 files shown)


mpasternacki:
Update: `x"$whoami" = *'(:'*` might be needed after all: screen on Linux (GNOME 3 session on Ubuntu 16.04) ends the line with `(:pts/2:S.0)`.

Rycieos:
I am curious to see if this fix is still needed. Can you test latest `master` to check?

If so, we need some data to test the amended data function, and to prevent a regression. Can you please use the new `tools/external-tool-tester.sh` script to upload sample data for the different machines you are talking about?

I am always willing to provide fixes, but if you want to fix this PR up, besides rebasing to get the latest changes, it would be good to add the test cases you gather to the `tests/test_connection.sh` test script.

Rycieos:
Thanks for the PR, but I have decided to solve this in a different, hopefully more reliable way.

PR #557, opened 2018-09-04
Improve fossil vcs support

milouse:
The following commit allows the use of dashes (-) and dots (.) in fossil branch names by replacing the `\w` regexp classes by the extended characters range `[a-zA-Z0-9_.-]`.

This avoid liquidprompt to display `no-branch` when the current tag contains a dot (very rare usage) or a dash (common use case).


Rycieos:
I'm closing this, as you mentioned it was already fixed by the work done for v2.0.0. And I have no idea what the original proposed change was; it seems that you kept pushing to this branch.

PR #561, opened 2018-12-29
Display selected node.js version if using NVM

kmARC:
This patch enables displaying the currently selected node.js version,
set by [NVM]. Implementation follows the python virtualenv display
setting.

nvm-liquidprompt

[NVM]: https://github.com/creationix/nvm

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -328,6 +328,7 @@ _lp_source_config()
     LP_ENABLE_TIME=${LP_ENABLE_TIME:-0}
     LP_ENABLE_RUNTIME=${LP_ENABLE_RUNTIME:-1}
     LP_ENABLE_VIRTUALENV=${LP_ENABLE_VIRTUALENV:-1}
+    LP_ENABLE_NVM=${LP_ENABLE_NVM:-1}
     LP_ENABLE_SCLS=${LP_ENABLE_SCLS:-1}
     LP_ENABLE_VCS_ROOT=${LP_ENABLE_VCS_ROOT:-0}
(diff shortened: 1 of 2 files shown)


kmARC:
We can set it up as disabled by default. I also wasn't sure whether this should be a feature at all - python virtualenv is a different concept, also nvm is not as popular as virtualenv. (I also agree it's slow. In fact is so slow that I rather aliased `nvm` [to be lazy loaded](https://github.com/kmARC/dotfiles/blob/master/bashrc.kmarc#L38) )

Yet I still think it's a huge help when developing with node, thus I made the PR.

liquidprompt
```diff
@@ -1740,6 +1742,14 @@ _lp_set_prompt()
         LP_VENV=
     fi
 
+    # Display the current NVM selection, if available
+    if [[ "$LP_ENABLE_NVM,${NVM_BIN-}" = 1,?* ]]; then
+        LP_NVM="${NVM_BIN##*node/v}"
+        LP_NVM=" [${LP_COLOR_NVM}node-${LP_NVM%/bin}${NO_COL}]"
```
Rycieos, line 1748: Looking at this again, why remove `node/v` to add back `node-`? This means a env name of `node/v1.2.3` is displayed as `node-1.2.3`, which could be confusing when people try to switch between versions. Why not simply display the actual env var?

Rycieos:

liquidprompt
Rycieos: The answer is that the env var contains the entire path to the `bin` directory, like:
```
/home/user/.nvm/versions/node/v18.1.0/bin
```

PR #562, opened 2019-01-03
Identify Guake process as local not telnet

### LINKED ISSUE #483: Shell inside Guake is detected as Telnet
Shell: bash
Operating system: Ubuntu 16.10
Liquid Prompt version (tag, commit): 9c1c8a3

It was already mentioned in https://github.com/nojhan/liquidprompt/issues/304, but somehow got lost. When a shell is started inside [Guake](http://guake-project.org/), it is wrongly detected as being a telnet session. The output from `who am i` looks like this:

    felix    pts/1        2016-11-09 10:24 (hostname)

Not sure if there is a generic way to solve this, but Guake sets `$GUAKE_TAB_UUID` which would be an easy way to detect this.

saildata:
Resolves #483

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -595,7 +595,8 @@ _lp_connection()
         # TODO check on *BSD
         local whoami="$(LC_ALL=C who am i)"
         local sess_parent="$(ps -o comm= -p $PPID 2> /dev/null)"
-        if [[ x"$whoami" != *'('* || x"$whoami" = *'(:'* || x"$whoami" = *'(tmux'* ]]; then
+        # Don't identify guake as telnet (GH #483)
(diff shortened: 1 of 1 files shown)


Rycieos:
Thanks for the PR!

We need some data to test the amended data function, and to prevent a regression. Can you please use the new `tools/external-tool-tester.sh` script to upload sample data for the different machines you are talking about?

I am always willing to provide fixes, but if you want to fix this PR up, besides rebasing to get the latest changes, it would be good to add the test cases you gather to the `tests/test_connection.sh` test script.

Rycieos:
Thanks for the PR, but I have decided to solve this in a different, hopefully more reliable way.

Issue #569, opened 2019-04-11
Display containerization status


alanhoyle:
Shell: bash/zsh
Operating system: 
Liquid Prompt version (tag, commit): 

Feature request: in HPC environments, we sometimes run interactive sessions in container software like Docker, LXC, or Singularity.  It would be great if there was some indications that a given command prompt was in a container or running natively.  

Detecting Singularity may be done by looking for the existence of these environment variables:
```
SINGULARITY_CONTAINER=ubuntu-latest.simg
SINGULARITY_NAME=ubuntu-latest.simg
```

Detecting Docker may be done by looking for the existence of the file `/.dockerenv` or looking at `/proc/1/cgroup` and noting that things look something like this:

```
$ cat /proc/1/cgroup 
11:memory:/docker/925[...]22f
10:hugetlb:/docker/925[...]22f
```

My understanding is that LXC may be done with something similar to the `/proc/1/cgroup` as above with Docker, but the values are like `11:memory:/lxc/[...]` but I can't test this.  

Note:  if I can figure out how to do this, I'll submit a pull request.

alanhoyle:
Pull request submitted: https://github.com/nojhan/liquidprompt/pull/570

Rycieos:
I like this feature, and I think it needs to happen, but I'm wondering how you are getting Liquidprompt into a container.

If you run `zsh` from bash, it will get all the exported vars from bash. But that doesn't happen when you run `docker exec -it ... bash`.

So what are you doing? Are you mounting Liquidprompt and your config file into the container and the sourcing it? Are you also mounting your `.bashrc` into the container for when you log in?

alanhoyle:
I think I followed-up with this answer at some point, but I'm not sure.  as I said above, I'm working in an HPC environment where we run scientific workflows and containerizing our software lets us more reproducibly run them.  In addition to batch-mode scripted runs, this use-case is for development, experimentation, ad-hoc analyses, etc.  

We mostly used it in the following cases:  

*  Inject the $USER into the container, and it mounts the user's home directory (we had a plugin that did this for Docker)
* Inject environment variables with `docker run -e VAR ...` or `... --env-file FILE ...`
* User-space containers using Singularity. This normally both mounts the calling user's $HOME, and mostly inherits the environment unless cleaned with `singularity run [-e/--cleanenv] ...`  

The Singularity case is by far our most common case in our current architecture, and having the prompt obviously show "You're running in a Singluarity container!" is really helpful.  

[...]

PR #570, opened 2019-04-11
added ability to detect/display container status

alanhoyle:
This modification adds the ability to display whether the current shell is running in a container (i.e. Docker, Singularity or LXC).  It also fixes a teensy typographical error on the README.md.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1707,22 +1707,24 @@ _lp_container()
 
     _lp_container=""   
 
-    if test -e /.dockerenv ||  grep -q "docker" /proc/1/cgroup  ; then
-	_lp_container="Docker"
+    if [[ "$OSTYPE" == "linux-gnu" ]] ; then 
+        if test -e /.dockerenv ||  grep -q "docker" /proc/1/cgroup  ; then
+	    _lp_container="Docker"
+	    
+	elif  grep -q "lxc" /proc/1/cgroup  ; then
+	    _lp_container="LXC"
+	    
+	fi
 	
-    elif  grep -q "lxc" /proc/1/cgroup  ; then
-	_lp_container="LXC"
-
-    fi
-
-    if test -z ${SINGULARITY_CONTAINER+x} || test -z ${SINGULARITY_NAME+x}  ; then
-    	 :
-    else
-	_lp_container="Singularity"
-    fi
-    
-    if test -n "$_lp_container" ; then 
-	echo -nE "«$LP_COLOR_CONTAINER$_lp_container$NO_COL»"
+	if test -z ${SINGULARITY_CONTAINER+x} || test -z ${SINGULARITY_NAME+x}  ; then
+    	    :
+	else
+	    _lp_container="Singularity"
+	fi
+	
+	if test -n "$_lp_container" ; then 
+	    echo -nE "«$LP_COLOR_CONTAINER$_lp_container$NO_COL»"
+	fi
     fi
 
 }
--- a/liquidprompt
+++ b/liquidprompt
@@ -23,6 +23,7 @@
 # AUTHORS #
 ###########
 
+# Alan Hoyle        <alan@alanhoyle.com>          # Containerization status
 # Alex Prengère     <alexprengere@gmail.com>      # Untracked git files
 # Anthony Gelibert  <anthony.gelibert@me.com>     # Several fix
 # Aurelien Requiem  <aurelien@requiem.fr>         # Major clean refactoring, variable path length, error codes, several bugfixes.
--- a/liquidprompt
+++ b/liquidprompt
@@ -336,6 +336,7 @@ _lp_source_config()
     LP_ENABLE_FQDN=${LP_ENABLE_FQDN:-0}
     # LP_DISABLED_VCS_PATH="${LP_DISABLED_VCS_PATH}"
     LP_ENABLE_SUDO=${LP_ENABLE_SUDO:-0}
+    LP_ENABLE_CONTAINER=${LP_ENABLE_CONTAINER:-1}
 
     LP_MARK_DEFAULT="${LP_MARK_DEFAULT:-$_LP_MARK_SYMBOL}"
     LP_MARK_BATTERY="${LP_MARK_BATTERY:-"⌁"}"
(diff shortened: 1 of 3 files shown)


alanhoyle:
This pull request fixes request  #569 

Here's an example showing what happens in Singularity and Docker with the pull request implemented.  
```
$ . $HOME/liquidprompt/liquidprompt
[@workstation:~] $ set |grep LP_ENABLE_CONTAINER=
LP_ENABLE_CONTAINER=1
[@workstation:~] $ singularity run  ~/singularity/ubuntu-latest.simg 
«Singularity» [@workstation:~] $ LP_ENABLE_CONTAINER=0
[@workstation:~] $ exit
exit
[@workstation:~] 15s $ docker run --rm -e HOME -v $HOME:$HOME:shared -ti ubuntu
«Docker» [:/] $ LP_ENABLE_CONTAINER=0
[:/] $ exit
exit
[@c6145-docker-2-0:~] 15s $ singularity run  ~/singularity/ubuntu-latest.simg 
«Singularity» [@workstation:~] $ 
```

alanhoyle:
with the "edb8ac9" commit, it now checks the OSTYPE variable and only does the container checks when it detects that the value is "linux-gnu"  Without that check, it generates spurious error messages on macOS.  However, since Docker is implemented as a lightweight linux VM on macOS, it can potentially still work even in that context.

alanhoyle:
I tested this with Docker and Singularity runs with the docker image zshusers/zsh-master:latest and it appears to work in that context too.

liquidprompt
```diff
@@ -1695,6 +1698,38 @@ else
...
+##################################
+
+_lp_container()
+{
+    (( LP_ENABLE_CONTAINER )) ||  return
+
+    _lp_container=""   
+
+    if [[ "$OSTYPE" == "linux-gnu" ]] ; then 
+        if test -e /.dockerenv ||  grep -q "docker" /proc/1/cgroup  ; then
```
Rycieos, line 1712: I have tested these checks myself and they work, but I would like to see links to documentation on this in the commit messages somewhere.

liquidprompt
```diff
@@ -1695,6 +1698,38 @@ else
...
+	    _lp_container="Docker"
+	    
+	elif  grep -q "lxc" /proc/1/cgroup  ; then
+	    _lp_container="LXC"
+	    
+	fi
+	
+	if test -z ${SINGULARITY_CONTAINER+x} || test -z ${SINGULARITY_NAME+x}  ; then
```
Rycieos, line 1720: `test` is less prefered than `[[` forms. All our supported shells support `[[`, use that instead. `[[` does not require quoting variables, while `[` and `test` do.

`-n ${var-}` has less issues than `-z ${var+x}` generally, so use that instead. Unless you are trying to test if the variable is set? I don't see how that would be a case were this should be triggered though.

liquidprompt
```diff
@@ -1695,6 +1698,38 @@ else
...
+# Detect if we're in a container #
+##################################
+
+_lp_container()
+{
+    (( LP_ENABLE_CONTAINER )) ||  return
+
+    _lp_container=""   
+
+    if [[ "$OSTYPE" == "linux-gnu" ]] ; then
```
Rycieos, line 1711: `OSTYPE` is never set. It wasn't set in my test container either. I'm not sure what you are trying to compare against.

Rycieos (changes requested):

liquidprompt
alanhoyle: I had seen $OSTYPE in:
* native 
     * macOS, 
     * Ubuntu, 
     * Fedora
* singularity
     * Ubuntu, 
     * Fedora
* docker
     * Ubuntu, 
     * Fedora

However, I found that $OSTYPE does not seem to be  set in Alpine-based containers.  What is your native OS?

alanhoyle:
I recently discovered podman, so I need to add support for that too.  Apparently the way to check for that is to look for 
`/run/.containerenv` documented at https://github.com/containers/libpod/blob/master/docs/podman-run.1.md

alanhoyle:
Would it be better to pull a completely fresh version and resubmit?

Rycieos:
> 
> 
> Would it be better to pull a completely fresh version and resubmit?

It doesn't make much of a difference to me. If you are running into issues because you opened this PR from your `master` branch, and don't want to mess that up with a rebase, then sure close this PR and open a new one.

liquidprompt
Rycieos: > I found that $OSTYPE does not seem to be set in Alpine-based containers.

Yup, that's what I tested on:
```
$ docker run -it --rm alpine:3.9 ash
/ # printf '%s\n' "$OSTYPE"

```

Why not test against `$LP_OS`?

alanhoyle: It's been a while since I messed with this so I think I was just using that to avoid some spurious errors on native macOS.  I don't know that I actually need the OS.

alanhoyle:
I'm totally changing my implementation to align better with 2.x.  In any case, I'd need to update the documentation as the README.md is no longer where any of that stuff would go.

alanhoyle:
@Rycieos:  liquidprompt has greatly changed in the updated version.  In the version that I had forked before, `_lp_set_prompt()` directly set variables through function calls, but that doesn't seem to be the case anymore.  

E.g. `LP_BATT="$(_lp_sr "$(_lp_battery_color)")"` would put the battery status. 

Now, it seems that this behavior is set somewhere else.  Can you point me in the right direction for how to do this?

Rycieos:
> 
> 
> Can you point me in the right direction for how to do this?

Commit f35d9acfddacc1f7f74174b45cf4c4aa9c84beca is a good example of the things you need to change.

[`__lp_set_prompt()`](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/functions/internal.html#lp_set_prompt) does still set [template variables](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/theme/default.html#template-sections) through function calls, it just does it slightly differently. You probably want to do this:
```bash
    if _lp_container_color; then
        LP_CONTAINER="$lp_container_color "
    else
        LP_CONTAINER=
    fi
```

You'll need to split your function into a [data function](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/functions/data.html) and a [theme function](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/functions/theme.html). The theme function is where the color and formatting gets added. They are now split so that other themes can access the data directly.

alanhoyle:
Thank you!  I now see that you had mentioned the new architecture in a previous comment, so my apologies for not reading closely enough.  

Once I get containerization worked out, maybe I'll add a pull request that detects interactive cluster jobs (SLURM, SGE, etc...)  ;-)

Rycieos:
> maybe I'll add a pull request that detects interactive cluster jobs (SLURM, SGE, etc...)

I'm open to it. With this new architecture, adding features doesn't slow things down, and even if a data function does occur a runtime cost, we can just ship it disabled by default.

PR #571, opened 2019-05-14
Add LP_RUNTIME_BELL to ring bell for slow commands

nolanl:
Disabled by default, if enabled the prompt will ring the terminal bell
after any command takes longer than LP_RUNTIME_THRESHOLD to execute.

This is handy when you have a long running process on a different
virtual desktop and want to be notified when it completes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ You can configure some variables in the `~/.config/liquidpromptrc` file:
 * `LP_LOAD_THRESHOLD`, the minimal value after which the load average is displayed
 * `LP_TEMP_THRESHOLD`, the minimal value after which the average temperature is displayed
 * `LP_RUNTIME_THRESHOLD`, the minimal value after which the runtime is displayed
+*  LP_RUNTIME_BELL , ring the bell when LP_RUNTIME_THRESHOLD is exceeded.
 * `LP_PATH_LENGTH`, the maximum percentage of the screen width used to display the path
 * `LP_PATH_KEEP`, how many directories to keep at the beginning of a shortened path
 * `LP_HOSTNAME_ALWAYS`, a choice between always displaying the hostname or
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ You can configure some variables in the `~/.config/liquidpromptrc` file:
 * `LP_LOAD_THRESHOLD`, the minimal value after which the load average is displayed
 * `LP_TEMP_THRESHOLD`, the minimal value after which the average temperature is displayed
 * `LP_RUNTIME_THRESHOLD`, the minimal value after which the runtime is displayed
-*  LP_RUNTIME_BELL , ring the bell when LP_RUNTIME_THRESHOLD is exceeded.
+* `LP_RUNTIME_BELL_THRESHOLD`, the minimal value after which the bell is rung.
 * `LP_PATH_LENGTH`, the maximum percentage of the screen width used to display the path
 * `LP_PATH_KEEP`, how many directories to keep at the beginning of a shortened path
 * `LP_HOSTNAME_ALWAYS`, a choice between always displaying the hostname or
@@ -174,6 +174,7 @@ prompt-building process:
 * `LP_ENABLE_SCREEN_TITLE`, if you want to use the prompt as your screen window's title
 * `LP_ENABLE_SSH_COLORS`, if you want different colors for hosts you SSH into
 * `LP_ENABLE_RUNTIME`, if you want to display the runtime of the last command
+* `LP_ENABLE_RUNTIME_BELL`, if you want to ring the bell when a runtime threshold is exceeded.
 * `LP_ENABLE_SUDO`, if you want the prompt mark to change color while you have password-less root access
 * `LP_ENABLE_FQDN`, if you want the display of the fully qualified domain name
 * `LP_ENABLE_TIME`, if you want to display the time at which the prompt was shown
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ You can configure some variables in the `~/.config/liquidpromptrc` file:
 * `LP_LOAD_THRESHOLD`, the minimal value after which the load average is displayed
 * `LP_TEMP_THRESHOLD`, the minimal value after which the average temperature is displayed
 * `LP_RUNTIME_THRESHOLD`, the minimal value after which the runtime is displayed
-* `LP_RUNTIME_BELL_THRESHOLD`, the minimal value after which the bell is rung.
+* `LP_RUNTIME_BELL_THRESHOLD`, the minimal value after which the bell is rung. See LP_ENABLE_RUNTIME_BELL.
 * `LP_PATH_LENGTH`, the maximum percentage of the screen width used to display the path
 * `LP_PATH_KEEP`, how many directories to keep at the beginning of a shortened path
 * `LP_HOSTNAME_ALWAYS`, a choice between always displaying the hostname or
--- a/liquidprompt
+++ b/liquidprompt
@@ -1422,7 +1422,9 @@ _lp_runtime()
         # if enabled, ring the bell
         if (( LP_RUNTIME_BELL ))
         then
-            tput bel
+            #bash reprints $PS1 on every SIGWINCH, so we tput to stderr here so the bell is not
+            #captured in PS1.  This way the bell only rings once, not on every terminal resize.
+            tput bel >&2
         fi
     fi
     :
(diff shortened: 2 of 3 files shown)


nolanl:
@Rycieos Thanks for the review.

Your assessment of && vs if was the exact reasoning I used to choose.  Happy to change it if folks prefer the other way.

I generally agree with being conservative on prompt performance.  The reason I submitted this instead of keeping it in my fork is that the performance impact when enabled only applies immediately after finishing running a long (seconds to minutes, depending on config) process, so a few extra milliseconds or 10s of milliseconds in only that case seemed irrelevant.

nolanl:
I've updated this slightly (4 character change) to fix a bug whereby the bell would be rung again every time the terminal is resized after a slow command.  This is because bash reprints PS1 after SIGWINCH.  The fix is to print the bell character to stderr, so that it goes out immediately instead of being captured in PS1.

nolanl:
@Rycieos Noticed you've started merging PRs lately, any chance you could merge this one?

Rycieos:
I have been, yes.
I still do like this feature, but giving it another look over, I'm wondering if it can be improved.

1. The default RUNTIME_THRESHOLD is 2. This means the bell would be ringing a lot, though I realize that this feature is disabled by default.
2. This is tied to the RUNTIME_THRESHOLD. It would be nice to have two different time settings for this. IE 10 seconds for the runtime to print, 30 seconds for a bell to trigger.
3. Would be super cool for the bell to only ring when the terminal is not in the foreground. This is probably outside the scope of Liquidprompt.

I'm not saying these need to get changed before I would merge, just throwing my thoughts out there for some feedback.

nolanl:
I'll do #2 this weekend.

I don't think this change is the right place to change the existing LP_RUNTIME_THRESHOLD default.

#3 would be cool, but I'm not sure it is possible in general. In my usage of this, it doesn't end up mattering, because what I'm after is the terminal notifying the window manager, which highlights the virtual desktop, if I'm on a different virtual desktop, and highlights the terminal window if I am on the same virtual desktop. No actual noise is generated.

nolanl:
@Rycieos I've added a LP_RUNTIME_BELL_THRESHOLD config param and updated the PR.

README.md
```diff
@@ -147,6 +147,7 @@ You can configure some variables in the `~/.config/liquidpromptrc` file:
 * `LP_LOAD_THRESHOLD`, the minimal value after which the load average is displayed
 * `LP_TEMP_THRESHOLD`, the minimal value after which the average temperature is displayed
 * `LP_RUNTIME_THRESHOLD`, the minimal value after which the runtime is displayed
+* `LP_RUNTIME_BELL_THRESHOLD`, the minimal value after which the bell is rung.
```
Rycieos, line 150: Maybe mention "see LP_ENABLE_RUNTIME_BELL" here?

liquidprompt
```diff
@@ -1419,10 +1421,16 @@ _lp_runtime()
         (( _LP_RUNTIME_SECONDS >= 60 )) && echo -n $((_LP_RUNTIME_SECONDS % 3600 / 60))m
         echo -n $((_LP_RUNTIME_SECONDS % 60))"s${NO_COL}"
     fi
+    if (( LP_ENABLE_RUNTIME_BELL && _LP_RUNTIME_SECONDS >= LP_RUNTIME_BELL_THRESHOLD ))
+    then
+        #bash reprints $PS1 on every SIGWINCH, so we tput to stderr here so the bell is not
+        #captured in PS1.  This way the bell only rings once, not on every terminal resize.
+        tput bel >&2
```
Rycieos, line 1428: I just learned this this week myself: note that `tput` isn't exactly universal in its capability codes. Stupid POSIX doesn't actually define what or how `tput` should define the codes that the user uses.

Some BSDs use slightly different codes. We actually already code for this here:
https://github.com/nojhan/liquidprompt/blob/da9e97f8025f01fbe4c0ff06eccdc1454b63256e/liquidprompt#L226-L237
Reading the [FreeBSD docs](https://www.freebsd.org/cgi/man.cgi?query=terminfo&sektion=5&apropos=0&manpath=FreeBSD+12.1-RELEASE+and+Ports), it looks like the code of `bl` should work:
```suggestion
        { tput bel || tput bl ; } 2>/dev/null >&2
```
Note the addition of `2>/dev/null`. If the `tput` command errors out for some reason (or on the first try), I don't want it to be printed to terminal.

I don't have any BSD environments to test on (hoping to fix that soon), but I'm comfortable with this change since it matches our other uses of `tput`.

liquidpromptrc-dist
```diff
@@ -126,6 +126,15 @@ LP_ENABLE_RUNTIME=0
 # Recommended value is 2
 LP_RUNTIME_THRESHOLD=2
 
+# Ring the terminal bell if the runtime of the previous command exceeded
+# LP_ENABLE_RUNTIME_THRESHOLD
```
Rycieos, line 130: Simple typo
```suggestion
# LP_RUNTIME_BELL_THRESHOLD
```

nolanl:
Done. Thanks for the review!

nolanl:
Oh wait, the tput change doesn't seem to work, let me track that down.

nolanl:
OK, fixed, amended and force pushed.

Rycieos:
Looks all good, but I just tested it, and it is sending a bell every time I resize my terminal. I really can't figure out why, tried printing out PS1, but I don't see the code in there anywhere.
```bash
$ echo -E "$PS1" | sed 's/\(.\?\)/ \1/g'
 [ \ u \ [   m \ ] \ [  3 3 m \ ] @ \ [   m \ ] \ [  3 4 m \ ] \ h \ [   m \ ] \ [  3 2 m \ ] : \ [   m \ ] \ [  1 m \ ] ~ / d e v / l i q u i d p r o m p t \ [   m \ ] ]   \ [  3 2 m \ ] n o l a n l - f e a t u r e / r u n t i m e - b e l l \ [  3 3 m \ ] + \ [  3 1 m \ ] * \ [  3 1 m \ ] \ [   m \ ]   \ [  3 3 m \ ] 1 1 s \ [   m \ ]    \ [  1 m \ ] ± \ [   m \ ]
```
But that also isn't working for just the bell:
```bash
$ tput bel | sed 's/\(.\?\)/ \1/g'
 # bell goes off
```

I think that the bell character `^G`, since it is a single character, there is no way to prevent it from being caught by the terminal. And if I store PS1 and echo it later, I get a bell again:
```bash
$ var="$PS1"
$ echo $var  # bell goes off
```
So it has to be getting in there somehow.

nolanl:
Yeah, it was getting captured. Fixed now, please retest.

nolanl:
BTW, I tried it in FreeBSD 12, and "tput bl" works.

liquidprompt
```diff
@@ -1419,10 +1421,16 @@ _lp_runtime()
         (( _LP_RUNTIME_SECONDS >= 60 )) && echo -n $((_LP_RUNTIME_SECONDS % 3600 / 60))m
         echo -n $((_LP_RUNTIME_SECONDS % 60))"s${NO_COL}"
     fi
+    if (( LP_ENABLE_RUNTIME_BELL && _LP_RUNTIME_SECONDS >= LP_RUNTIME_BELL_THRESHOLD ))
+    then
+        #bash reprints $PS1 on every SIGWINCH, so we tput to stderr here so the bell is not
+        #captured in PS1.  This way the bell only rings once, not on every terminal resize.
+        { tput bel || tput bl 2>/dev/null ; } >&2
```
Rycieos, line 1428: I was an idiot and put these redirections in the wrong order. That's why it stopped working right. See here for an explanation: http://mywiki.wooledge.org/BashFAQ/055
```suggestion
        { tput bel || tput bl ; } >&2 2>/dev/null
```

nolanl: Is there any reason to favor the redirect to devnull outside the {} vs leaving it inside?

nolanl: Tested and force pushed.

nolanl:
BTW, I don't so much have FreeBSD as I have installed vagrant. I just did:
```
$ mkdir freebsd; cd freebsd
$ vagrant init generic/freebsd12
$ vagrant up
$ vagrant ssh
$ vagrant destroy
$ cd ..; rm -rf freebsd
$ vagrant box remove generic/freebsd12
```
Very convenient for this sort of testing.

nolanl:
Thanks for taking the time to review/test!

I usually squash, but I don't think it much matters either way.

Rycieos:
Merged in af63c4c

PR #572, opened 2019-05-26
Update default template file

SuperSandro2000:


SuperSandro2000:
Yes exactly. I thought a few minutes about a commit message and didn't really came up with a better one. If you have a suggestion I rebase it.

We could add a special comment before and after the section and write a script that updates the template.

SuperSandro2000:
bump

Issue #576, opened 2019-06-26
Prompt stay in red after sudo removal


milouse:
Shell: zsh
Operating system: Archlinux
Liquid Prompt version (tag, commit): develop^HEAD

LP_MARK variable is only computed under the `if [[ "${LP_OLD_PWD-}" != "LP:$PWD" ]]; then` check. Thus if you enter a sudo command and after that you type `sudo -K` to erase your priviledges, without changing of directory, the prompt will stay red.

Current behavior:

![bug](https://user-images.githubusercontent.com/349239/60160785-6a23f080-97f6-11e9-9e8c-08f1202ac008.png)

Expected behavior:

![solution](https://user-images.githubusercontent.com/349239/60160803-71e39500-97f6-11e9-8777-3af7bf2ef05a.png)

What is weird is that it seems to be only the case when you are currently visiting a VCS directory. Still, after moving the LP_MARK computation outside the PWD check it works at any time.

Rycieos:
Unfortunately, this bug is only on the `develop` (now named `nojhan/develop`) branch. It was introduced here:

https://github.com/nojhan/liquidprompt/commit/b08576844599a6ed5323bf6ba71158355b0f1a69#diff-ef34fce883a908e05fe28695e45b06605246681c63a017518178cb02f456af09R1830

This branch is no longer the main branch, and for reasons that are exampled above, should not be used. Most of the changes on that branch have been merged already and are in v1.12.0.

If there are features or fixes that are on `nojhan/develop` that I missed and are not in `rework` (soon to be merged to `master`), please open feature requests or bug reports.

PR #577, opened 2019-06-26
Always recompute LP_MARK

milouse:
This allow the immediate reflection of sudo state change

fix: #576


dolmen:
Unfortunately this slows the prompt for a feature which is not used that often (few users care about the duration of their sudo credentials).

Rycieos:
Closing as #576 is invalid.

Issue #559, opened 2018-12-04
Kubernetes support


imsky:
It'd be great to see the current Kubernetes config context. It's most helpful as a persistent display, so it should probably be off by default. Also, since context names can become quite long, the display should probably be truncated.

Shell: bash
Operating system: macOS 10.12.6
Liquid Prompt version (tag, commit): eda83efe4e0044f880370ed5e92aa7e3fdbef971

Rycieos:
I merged @mschuett's commit 0660342 in #578, which implements exactly what is described here. Will be part of the v2.1 release.

PR #578, opened 2019-07-07
add Kubernetes context

mschuett:
This commit adds the current Kubernetes context to the prompt
(matches feature request #559).

I am not that good with shell programming and I have the feeling the unconditional call to `_lp_kube_context` could be handled more elegantly. So any suggestions for improvements are welcome.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -177,6 +177,28 @@ General
 Features
 --------
 
+.. attribute:: LP_DELIMITER_KUBECONTEXT
+   :type: string
+   :value: ""
+
+   Delimiter to shorten the Kubernetes context.
+
+   Usage example:
+
+   * if your context names are dev-cluster and test-cluster,
+     then set this to "-" in order to output "dev" and "test" in prompt.
+   * if your context names are dev.k8s.example.com and test.k8s.example.com,
+     then set this to "." in order to output "dev" and "test" in prompt.
+   * if using OpenShift then set this to "/" to show only the project name
+     without the cluster and user parts.
+
+   If set to the empty string no truncating will occur (this is the default).
+
+   See also: :attr:`LP_ENABLE_KUBECONTEXT`, :attr:`LP_COLOR_KUBECONTEXT`,
+   and :attr:`LP_MARK_KUBECONTEXT`.
+
+   .. versionadded:: 2.1
+
 .. attribute:: LP_DISABLED_VCS_PATH
    :type: string
    :value: ""
@@ -330,6 +352,19 @@ Features
    Display the number of running and sleeping shell jobs.
 
    See also: :attr:`LP_COLOR_JOB_R` and :attr:`LP_COLOR_JOB_Z`.
+
+.. attribute:: LP_ENABLE_KUBECONTEXT
+   :type: bool
+   :value: 0
+
+   Display the current `Kubernetes <https://kubernetes.io/>`_ `context`_.
+
+   .. _`context`: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/
+
+   See also: :attr:`LP_DELIMITER_KUBECONTEXT`, :attr:`LP_COLOR_KUBECONTEXT`,
+   and :attr:`LP_MARK_KUBECONTEXT`.
+
+   .. versionadded:: 2.1
 
 .. attribute:: LP_ENABLE_LOAD
    :type: bool
@@ -702,6 +737,26 @@ Marks
 
    See also: :attr:`LP_ENABLE_HG`.
 
+.. attribute:: LP_MARK_KUBECONTEXT
+   :type: string
+   :value: "⎈"
+
+   Mark used to prefix the current Kubernetes context.
+
+   Used to visually distinguish the Kubernetes context from other
+   context fields like the Python virtual environment (see
+   :attr:`LP_ENABLE_VIRTUALENV`) and the Red Hat Software Collection
+   (see :attr:`LP_ENABLE_SCLS`).
+
+   The display of Unicode characters varies among Terminal and Font settings,
+   so you might try alternative marks. Single symbol alternatives to the
+   default "⎈" (U+2388, Helm Symbol) are "☸" (U+2638, Wheel of Dharma)
+   or "κ" (U+03BA, Greek Small Letter Kappa).
+
+   See also: :attr:`LP_ENABLE_KUBECONTEXT`.
+
+   .. versionadded:: 2.1
+
 .. attribute:: LP_MARK_LOAD
    :type: string
    :value: "⌂"
@@ -957,6 +1012,16 @@ Valid preset color variables are:
    Color used for sleeping shell jobs.
 
    See also: :attr:`LP_ENABLE_JOBS`.
+
+.. attribute:: LP_COLOR_KUBECONTEXT
+   :type: string
+   :value: $CYAN
+
+   Color used for the current Kubernetes context.
+
+   See also: :attr:`LP_ENABLE_KUBECONTEXT`.
+
+   .. versionadded:: 2.1
 
 .. attribute:: LP_COLOR_MARK
    :type: string
--- a/docs/theme/default.rst
+++ b/docs/theme/default.rst
@@ -180,6 +180,13 @@ default order if the user does not configure a different template.
    The current Python (or Conda) virtual environment. See
    :attr:`LP_ENABLE_VIRTUALENV`.
 
+.. attribute:: LP_KUBECONTEXT
+
+   The current Kubernetes context. See
+   :attr:`LP_ENABLE_KUBECONTEXT`.
+
+   .. versionadded:: 2.1
+
 .. attribute:: LP_PROXY
 
    A ↥ (:attr:`LP_MARK_PROXY`) if an HTTP proxy is in use. See
--- a/liquid.theme
+++ b/liquid.theme
@@ -24,6 +24,7 @@ if [[ "$(locale -k LC_CTYPE | sed -n 's/^charmap="\(.*\)"/\1/p')" == *"UTF-8"* ]
     LP_MARK_STASH="+"          # if git has stashs
     LP_MARK_SHORTEN_PATH=" … " # prompt mark in shortened paths
     LP_MARK_PERM=":"           # separator between host and path
+    LP_MARK_KUBECONTEXT="⎈"    # Kubernetes context
 else
     # If charset is anything else, fallback to ASCII chars
     LP_MARK_BATTERY="b"
@@ -40,6 +41,7 @@ else
     LP_MARK_STASH="+"
     LP_MARK_SHORTEN_PATH=" ... "
     LP_MARK_PERM=":"
+    LP_MARK_KUBECONTEXT="k8s:"
 fi
 
 LP_MARK_BRACKET_OPEN="["  # open bracket
@@ -115,6 +117,9 @@ LP_COLOR_IN_MULTIPLEXER="$BOLD_BLUE"
 
 # Virtual environment
 LP_COLOR_VIRTUALENV="$CYAN"
+
+# Kubernetes
+LP_COLOR_KUBECONTEXT="$CYAN"
 
 # Runtime
 LP_COLOR_RUNTIME="$YELLOW"
--- a/liquidprompt
+++ b/liquidprompt
@@ -194,6 +194,8 @@ __lp_source_config() {
     LP_PS1=${LP_PS1:-""}
     LP_PS1_PREFIX=${LP_PS1_PREFIX:-""}
     LP_PS1_POSTFIX=${LP_PS1_POSTFIX:-""}
+    LP_MARK_KUBECONTEXT=${LP_MARK_KUBECONTEXT:-"⎈"}
+    LP_DELIMITER_KUBECONTEXT=${LP_DELIMITER_KUBECONTEXT:-""}
 
     LP_ENABLE_PERM=${LP_ENABLE_PERM:-1}
     LP_ENABLE_SHORTEN_PATH=${LP_ENABLE_SHORTEN_PATH:-1}
@@ -224,6 +226,7 @@ __lp_source_config() {
     LP_ENABLE_COLOR=${LP_ENABLE_COLOR:-1}
     LP_ENABLE_ERROR=${LP_ENABLE_ERROR:-1}
     LP_ENABLE_DIRSTACK=${LP_ENABLE_DIRSTACK:-0}
+    LP_ENABLE_KUBECONTEXT=${LP_ENABLE_KUBECONTEXT:-0}
 
     LP_MARK_DEFAULT="${LP_MARK_DEFAULT:-$_LP_MARK_SYMBOL}"
     LP_MARK_BATTERY="${LP_MARK_BATTERY:-"⌁"}"
@@ -288,6 +291,7 @@ __lp_source_config() {
     LP_COLOR_RUNTIME=${LP_COLOR_RUNTIME:-$YELLOW}
     LP_COLOR_VIRTUALENV=${LP_COLOR_VIRTUALENV:-$CYAN}
     LP_COLOR_DIRSTACK=${LP_COLOR_DIRSTACK:-$BOLD_YELLOW}
+    LP_COLOR_KUBECONTEXT=${LP_COLOR_KUBECONTEXT:-$CYAN}
 
     if [[ -z "${LP_COLORMAP-}" ]]; then
         LP_COLORMAP=(
@@ -486,6 +490,8 @@ lp_activate() {
     else
         _lp_require_tool BATT acpi
     fi
+
+    _lp_require_tool KUBECONTEXT kubectl
 
     unset -f _lp_require_tool
 
@@ -1141,6 +1147,28 @@ _lp_software_collections_color() {
     _lp_software_collections || return "$?"
 
     lp_software_collections_color="[${LP_COLOR_VIRTUALENV}${lp_software_collections}${NO_COL}]"
+}
+
+
+_lp_kubernetes_context() {
+    (( LP_ENABLE_KUBECONTEXT )) || return 2
+
+    local kubernetes_context
+    kubernetes_context=$(kubectl config current-context 2>/dev/null) || return 1
+
+    if [[ -n "$LP_DELIMITER_KUBECONTEXT" ]]; then
+        kubernetes_context="${kubernetes_context%%${LP_DELIMITER_KUBECONTEXT}*}"
+    fi
+
+    local ret
+    __lp_escape "$kubernetes_context"
+    lp_kubernetes_context=$ret
+}
+
+_lp_kubernetes_context_color() {
+    _lp_kubernetes_context || return "$?"
+
+    lp_kubernetes_context_color="[${LP_MARK_KUBECONTEXT}${LP_COLOR_KUBECONTEXT}${lp_kubernetes_context}${NO_COL}]"
 }
 
 # Same as bash '\l', but can be inlined as a constant as the value will not
@@ -2695,6 +2723,12 @@ _lp_default_theme_prompt_data() {
         LP_VENV=
     fi
 
+    if _lp_kubernetes_context_color; then
+        LP_KUBECONTEXT=" $lp_kubernetes_context_color"
+    else
+        LP_KUBECONTEXT=
+    fi
+
     if _lp_software_collections_color; then
         LP_SCLS=" $lp_software_collections_color"
     else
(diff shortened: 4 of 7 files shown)


lyoshenka:
@dolmen what's the next step to getting this PR merged? :wink:

Rycieos:
[Release Candidate v2.0.0-rc.1](https://github.com/nojhan/liquidprompt/releases/tag/v2.0.0-rc.1) is now out, which means that the rework is complete. So the next step is to fix the merge conflicts.

This looks really good, just needs a call to `_lp_require_tool KUBECONTEXT kubectl` like other tools have:
https://github.com/nojhan/liquidprompt/blob/2daf0e453039950d79dcf0eec5a04931e363a92a/liquidprompt#L473

liquidprompt
```diff
@@ -2730,7 +2764,7 @@ _lp_default_theme_prompt_template() {
         # add user, host and permissions colon
         PS1+="${LP_BRACKET_OPEN}${LP_USER}${LP_HOST}${LP_PERM}"
 
-        PS1+="${LP_PWD}${LP_DIRSTACK}${LP_BRACKET_CLOSE}${LP_SCLS}${LP_VENV}${LP_PROXY}"
+        PS1+="${LP_PWD}${LP_BRACKET_CLOSE}${LP_SCLS}${LP_VENV}${LP_PROXY}${LP_KUBE}"
```
Rycieos: This section is obviously messed up from a bad merge (deletes `LP_DIRSTACK`).

liquidprompt
```diff
@@ -2343,6 +2348,28 @@ _lp_runtime_color() {
...
 }
 
+######################
+# Kubernetes context #
+######################
+
+_lp_kube_context() {
+   (( LP_ENABLE_KUBECONTEXT )) || return
+
+   local _LP_KUBE_CONTEXT=$(kubectl config current-context 2>/dev/null)
```
Rycieos, line 2358: Make this local lowercase.

Is the last line's check needed? Could it be moved earlier to save time? The docs don't say if it errors if there is no context: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#config
But in my testing it errors:
> error: current-context is not set

```suggestion
   local kubernetes_context
   kubernetes_context=$(kubectl config current-context 2>/dev/null) || return 1
```

liquidprompt
```diff
@@ -2343,6 +2348,28 @@ _lp_runtime_color() {
     lp_runtime_color="${LP_COLOR_RUNTIME}${lp_runtime_format}${NO_COL}"
 }
 
+######################
+# Kubernetes context #
+######################
+
+_lp_kube_context() {
```
Rycieos, line 2355: This should be named after the full name:
```suggestion
_lp_kubernetes_context() {
```

liquidprompt
```diff
@@ -2343,6 +2348,28 @@ _lp_runtime_color() {
...
+   (( LP_ENABLE_KUBECONTEXT )) || return
+
+   local _LP_KUBE_CONTEXT=$(kubectl config current-context 2>/dev/null)
+   if [[ "$LP_DELIMITER_KUBECONTEXT" ]]; then
```
Rycieos, line 2359: Use `-n` to be explicit:
```suggestion
   if [[ -n $LP_DELIMITER_KUBECONTEXT ]]; then
```

liquidprompt
```diff
@@ -2343,6 +2348,28 @@ _lp_runtime_color() {
...
+     lp_kube_context="${_LP_KUBE_CONTEXT%%${LP_DELIMITER_KUBECONTEXT}*}"
+   else
+     lp_kube_context="$_LP_KUBE_CONTEXT"
+   fi
```
Rycieos, line 2363: You can drop the else, because this needs to be passed through `__lp_escape()`:
```suggestion
   fi
 
    local ret
    __lp_escape "$kubernetes_context"
    lp_kube_context=$ret
```

liquidprompt
```diff
@@ -2343,6 +2348,28 @@ _lp_runtime_color() {
...
+   [[ -n "$lp_kube_context" ]]
+}
```
Rycieos, line 2365: This function is indented 3 spaces, should be 4.

liquidprompt
```diff
@@ -2343,6 +2348,28 @@ _lp_runtime_color() {
     lp_runtime_color="${LP_COLOR_RUNTIME}${lp_runtime_format}${NO_COL}"
 }
 
+######################
+# Kubernetes context #
+######################
```
Rycieos, line 2353: Don't add a new section for this, add it to the Development Environment section.

liquidprompt
```diff
@@ -2343,6 +2348,28 @@ _lp_runtime_color() {
...
+   else
+     lp_kube_context="$_LP_KUBE_CONTEXT"
+   fi
+   [[ -n "$lp_kube_context" ]]
+}
+
+_lp_kube_context_color() {
+    _lp_kube_context || return "$?"
+
+    lp_kube_context_color="${LP_COLOR_KUBECONTEXT}${LP_BRACKET_OPEN}${LP_MARK_KUBECONTEXT}${lp_kube_context}${LP_BRACKET_CLOSE}${NO_COL}"
```
Rycieos, line 2370: Our other Development Environment theme functions are using hard coded "[]" (see python_env_color():
https://github.com/nojhan/liquidprompt/blob/2daf0e453039950d79dcf0eec5a04931e363a92a/liquidprompt#L1126

While I would be open to a PR adding a config option for those brackets, they are different than [`LP_MARK_BRACKET_OPEN`](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/config.html#LP_MARK_BRACKET_OPEN), which the docs specify are for the central part of the prompt.

Not to mention that if in a multiplexer, the color added to these vars would break the color here:
https://github.com/nojhan/liquidprompt/blob/2daf0e453039950d79dcf0eec5a04931e363a92a/liquidprompt#L2607-L2609

Use hardcoded brackets here (for now):
```suggestion
    lp_kube_context_color="${LP_COLOR_KUBECONTEXT}[${LP_MARK_KUBECONTEXT}${lp_kube_context}]${NO_COL}"
```

docs/config.rst
```diff
@@ -319,6 +319,28 @@ Features
 
    See also: :attr:`LP_COLOR_JOB_R` and :attr:`LP_COLOR_JOB_Z`.
 
+.. attribute:: LP_ENABLE_KUBECONTEXT
+   :type: bool
+   :value: 1
+
+   Display the current Kubernetes context.
```
Rycieos, line 326: Add a link to both Kubernetes and k8s context:
```suggestion
   Display the current `Kubernetes <https://kubernetes.io/>`_ `context`_.
   
   .. _`context`: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/
```

docs/config.rst
```diff
@@ -319,6 +319,28 @@ Features
...
+
+   See also: :attr:`LP_DELIMITER_KUBECONTEXT`, :attr:`LP_COLOR_KUBECONTEXT`, and :attr:`LP_MARK_KUBECONTEXT`.
+
+.. attribute:: LP_DELIMITER_KUBECONTEXT
```
Rycieos, line 332: Move this section higher, should be in alphabetical order.

liquidprompt
```diff
@@ -475,6 +479,7 @@ lp_activate() {
     _lp_require_tool FOSSIL fossil
     _lp_require_tool HG hg
     _lp_require_tool BZR bzr
+    _lp_require_tool KUBECONTEXT kubectl
```
Rycieos, line 482: Move this down 13ish lines, this section is for VCSs

liquidprompt
```diff
@@ -2343,6 +2348,28 @@ _lp_runtime_color() {
     lp_runtime_color="${LP_COLOR_RUNTIME}${lp_runtime_format}${NO_COL}"
 }
 
+######################
+# Kubernetes context #
+######################
+
+_lp_kube_context() {
+   (( LP_ENABLE_KUBECONTEXT )) || return
```
Rycieos, line 2356: Should return `2` if disabled.

liquidprompt
```diff
@@ -224,6 +224,7 @@ __lp_source_config() {
     LP_ENABLE_COLOR=${LP_ENABLE_COLOR:-1}
     LP_ENABLE_ERROR=${LP_ENABLE_ERROR:-1}
     LP_ENABLE_DIRSTACK=${LP_ENABLE_DIRSTACK:-0}
+    LP_ENABLE_KUBECONTEXT=${LP_ENABLE_KUBECONTEXT:-1}
```
Rycieos, line 227: This should be disabled by default, as unlike other development environments (E.g. in Python where you need to `. venv/bin/activate`) a `kubectl` context is always on, and it is not possible (or at least not intended) to have no context at all. Meaning if you have `kubectl` installed, a context will _always_ show up in the prompt.

docs/config.rst
```diff
@@ -319,6 +319,32 @@ Features
...
 
+.. attribute:: LP_ENABLE_KUBECONTEXT
+   :type: bool
+   :value: 1
+
+   Display the current Kubernetes context.
+
+   See also: :attr:`LP_DELIMITER_KUBECONTEXT`, :attr:`LP_COLOR_KUBECONTEXT`, and :attr:`LP_MARK_KUBECONTEXT`.
+
+   .. versionadded:: 2.0
```
Rycieos, line 330: This will be merged after v2.0 is released, this should be `2.1`.

docs/theme/default.rst
```diff
@@ -180,8 +180,13 @@ default order if the user does not configure a different template.
    The current Python (or Conda) virtual environment. See
    :attr:`LP_ENABLE_VIRTUALENV`.
 
+.. attribute:: LP_KUBE
+
+   The current Kubernetes context. See
+   :attr:`LP_ENABLE_KUBECONTEXT`.
 .. attribute:: LP_PROXY
 
+
```
Rycieos, line 189: This extra line should be after line 186.

Rycieos (changes requested):
This patch changes `liquidprompt`'s file permissions, removing the execute bit. This can not happen for historical reasons.

This needs to be rebased and/or squashed to get rid of the messy merge commit and multiple fixup commits (to 1-3 commits). I can do this myself (once the patch looks good), but then Github won't realize that I merged the PR and it will show it "closed". If you care and want the PR to be marked as "merged", you will need to do the commit fixups yourself.

mschuett:
I think I got it. With the new structure for code and documentation this certainly needs a proof-reading...

One question: how should we color the brackets?

My current code has the brackets in cyan and they are visibly different from the Python Virtualenv block:

liquidprompt-kubecontext

Alternative with no color for brackets and mark:
liquidprompt-kubecontext2

Rycieos:
> One question: how should we color the brackets?
> 
> My current code has the brackets in cyan and they are visibly different from the Python Virtualenv block:

Good catch, I saw that, and forgot to mention it.

I would say follow the "standard" that the Python virtual env theme function does. It's probably just personal preference, and if someone really wants it the other way, they can make their own theme (or modify the default one) to do that.

Usually, the mark is colored the same as the text, but I don't think that is always the case. Probably personal preference, I would say color it to be consistent.

liquidprompt
mschuett: Wow your review was even faster than my question about it :-)

docs/config.rst
```diff
@@ -165,6 +165,22 @@ General
 Features
 --------
 
+.. attribute:: LP_DELIMITER_KUBECONTEXT
+   :type: string
+   :value: ""
+
+   Delimiter to shorten the Kubernetes context.
```
Rycieos, line 172: Should be obvious, but always better to be explicit. Add something like
> If set to the empty string, no truncating will occur.

Rycieos (changes requested):
This is looking much better!

Only a few nitpicks now. If you squash/rebase/whatever again, I should be able to merge this. It makes sense to me that it should only take one commit for this, but if you can think of some logical breakdown otherwise that's fine too. I just don't like merging in "fix" commits when the commit that broke the thing isn't in the mainline yet, if that makes sense.

Oh, and toss in a few descriptions in the commit message. I'd probably write the commit message like this:

> Add current Kubernetes context data source feature
> 
> Add a data source to get the current Kubernetes context. Includes a
> mechanism for truncating the context to the first matching character,
> configurable.
> 
> The data source is disabled by default. The truncation is also disabled
> by default (disabled by setting the config option to the empty string).
> 
> Add a default theme function to format the context like the other
> development environments.

mschuett:
Yes, I completely agree there should not be dozens of small bugfix commits. :+1: 

I should have let you squash all commits on merge, that would have been cleaner than my forced pushes :-/

liquidpromptrc-dist
```diff
@@ -178,4 +178,12 @@ LP_ENABLE_ERROR=1
...
 LP_DISABLED_VCS_PATHS=()
 
+# Show current Kubernetes kubectl context
+LP_ENABLE_KUBECONTEXT=0
+
+# Delimiter to shorten kubectl context.
+# E.g. when your context names are dev-cluster and test-cluster, set to "-"
+# in order to output "dev" and "test" in prompt.
+LP_DELIMITER_KUBECONTEXT=
+
```
Rycieos, line 188: I just added a conflict here with my latest merge, sorry about that. You'll need to rebase and fix that conflict.

No rush, if anything else comes up before the v2.0.0 release, it could add more conflicts too. I doubt it though, still no bug reports on v2.0.0-rc.1

liquidprompt
```diff
@@ -246,6 +247,8 @@ __lp_source_config() {
     LP_MARK_PREFIX="${LP_MARK_PREFIX:-" "}"
     LP_MARK_PERM="${LP_MARK_PERM:-":"}"
     LP_MARK_DIRSTACK="${LP_MARK_DIRSTACK:-"⚞"}"
+    LP_MARK_KUBECONTEXT=${LP_MARK_KUBECONTEXT:-"☸ "}
+    LP_DELIMITER_KUBECONTEXT=${LP_DELIMITER_KUBECONTEXT:-""}
```
Rycieos, line 251: Pedantic, but this really should go in the first block, after line 196.

Rycieos (changes requested):
Looks (mostly) good! (other than what I broke, again sorry, and what I didn't notice before.)

I just got around to actually testing this, it does work exactly as advertised. Though I must say that the ☸ symbol looks just terrible on my 11 point terminal. I'm not entirely sure what I'll replace it with, but I will be turning this feature on.

liquidpromptrc-dist
Rycieos: v2.0.2 is released, so now is a good time to rebase and fix the few issues. Then this is ready to merge.

mschuett: Thanks for the reminder. I did not get around to it in the last few weeks :(

mschuett:
> I just got around to actually testing this, it does work exactly as advertised. Though I must say that the wheel_of_dharma symbol looks just terrible on my 11 point terminal. I'm not entirely sure what I'll replace it with, but I will be turning this feature on.

I picked it only because it looked good in the Ubuntu Gnome Terminal.
On Windows it is also terrible (here in Windows Terminal and mintty from GitBash):
liquidprompt_kubecontext_windows

I decided to change the mark to an empty default, and only add possible Unicode symbols to the documentation.

docs/config.rst
```diff
@@ -702,6 +734,23 @@ Marks
...
+   :value: ""
+
+   Mark used to prefix the current Kubernetes context (empty per default).
+
+   You can set this to visually distinguish the Kubernetes context from other
+   context fields like the Python virtual environment (see :attr:`LP_ENABLE_VIRTUALENV`)
+   and the Red Hat Software Collection (see :attr:`LP_ENABLE_SCLS`).
+
+   Depending on your Terminal and Font settings, try the characters
+   "⎈" (U+2388, Helm Symbol), "☸" (U+2638, Wheel of Dharma), or "κ" (U+03BA, Greek Small Letter Kappa).
```
Rycieos, line 748: Line too long, keep it to 80 characters max.

docs/config.rst
```diff
@@ -331,6 +351,18 @@ Features
...
 
+.. attribute:: LP_ENABLE_KUBECONTEXT
+   :type: bool
+   :value: 0
+
+   Display the current `Kubernetes <https://kubernetes.io/>`_ `context`_.
+
+   .. _`context`: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/
+
+   See also: :attr:`LP_DELIMITER_KUBECONTEXT`, :attr:`LP_COLOR_KUBECONTEXT`, and :attr:`LP_MARK_KUBECONTEXT`.
```
Rycieos, line 362: Line too long, keep it to 80 characters max.

docs/config.rst
```diff
@@ -177,6 +177,26 @@ General
...
+
+   * if your context names are dev-cluster and test-cluster,
+     then set this to "-" in order to output "dev" and "test" in prompt.
+   * if your context names are dev.k8s.example.com and test.k8s.example.com,
+     then set this to "." in order to output "dev" and "test" in prompt.
+   * if using OpenShift then set this to "/" to show only the project name without the cluster and user parts.
+
+   If set to the empty string no truncating will occur (this is the default).
+
+   See also: :attr:`LP_ENABLE_KUBECONTEXT`, :attr:`LP_COLOR_KUBECONTEXT`, and :attr:`LP_MARK_KUBECONTEXT`.
```
Rycieos, line 196: Line too long, keep it to 80 characters max.

docs/config.rst
```diff
@@ -177,6 +177,26 @@ General
...
+
+   Delimiter to shorten the Kubernetes context.
+
+   Usage example:
+
+   * if your context names are dev-cluster and test-cluster,
+     then set this to "-" in order to output "dev" and "test" in prompt.
+   * if your context names are dev.k8s.example.com and test.k8s.example.com,
+     then set this to "." in order to output "dev" and "test" in prompt.
+   * if using OpenShift then set this to "/" to show only the project name without the cluster and user parts.
```
Rycieos, line 192: Line too long, keep it to 80 characters max.

docs/theme/default.rst
```diff
@@ -180,6 +180,11 @@ default order if the user does not configure a different template.
    The current Python (or Conda) virtual environment. See
    :attr:`LP_ENABLE_VIRTUALENV`.
 
+.. attribute:: LP_KUBECONTEXT
+
+   The current Kubernetes context. See
+   :attr:`LP_ENABLE_KUBECONTEXT`.
+
```
Rycieos, line 187: Add a version added block
```suggestion

   .. versionadded:: 2.1
   
```

liquidprompt
```diff
@@ -2695,6 +2723,12 @@ _lp_default_theme_prompt_data() {
         LP_VENV=
     fi
 
+    if _lp_kubernetes_context_color; then
+        LP_KUBECONTEXT="$lp_kubernetes_context_color"
```
Rycieos, line 2727: This should have a space added to the left, to match other devenvs
```suggestion
        LP_KUBECONTEXT=" $lp_kubernetes_context_color"
```

liquid.theme
```diff
@@ -24,6 +24,7 @@ if [[ "$(locale -k LC_CTYPE | sed -n 's/^charmap="\(.*\)"/\1/p')" == *"UTF-8"* ]
     LP_MARK_STASH="+"          # if git has stashs
     LP_MARK_SHORTEN_PATH=" … " # prompt mark in shortened paths
     LP_MARK_PERM=":"           # separator between host and path
+    LP_MARK_KUBECONTEXT=""     # Kubernetes context
```
Rycieos, line 27: If you are sticking with the empty string as the default mark, I think the changes to this file should be removed. This file is an example, and it does not need every new feature added.

liquidpromptrc-dist
```diff
@@ -154,6 +154,14 @@ LP_ENABLE_VIRTUALENV=1
...
 LP_ENABLE_SCLS=1
 
+# Show current Kubernetes kubectl context
+LP_ENABLE_KUBECONTEXT=0
+
+# Delimiter to shorten kubectl context.
+# E.g. when your context names are dev-cluster and test-cluster, set to "-"
+# in order to output "dev" and "test" in prompt.
+LP_DELIMITER_KUBECONTEXT=
+
```
Rycieos, line 164: This chunk is in this file twice.

Rycieos (changes requested):
> I decided to change the mark to an empty default, and only add possible Unicode symbols to the documentation.

I'm not sure I like that. We already have a few "developer environments" that just show a string without any marks, and having multiple show up without any distinguishing marks could get confusing.

It does work of course without it, and having the option of adding mark is better than not.

I'll leave the choice up to you, but if it was up to me, I would pick U+2388, since it does not have the ugly box around it on Windows. I'm sure each OS will be different though.

Just a few other docs issues also.

mschuett:
:thinking: then let's use U+2388 (⎈) to set the context apart ...

docs/config.rst
Rycieos: Missed this line.

Rycieos (changes requested):
Two small issues.

Rycieos (changes requested):
Looks good! I would just prefer a squash, to 1 or 2 commits. The last commit's message is not ideal, so just squash it into the previous commit.

I can do it if you want, but I don't like doing it with the Github UI, since then it doesn't sign the commit with my signature. And if I do it from the command line, Github won't mark the PR as merged.

mschuett:
Ok, I made another force-push with the same data as before.

Rycieos:
Merged! Thanks for the PR, and the work to fix it up.

Congrats on the first merge in the v2.1 window!

mschuett:
:tada: 

@Rycieos: Thanks a lot for the patient proof-reading!

PR #582, opened 2019-08-16
Fix battery charging symbol

ste-fan:
Return code of `_lp_battery` when charging was [1, 2], now it is the
desired [2, 3].

 *1 => discharging above threshold
 *2 => charging under threshold
 *3 => charging above threshold

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1289,6 +1289,9 @@ case "$LP_OS" in
         if [[ "$acpi" == *"Discharging"* ]]; then
             # under => 0, above => 1
             return $(( bat > LP_BATTERY_THRESHOLD ))
+        # not charging
+        elif [[ "$acpi" == *"Not charging"* ]]; then
+            return 4
         # charging
         else
             # under => 2, above => 3
--- a/liquidprompt
+++ b/liquidprompt
@@ -1292,7 +1292,7 @@ case "$LP_OS" in
         # charging
         else
             # under => 2, above => 3
-            return $(( 1 + ( bat > LP_BATTERY_THRESHOLD ) ))
+            return $(( 2 + ( bat > LP_BATTERY_THRESHOLD ) ))
         fi
     }
     ;;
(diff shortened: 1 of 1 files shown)


ste-fan:
As suggested in #583 I added this commit to make one pull request concerning the battery charging symbol:

**Hide battery symbol when not charging**

When charging thresholds for battery health are set (e.g. on ThinkPads,
the default is never to charge when battery level is at least 96%) it is
common for the battery to be neither charging nor discharging, while not
having 100% charge. Which lead to the green charging symbol being always
displayed.

Example output, in given ThinkPad scenario:
```
$ sudo tlp-stat --battery
--- TLP 1.2.2 --------------------------------------------

+++ Battery Features: Charge Thresholds and Recalibrate
natacpi    = inactive (no kernel support)
tpacpi-bat = active (thresholds, recalibrate)
tp-smapi   = inactive (ThinkPad not supported)

+++ ThinkPad Battery Status: BAT0 (Main / Internal)
/sys/class/power_supply/BAT0/manufacturer                   = SMP
/sys/class/power_supply/BAT0/model_name                     = LNV-45N1
/sys/class/power_supply/BAT0/cycle_count                    = (not supported)
/sys/class/power_supply/BAT0/energy_full_design             =  56250 [mWh]
/sys/class/power_supply/BAT0/energy_full                    =  43670 [mWh]
/sys/class/power_supply/BAT0/energy_now                     =  43610 [mWh]
/sys/class/power_supply/BAT0/power_now                      =      0 [mW]
/sys/class/power_supply/BAT0/status                         = Not charging (threshold effective)

tpacpi-bat.BAT0.startThreshold                              =     96 [%]
tpacpi-bat.BAT0.stopThreshold                               =    100 [%]
tpacpi-bat.BAT0.forceDischarge                              =      0

Charge                                                      =   99.9 [%]
Capacity                                                    =   77.6 [%]

$ acpi
Battery 0: Not charging, 99%
```

PR #583, opened 2019-08-16
Hide battery symbol when not charging

ste-fan:
When charging thresholds are set (e.g. on Thinkpads) there are scenarios
where the battery is neither charging nor discharging, while not having
100% charge. Which lead to the green charging symbol being always
displayed.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1267,12 +1267,13 @@ _lp_wifi()
 # returns 1 (and battery level) if battery is discharging and above threshold
 # returns 2 (and battery level) if battery is charging but under threshold
 # returns 3 (and battery level) if battery is charging and above threshold
-# returns 4 if no battery support
+# returns 4 (and battery level) if battery is neither charging nor discharging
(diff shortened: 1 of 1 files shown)


Rycieos:
There is no linked issue to this PR, you will need to describe the problem you are trying to solve better (acpi output etc).

This seems related to #582, I think they should be together as one PR.

I don't see the point in changing the value returned when there is no battery. Why not only make the additions on lines 1293-1295 and not add a new type returned. The comments you added wouldn't make any sense, but it would achieve the same output.

ste-fan:
Pushed suggested changes to #582 with a bit more context.

My motivation for the the original commit was to make comments reflect the actual behavior. With the new *Not charging* rule it is possible for `_lp_battery` to return both the battery level and 4 in spite of `# returns 4 if no battery support`. Also `# no battery support or battery full: nothing displayed` in `_lp_battery_color` doesn't reflect the possibility of not charging a non-full battery.

PR #587, opened 2019-10-10
fix error when cannot access /proc/loadavg

go2null:
encountered this error on Termux
```
bash: /proc/loadavg: Permission denied
```

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -185,10 +185,16 @@ _LP_SED_EXTENDED=r
 # get current load
 case "$LP_OS" in
     Linux)
-        _lp_cpu_load () {
-            local eol
-            read lp_cpu_load eol < /proc/loadavg
-        }
+        if [ ! -r '/proc/loadavg' ] && command -v >/dev/null 'uptime'; then
+            _lp_cpu_load () {
+                lp_cpu_load="$(uptime | sed 's/.*load average: \([0-9.]*\).*/\1/')"
+            }
+        else
+            _lp_cpu_load () {
+                local eol
+                read lp_cpu_load eol < /proc/loadavg
+            }
+        fi
         ;;
     FreeBSD|Darwin|OpenBSD)
         _lp_cpu_load () {


Rycieos:
I don't think this is a good idea.

If there is an issue reading from `/proc/loadavg`, the user needs to know so that they can fix that. Suppressing the error message will make it much harder to debug.

What even happens if the lookup is skipped? The variable will not have any value in it, which might cause an error somewhere else.

I'd much rather say that users must be able to read from `/proc/loadavg`, since that's pretty standard. Why can't you read it? What reasons might system admins disable that?

go2null:
I changed it to only skip if in Termux.

This issue arises in Termux (linux shell on Android) which doesn't support `/proc`
- https://wiki.termux.com/wiki/Differences_from_Linux
- https://github.com/termux/termux-app/issues/299

Rycieos:
Thanks for the links.

From the issue thread:
> They are effectively abandoning practical Linux compatibility.

I agree. Yikes. IMO, this means, from a shell perspective, Android is no longer Linux.

I don't like where you added this extra check, since it will be run every time for all Linux boxes. We are trying to remove total running code, not add more.

For those two reasons, I see an argument for adding a new OS type, Android. It would add a slight amount of extra startup time, but then we could support all these divergences from normal Linux.

go2null:
> I don't like where you added this extra check, since it will be run every time for all Linux boxes. We are trying to remove total running code, not add more.

I've changed 
- to a more elegant solution - use `uptime` if available, otherwise revert to original behaviour
- created appropriate function definition to avoid checking everytime

Rycieos:
Better, but now you have a situation where if both `/prod/loadavg` is not readable and `uptime` doesn't exist, then `_lp_cpu_load()` won't be defined, which won't be a helpful error message for any user that runs into that.

go2null:
not really, the `uptime` function is only defined BOTH conditions are true, otherwise it uses the normal `/proc/loadavg` parsing.

Rycieos:
Hm, true. I blame Monday.

Seems like an OK workaround then.

adamryczkowski:
I hope this will get pushed mainstream, so I don't have to patch the code manually on each Android device.

liquidprompt
```diff
@@ -185,10 +185,16 @@ _LP_SED_EXTENDED=r
 # get current load
 case "$LP_OS" in
     Linux)
-        _lp_cpu_load () {
-            local eol
-            read lp_cpu_load eol < /proc/loadavg
-        }
+        if [ ! -r '/proc/loadavg' ] && command -v >/dev/null 'uptime'; then
+            _lp_cpu_load () {
+                lp_cpu_load="$(uptime | sed 's/.*load average: \([0-9.]*\).*/\1/')"
```
Rycieos, line 190:
```suggestion
                lp_cpu_load="$(LC_ALL=C uptime | sed 's/.*load average: \([0-9.]*\).*/\1/')"
```

Rycieos (changes requested):
Sorry for the delay, I was on a much needed vacation this past week.

Yikes, it's already been almost a year since this was opened? Lucky for you I guess, since I hate this much less than I did back then.

I'm mostly fine with this, I just need a few simple things (plus the specific suggestion):
1. Rebase to fix the conflicts.
2. Paste me the output of `LC_ALL=C uptime` on your Android machine, so I can verify that it will parse similar to other `uptime` outputs.

As a bonus (otherwise as a note to myself for the future), this should be optimized somehow so as to not duplicate the definition in the `SunOS` section, since it's the same.

Rycieos:
I pulled, rebased, and fixed conflicts to merge this. GitHub won't show it as merged, since the changes won't come from your personal branch.

PR #589, opened 2019-10-23
add ZSH example for newline in PS1

YouthfulIndiscretion:
Add an example in README.md for a newline in ZSH's PS1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,10 @@ To postfix the prompt, use the `LP_PS1_POSTFIX` variable. For example, to add a
 newline and a single character:
 
     LP_PS1_POSTFIX="\n>"
+    
+With ZSH, this example would be:
+
+    LP_PS1_POSTFIX=$'\n'">"
 
 Note: the `prompt_tag` function is a convenient way to add a prefix. You can use
 it to add a keyword to each of your different terminals:


Rycieos:
This is good stuff, though I think the first example (supposedly for Bash) is just wrong. I think regardless of shell it should be a literal newline character. And the example is wrong anyway, `LP_PS1_POSTFIX` is added after everything, including the prompt mark. Making the example render:
```
... $
>
```
None of which is your fault, but why not fix it? It would be better to show this example for `LP_MARK_PREFIX`, which is added right before the prompt mark.

This section got lost on conversion to the new documentation (hence the merge conflict). I'd be happy to merge it if you added it to the correct spot: [docs/config.rst](https://github.com/nojhan/liquidprompt/blob/master/docs/config.rst) at line 50:

```diff
    parts of the prompt. Can be used to tag the prompt in a way that is less
-   intrusive than :attr:`LP_PS1_PREFIX`.
+   intrusive than :attr:`LP_PS1_PREFIX`, or add a newline before the prompt
+   mark. For example::
+
+      LP_MARK_PREFIX=$'\n'

```

Rycieos:
I made that change myself so that it could make it into v2.0.0. Thanks for bringing the issue to my attention!

PR #594, opened 2020-01-27
Added zinit installation instructions

johnhamelink:
I recently made the switch to [zinit](https://github.com/zdharma/zinit), and wanted to keep liquidprompt. One of the cool things `zinit` does (and as far as I can tell it works perfectly with liquidprompt!) is to be able to load plugins asynchronously.

This PR contains the configuration I use to setup liquidprompt asynchronously using zinit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,15 @@ To install via antigen, simply add the following line in your `.zshrc` after act
 
     antigen bundle nojhan/liquidprompt
 
+### Installation via zinit
+
+To install via zinit, simply add the following lines to your `.zshrc` after activating zinit:
+
+    zinit ice load'[[ $- = *i* ]]' \
+	            wait'!' \
+	            lucid nocd \
+	            atload'!_lp_set_prompt'
+    zinit load nojhan/liquidprompt
 
 ## Dependencies
 


Rycieos (changes requested):
Thanks for the PR! Sorry for the delay, it was 2020, and I've been busy working on general rework and documentation.

I want to merge this but I need a few things first:
1. Add this documentation instead to the new install documentation, under the [Antigen section](https://github.com/nojhan/liquidprompt/blob/bc5eebb6c02d1742a945be13d45ab48a25a9f712/docs/install.rst#id15). (sorry for the moving target!) Might look something like this:
```rst
Installation via Zinit
------------------------

To install via `Zinit <https://github.com/zdharma/zinit>`_, simply add the
following line in your ``.zshrc`` after activating Zinit::

   zinit load nojhan/liquidprompt
```
2. Add a link to the Zinit page in the documentation instead of the commit message (see example above).
3. Is that actually the best way to load Liquidprompt with Zinit? I will admit I am no expert, but as part of https://github.com/nojhan/liquidprompt/issues/634#issuecomment-742219591 I found that a simple `zinit load nojhan/liquidprompt` works great, because of our `liquidprompt.plugin.zsh` file (I think anyway). If that `ice load` command does improve it in any way, then obviously:
4. Explain the `ice load` command in the commit message.
5. Replace `_lp_set_prompt` with `__lp_set_prompt`, as that has changed in v2.

I look forward to getting this merged!

johnhamelink:
Thakns @Rycieos, I've created a new PR based on the changes you've described, and against the latest master :)

#654

Issue #445, opened 2016-06-04
ERROR: Can't get value of subfeature temp1_input: Can't read


mikeSimonson:
When I update to the tag v_1.10 I get the error above.
The error is not present in v_1.9.

dolmen:
On which operating system? Linux?

Could you give me the output of those commands:
```
LANG=C sensors -u
LANG=C acpi -t
```

mikeSimonson:
Linux indeed with bash.

I had to install acpi.
I don't know if it's also used by liquidprompt.

The ouptut is coming from version v_1.10.
It's the same on v_1.9 minus the error message.

```bash
$ LANG=C sensors -u
acpitz-virtual-0
Adapter: Virtual device
temp1:
  temp1_input: 48.000
  temp1_crit: 103.000

nouveau-pci-0100
Adapter: PCI adapter
temp1:
ERROR: Can't get value of subfeature temp1_input: Can't read
  temp1_max: 95.000
  temp1_max_hyst: 3.000
  temp1_crit: 105.000
  temp1_crit_hyst: 5.000
  temp1_emergency: 135.000
  temp1_emergency_hyst: 5.000

thinkpad-isa-0000
Adapter: ISA adapter
fan1:
  fan1_input: 2136.000

coretemp-isa-0000
Adapter: ISA adapter
Physical id 0:
  temp1_input: 47.000
  temp1_max: 87.000
  temp1_crit: 105.000
  temp1_crit_alarm: 0.000
Core 0:
  temp2_input: 47.000
  temp2_max: 87.000
  temp2_crit: 105.000
  temp2_crit_alarm: 0.000
Core 1:
  temp3_input: 44.000
  temp3_max: 87.000
  temp3_crit: 105.000
  temp3_crit_alarm: 0.000
Core 2:
  temp4_input: 45.000
  temp4_max: 87.000
  temp4_crit: 105.000
  temp4_crit_alarm: 0.000
Core 3:
  temp5_input: 43.000
  temp5_max: 87.000
  temp5_crit: 105.000
  temp5_crit_alarm: 0.000

ERROR: Can't get value of subfeature temp1_input: Can't read


$ LANG=C acpi -t
Thermal 0: ok, 50.0 degrees C
ERROR: Can't get value of subfeature temp1_input: Can't read
```

dolmen:
[...]

PR #597, opened 2020-03-17
suppress error messages on each prompt

jnweiger:
Seen after an online update to kernel 5.3.0-42 on Linux Mint Tricia.
With this kernel, /usr/bin/sensors -u babbles on stderr, as - yes - my
/sys/class/hwmon/hwmon4/temp2_input is not readable. All other temp*_input are readable.

As part of liquidprompt the output is not helpful. Users have no context. (the command that fails is not mentioned, the /sys/ node that fails is not mentioned). Users get spammed with the error. (seen on every prompt)

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1598,7 +1598,7 @@ if (( LP_ENABLE_TEMP )); then
         # Return the hottest system temperature we get through the sensors command
         # Only the integer part is retained
         local -i i
-        for i in $(sensors -u |
+        for i in $(sensors -u 2>/dev/null |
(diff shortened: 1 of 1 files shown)


Rycieos:
This looks related (duplicate?) to #445. I doubt this will get merged. It's too heavy handed to _always_ suppress errors. Users need to be able to see if something is actually broken.

Ideally you would fix the OS issue that is preventing you from reading the file. See some of the later comments on the issue I linked.

Rycieos:

PR #602, opened 2020-05-09
Fix LP_ENABLE_RUNTIME when autojump is also enabled.

bdejean:


Rycieos:
This needs some description. Although it is implicit, it could also use a description of the issue that this is fixing.

From searching `autojump_add_to_database`, I see that this is to enable compatibility with [wting/autojump](https://github.com/wting/autojump). However, `liquidprompt` has long taken the stance that it won't work to support other prompt integrations.

Also, it has the current stance that we won't add features that will slow down the prompt generation, and/or won't be used by many. See https://github.com/nojhan/liquidprompt/pull/577#issuecomment-542777329.

bdejean:
Could we have an optional customizable array of commands to skip ? I understand that it's not good to have something related to autojump hardcoded.

Rycieos:
This PR has had some extra commits added to it. On top of that, I'm rejecting this PR anyway because it specifically is for compatibility with one project.

> Could we have an optional customizable array of commands to skip ? I understand that it's not good to have something related to autojump hardcoded.

Sure, please open an issue for this feature request. I don't think it will be that simple, but obviously some functionality is needed here.

PR #605, opened 2020-06-11
handle unset _LP_RUNTIME_LAST_SECONDS for non-zsh shells

Habbie:
Fixes https://github.com/NixOS/nix/issues/787

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1448,7 +1448,10 @@ if (( LP_ENABLE_RUNTIME )); then
                 _LP_RUNTIME_SECONDS=-1 _LP_RUNTIME_LAST_SECONDS=$SECONDS
             else
                 # Compute number of seconds since program was started
-                (( _LP_RUNTIME_SECONDS=SECONDS-_LP_RUNTIME_LAST_SECONDS ))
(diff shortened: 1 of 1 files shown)


Habbie:
> In general, Liquidprompt (version 1, the rework will hopefully allow this) doesn't support mixing with other shell prompt modifications.

Understood - that was a secondary issue in that ticket. The crash that this PR fixes is much more important.

> However, as this is a problem with `set -u` compatibility, I like this change. Liquidprompt has a lot of issues with `set -u` right now, so I can't claim it is anywhere close to working with that enabled, but I won't reject patches that get it closer to that goal.

I suspected there might be more spots. I wonder if we can do something to make clear to the user what happened? Perhaps say something via a `trap`? I spent quite some time figuring out this one.

> Note that I don't maintain the packages for Debian (or any other OS). Liquidprompt 1.11 is also 4 years old now, and there have been many fixes since then, now including this one. I'm considering cutting a 1.12 release soon, but my primary focus is on the 2.0 rework.

Debian in general wouldn't pull a fix like this into a stable version anyway. I've switched to running liquidprompt directly from a git checkout :)

Rycieos:
Merged in f55a56b79bab1e7a0f89e0e0430c5f966e35b1ea

Issue #599, opened 2020-04-01
Add a newline if the prompt is too long


Rmano:
Shell: zsh
Operating system: linux Ubuntu 18.04
Liquid Prompt version (tag, commit): 5f4aeece8d6cf98138e729f7833e11e985ca44d3

I am trying to add a newline to the prompt only if the characters left are less than, say, 30. In plain `zsh` [you can use the %-30 prompt directive](https://unix.stackexchange.com/a/537248/52205), so I tried 
```zsh
export LP_PS1_POSTFIX=$'%-30(l:\n> :)'
```

...but it is doing nothing. Is there a correct way to do this?

Thanks for the nice software!

Rycieos:
I'm assuming that that type of expansion doesn't work because we set the `nopromptsubst` option:
https://github.com/nojhan/liquidprompt/blob/af63c4cfb8afe3f3161f3ace4aba3b940847a7a8/liquidprompt#L1942-L1943

You could try enabling it yourself and seeing if it will then expand this option.

I like this idea though, and want to see it added as a feature. This means we couldn't rely on zsh specific things obviously.

Rmano:
I tried this in my `.zshrc`:

```zsh
liquidprompt=$HOME/software/liquidprompt/liquidprompt
[[ -f $liquidprompt ]] && source $liquidprompt
# see https://github.com/nojhan/liquidprompt/issues/599
setopt promptsubst
export LP_PS1_POSTFIX=$'%-30(l:\n> :)'
```
...but still not working. Well, I'll wait...

nojhan:
What are your ideas about the following options for generalizing @PandaWill's idea?

1. Allow to automatically switch theme when the current prompt overflows the terminal width. Cons: this would need some specific themes to actually implement the prompt shortening (like a heavily multi-line theme); this allows for only two options. Pros: this is very simple, with the current code; this may be generalized later as an automatic theme switcher ("if this state matches this, switch to this theme").
2. Allow for *several*, ordered, breakable markers, which would be activated if the prompt overflows. Themes would need to be updated to add markers (I'm thinking something like `$LP_BREAK_1`, `$LP_BREAK_2`), which would be replaced either by an empty space or a new line. Pros: this allows for adaptation to any terminal width. Cons: this is more complicated to implement (need to check that themes do not use duplicate breaks, for instance) which may hinder speed.

Rycieos:
> What are your ideas about the following options for generalizing @PandaWill's idea?

[...]

Issue #609, opened 2020-06-25
Make title bar work out-of-the-box with Gnome Terminal


ottok:
Shell: bash
Operating system: Ubuntu 20.04
Liquid Prompt version (tag, commit): 1.11 Debian revision 3 (includes customized settings, see https://sources.debian.org/src/liquidprompt/1.11-3/debian/patches/debian-default-cfg.patch/)

I recently learned about liquidprompt on the [Debian developer mailing list](https://lists.debian.org/debian-devel/2020/06/msg00055.html) and I love it after using for some weeks now.

The only problem seems to be that the Gnome Terminal window title is not set by default:
![image](https://user-images.githubusercontent.com/668724/85776795-b7001700-b729-11ea-97bb-b6eceef857c9.png)

I can enable it manually with `LP_ENABLE_TITLE=1` but it does not feel as usable as the rest of liquidprompt since the title bar has excess info that is not relevant for a title bar:
![image](https://user-images.githubusercontent.com/668724/85777062-f3337780-b729-11ea-8069-4afb7e3130d3.png)

In the title bar it would be enough to show what Bash shows by default (mainly user, host and path). Having time or temperature or git branch etc in the title bar does not feel very useful.

Could liquidprompt be made so that it does not override the title with *nothing* and instead keeps the default bash title there?

If feel that would be the optimal solution from an usability point of view.

This is partially a duplicate of #78 but I thought a new issue would be in place for discussing further development of the title bar feature.

Thanks for a great piece of software!

ottok:
My personal workaround for this was to add in `.bashrc` the line `PROMPT_COMMAND='echo -ne "\033]0;${USER}@${HOSTNAME}: ${PWD/$HOME/~}\007"'`

![image](https://user-images.githubusercontent.com/668724/85779212-0cd5be80-b72c-11ea-9a55-3e285f08639b.png)

Rycieos:
> keep the default bash title there

I'm not sure that there is a "default bash title". Bash by default doesn't interact with a terminal or terminal emulator much, if at all. For it to do so, you need to print out ANSI escape codes to tell the terminal to do something. My guess is that there is some system PS1 or similar in `/etc/bashrc` or similar on your system that sets a nice default for a terminal title. I'd be happy to be proven wrong though.

What you see right now ("Terminal") is what happens when no terminal is set, which is the default for Gnome terminal. For Mintty, the default title is the launched command (`-bash` in my case).

[...]

Issue #608, opened 2020-06-15
Ksh compatibility


Rycieos:
This issue is to log research and work done to make Liquidprompt compatible with more shells.

Right now, Liquidprompt claims to work with Bash and Zsh only, which it does well. There are no trackers in Liquidprompt, so there is no way to know the Bash/Zsh user share, but I would guess that the Bash user share is higher (since Zsh users often prefer Oh My Zsh), and quick scan of self reporting from the issues would back that up.

Since (hopefully) a stronger test suite will be coming soon, it makes sense to evaluate what other shells Liquidprompt can support without extra work.

Classic Bourne shell should be out, since it's so hard to find now and doesn't support POSIX.
POSIX shells like ash and its derivative dash are unlikely to be good ideas, since they try to stick closely to the POSIX standard. This is good for speed, but ironically, since they ship with less features, many of our tests and checks would have to fork out to `awk` or `sed`, which would actually be slower. They are also not designed to be used as login shells, so it's probably not a good idea anyway.

Ksh is a possible candidate though. While ksh93 is obviously from 1993, it has been patched since then to work on many systems, and it is the only shell that some systems ship with. There are also some derivatives (other than bash) like pdksh and mksh. While I don't personally like or use ksh, it seems that Liquidprompt could work in it with little modifications.

Input is welcome from all.

Rycieos:
Here is what I have found so far:
1. Ksh93 is surprisingly feature rich. It's quite close to Bash in most areas. Ksh88 is not so much, but it is also very hard to find. From now on, when I say Ksh I mean Ksh93.
2. Ksh lacks `local` and `declare`, but has `typeset`. Bash and Zsh have all three, but they all are aliases. While Liquidprompt mostly uses `local`, it also uses `typeset` a few times, and a simple `s/local/typeset/` should just work.
3. Ksh only treats `typeset`ed vars as local if they are declared in a `function name { ... }` style function. The `name () { ... }` style is provided for compatibility with Bourne shell, and therefore doesn't have local variables. Bash and Zsh don't work like this as they didn't try to keep Bourne compatibility a top priority. In Bash and Zsh the different styles are the exact same, so there would be no downsides to switching.
4. Not a drawback, just noting that Ksh supports all the `[[ ... ]]` style tests that we are using from Bash and Zsh.
[...]

PR #617, opened 2020-09-09
Add option to insert newline when prompt is long

PandaWill:
Add new setting `LP_MINIMUM_CMD_LENGTH`, which controls when a newline is added. Setting defaults to 0 (disabled).

See feature request #599.

I had to use perl for removing the non-printable characters from the prompt. It doesn't feel like the nicest solution, but I couldn't remove them with bash pattern matching or sed. This was due to the requirement to match the 2 characters at the end of each block: `\]` .

Let me know if there's anything you'd like changed, I'm keen to do all I can to get this merged. I only discovered this project recently and it's almost perfect - this is the missing piece of the puzzle for me!

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -299,6 +299,7 @@ _lp_source_config()
     LP_PS1=${LP_PS1:-""}
     LP_PS1_PREFIX=${LP_PS1_PREFIX:-""}
     LP_PS1_POSTFIX=${LP_PS1_POSTFIX:-""}
+    LP_MINIMUM_SPACE_AFTER_PROMPT=${LP_MINIMUM_SPACE_AFTER_PROMPT:-0}
 
     LP_ENABLE_PERM=${LP_ENABLE_PERM:-1}
     LP_ENABLE_SHORTEN_PATH=${LP_ENABLE_SHORTEN_PATH:-1}
@@ -1594,6 +1595,23 @@ _lp_as_text()
     echo -nE "$1" | sed -$_LP_SED_EXTENDED "s,$_LP_CLEAN_ESC,,g"
 }
 
+# Get length of a given string, after evaluating escape codes (\u for user, etc)
+_lp_length()
+{
+    typeset len evaluated_ps1
+    typeset printable_ps1=$(_lp_as_text "$1")
+
+    if $_LP_SHELL_bash; then
+        evaluated_ps1=$(printf '%s' "${printable_ps1@P}")
+    else # zsh
+        evaluated_ps1=$(print -P $printable_ps1)
+    fi
+
+    len=${#evaluated_ps1}
+
+    echo $len
+}
+
 _lp_title()
 {
     (( LP_ENABLE_TITLE )) || return
@@ -1849,8 +1867,18 @@ _lp_set_prompt()
         # is set.
         PS1+="${LP_VCS}"
 
-        # add return code and prompt mark
-        PS1+="${LP_RUNTIME}${LP_ERR}${LP_MARK_PREFIX}${LP_COLOR_MARK}${LP_MARK}${LP_PS1_POSTFIX}"
+        # add return code
+        PS1+="${LP_RUNTIME}${LP_ERR}"
+
+        # determine if we need to add new line before prompt mark
+        if (( LP_MINIMUM_SPACE_AFTER_PROMPT )); then
+            if [[ $(_remaining_space) -lt "$LP_MINIMUM_SPACE_AFTER_PROMPT" ]]; then
+                PS1+=$'\n'
+            fi
+        fi
+
+        # add prompt mark
+        PS1+="${LP_MARK_PREFIX}${LP_COLOR_MARK}${LP_MARK}${LP_PS1_POSTFIX}"
 
         # "invisible" parts
         # Get the current prompt on the fly and make it a title
@@ -1866,6 +1894,17 @@ _lp_set_prompt()
     else
         PS1=$LP_PS1
     fi
+}
+
+# Calculates the remaining space on a terminal line after prompt has been printed
+_remaining_space()
+{
+    typeset len=$(_lp_length "$PS1")
+
+    # Modulo terminal width, so we handle the case where prompt is already on a new line
+    remaining_space=$(( $COLUMNS - ( $len % $COLUMNS ) ))
+
+    echo $remaining_space
 }
 
 prompt_tag()
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -96,6 +96,7 @@ Thanks again to everyone for all of the help and support!
 * Tristan Miller (@logological)
 * Markus Gebert (@mgeb)
 * Tore Anderson (@toreanderson)
+* William S. (@PandaWill)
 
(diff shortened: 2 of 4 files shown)


liquidprompt
```diff
@@ -1868,6 +1877,20 @@ _lp_set_prompt()
     fi
 }
 
+_remaining_space()
```
Rycieos, line 1880: This is pretty much implemented already in `_lp_as_text()`.

Besides, adding a `perl` dependency is a no-go. Not to mention I'm pretty sure this will only work for Bash, see how `_lp_as_text()` does this:

https://github.com/nojhan/liquidprompt/blob/deff598f30f097279d6f6959ba49441923dec041/liquidprompt#L1583-L1595

liquidprompt
```diff
@@ -1849,6 +1850,14 @@ _lp_set_prompt()
         # is set.
         PS1+="${LP_VCS}"
 
+        if [[ "$LP_MINIMUM_CMD_LENGTH" -ne 0 ]]; then
+            if [[ $(_remaining_space) -lt "$LP_MINIMUM_CMD_LENGTH" ]]; then
```
Rycieos, line 1854: No need for a new function: `_lp_as_text()` does the stripping already:
```suggestion
            local printable_ps1=$(_lp_as_text "$PS1")
            if [[ ${#printable_ps1} -lt "$LP_MINIMUM_CMD_LENGTH" ]]; then
```

Rycieos (changes requested):
Thanks for the PR! As I'm sure you saw on the feature request, I agree this is a needed feature.

A few things to note:

Even with this check, there a quite a few scenarios where this will not work right or make it worse. This ignores how wide your terminal is, and could add a newline where it doesn't make sense at all. A better implementation would decide to wrap based on how many columns are left, not how many have been taken. This is obviously harder. Imagine a scenario where the prompt is 60 characters, but the terminal is 55 characters. Should you add a newline? Probably not.

I wasn't planning on adding this feature until v2.1, as I wanted other reworking of stuff to happen first. Not saying that I can't merge this before then, only that I want this to look pretty close to the final form, not some partial thing that adds more work for v2.0.

Anyway, let's start with these changes and see if we can't come up with a solution for my above hypothetical scenarios.

PandaWill:
Thanks for the quick response.

> This ignores how wide your terminal is, and could add a newline where it doesn't make sense at all. A better implementation would decide to wrap based on how many columns are left, not how many have been taken. This is obviously harder. Imagine a scenario where the prompt is 60 characters, but the terminal is 55 characters. 

I actually used $COLUMNS to take into account the terminal width and handled the edge cases in the `_remaining_space` function. I calculate remaining space as: 
```
term_width - (prompt_length % term_width)
```
So for a prompt length of 60 and terminal 55, remaining space would be 50 `= 55 - (60 % 55)`.

However, I admit it wasn't clear from the variable names or documentation that this what I was doing, so I've hopefully made that clearer.

I've taken perl out and used `_lp_as_text` as recommended.

I also added an `_lp_length` function to make it even more explicit what I'm doing. I had to break this down by shell type as I couldn't figure out a bash and zsh compatible way of expanding the prompt escape codes (e.g. \u for user). We need to expand them to count accurately.

I'm a bash user, but I've tested it with zsh and it seems to be working as I expect.

liquidprompt
```diff
@@ -1594,6 +1595,23 @@ _lp_as_text()
     echo -nE "$1" | sed -$_LP_SED_EXTENDED "s,$_LP_CLEAN_ESC,,g"
 }
 
+# Get length of a given string, after evaluating escape codes (\u for user, etc)
+_lp_length()
+{
+    typeset len evaluated_ps1
+    typeset printable_ps1=$(_lp_as_text "$1")
```
Rycieos, line 1602: This (typeset and assignment on the same line) is unsafe in zsh < 5.1:
```zsh
% func() { 
  typeset var=$(echo foo bar)
  echo "$var"
}
% func
foo
```
Either wrap in quotes or break into two lines:
```suggestion
    typeset printable_ps1="$(_lp_as_text "$1")"
```
Check for other issues like this, I see a few other instances.

liquidprompt
```diff
@@ -1594,6 +1595,23 @@ _lp_as_text()
...
+
+    if $_LP_SHELL_bash; then
+        evaluated_ps1=$(printf '%s' "${printable_ps1@P}")
```
Rycieos, line 1605: This is a Bash-4.4+ feature:
https://github.com/bminor/bash/blob/76404c85d492c001f59f2644074333ffb7608532/CHANGES#L2400-L2401
Liquidprompt supports Bash-3.2+, so this is a no-go.

How to evaluate escape codes is a known problem in Liquidprompt. See:
* Terminal title: https://github.com/nojhan/liquidprompt/issues/609#issuecomment-751882825
* Ksh support: https://github.com/nojhan/liquidprompt/issues/608#issuecomment-645592979

On the surface it is simple, and should not be hard to fix, but I think there could be issues that spring up when we try to replace all escape codes with strings.

Anyway, that is on the top of my to-do list for v2.1. I don't see a way to make this PR work without that first, so you might need to wait...

liquidprompt
```diff
@@ -1594,6 +1595,23 @@ _lp_as_text()
...
+    else # zsh
+        evaluated_ps1=$(print -P $printable_ps1)
+    fi
+
+    len=${#evaluated_ps1}
+
+    echo $len
```
Rycieos, line 1612: Post-rework Liquidprompt avoids using stdout to return values. Use a variable, named after the function:
```suggestion
    lp_prompt_length=${#evaluated_ps1}
```
Oh and name the function `__lp_prompt_length()`. The name is too generic, and lies about what it does.

liquidprompt
```diff
@@ -1849,8 +1867,18 @@ _lp_set_prompt()
...
+
+        # determine if we need to add new line before prompt mark
+        if (( LP_MINIMUM_SPACE_AFTER_PROMPT )); then
+            if [[ $(_remaining_space) -lt "$LP_MINIMUM_SPACE_AFTER_PROMPT" ]]; then
+                PS1+=$'\n'
+            fi
+        fi
+
+        # add prompt mark
+        PS1+="${LP_MARK_PREFIX}${LP_COLOR_MARK}${LP_MARK}${LP_PS1_POSTFIX}"
```
Rycieos, line 1881: This part is elegant.

The problem is that this section is possibly in the user's control. If the user has a template file (`$LP_PS1_FILE`), this block will never be run.

If we are going to advertise this as a feature, it needs to work no matter what template or theme the user is running. I'm not sure at the moment how to fix that, something to worry about later after the other issues are fixed maybe.

Rycieos (changes requested):
Thanks for updating this PR. Unfortunately there is no way to merge this without first a rebase onto the current `master`. There have been too many changes since then.

I'm happy to help if you run into issues with rebasing, just let me know. You will want to look at the new `docs/` folder to add documentation, and if you still need to add a new function like `_remaining_space()`, you should add tests for it in `tests/test_utils.sh`.

Rmano:
If I can help (by testing, basically) this feature on `zsh`, just tell me. It will be a great feature....

Issue #607, opened 2020-06-14
Subshell and function calls speedup


Rycieos:
This issue is to log research and work done to speed up the prompt, specifically with calling shell functions with subshells and other binaries.

Input is welcome from all.

Rycieos:
References for some of my claims can be found here, but I think I back up my claims with testing enough.
http://mywiki.wooledge.org/BashFAQ/084
http://mywiki.wooledge.org/SubShell

```bash
# This is a real subshell + process run from _lp_jobcount_color()
$ test() { for i in {1..1000}; do var="$(jobs -s | wc -l)"; done }
$ time test
real    0m2.757s
user    0m0.959s
sys     0m2.660s
# Replace wc with parameter substitution
$ test() { for i in {1..1000}; do var="$(jobs -s)"; var="${var//[!$'\n']/}"; var=$((${#var} + 1)); done }
$ time test
real    0m0.922s
user    0m0.377s
sys     0m0.564s
# Refactor to a line_count method that echos, which means adding a subshell back
$ line_count() { var="${1//[!$'\n']/}"; echo $((${#var} + 1)); }
$ test() { for i in {1..1000}; do var="$(jobs -s)"; var=$(line_count "$var"); done }
$ time test
real    0m1.829s
user    0m0.757s
sys     0m1.181s
# Replace echo with return-by-variable, removing the subshell
$ line_count() { var="${1//[!$'\n']/}"; count=$((${#var} + 1)); }
$ test() { for i in {1..1000}; do var="$(jobs -s)"; line_count "$var"; var=$count; done }
$ time test
real    0m0.929s
user    0m0.436s
sys     0m0.544s
# Try using array expansion instead of parameter substitution; same result
$ line_count() { set -f; IFS=$'\n'; var=(${var}); count=${#var[@]}; }
$ time test
real    0m0.910s
user    0m0.417s
sys     0m0.544s
# Implementation with reading from stdin. This forces line_count back in a subshell
$ line_count() { var="$(</dev/stdin)"; var="${var//[!$'\n']/}"; count=$((${#var} + 1)); }
$ test() { for i in {1..1000}; do var="$(jobs -s | line_count)"; var=$count; done }
$ time test
real    0m2.947s
user    0m1.383s
sys     0m2.370s
```

Looks to me that we can save quite a bit of time not calling to `wc`. Assuming we can get an implementation that works well on both Bash and Zsh, this seems to be an easy place to save some time.

[...]

PR #623, opened 2020-10-02
Implement temperature monitoring with direct sysfs access

rfc1036:
I am proposing an implementation of temperature monitoring which does not require executing other binaries (as is being discussed in https://github.com/nojhan/liquidprompt/issues/607).
For the time being users need to set `LP_TEMP_HWMON` to a value like `hwmon4/temp1_input` after checking which sensor makes more sense for them, but with some more work it would not be hard to use e.g. the first CPU core as the default sensor.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1504,6 +1504,17 @@ if (( LP_ENABLE_TEMP )); then
 
     # Backends for TEMP. Each backend must return the result in $temperature.
 
+    # Implementation directly accessing sysfs
+    _lp_temp_hwmon() {
+      [ "$LP_TEMP_HWMON" ] || return 0
+      [ -e "/sys/class/hwmon/$LP_TEMP_HWMON" ] || return 0
+
+      local -i temp
+      read temp < /sys/class/hwmon/$LP_TEMP_HWMON
+      [ "$temp" ] || return 0
+      temperature=$((temp / 1000))
+    }
+
     # Implementation using lm-sensors
     _lp_temp_sensors()
     {
@@ -1555,7 +1566,7 @@ if (( LP_ENABLE_TEMP )); then
 
     # Try each _lp_temp method
     # If no function worked, disable the feature
-    _lp_temp_detect acpi sensors || LP_ENABLE_TEMP=0
+    _lp_temp_detect hwmon acpi sensors || LP_ENABLE_TEMP=0
     unset -f _lp_temp_detect
 fi
 
--- a/liquidpromptrc-dist
+++ b/liquidpromptrc-dist
@@ -30,6 +30,10 @@ LP_LOAD_THRESHOLD=60
 # Recommended value is 60
 LP_TEMP_THRESHOLD=60
 
+# The path to a file relative to /sys/class/hwmon/, containing a
+# temperature reading expressed in 1/1000 degrees.
+LP_TEMP_HWMON=""
+
 # Use the shorten path feature if the path is too long to fit in the prompt
 # line.
 # Recommended value is 1


Rycieos (changes requested):
This looks quite good. What remains is to test this on many systems, test on MacOS and SunOS (I doubt it will work, I don't think they have the `/sys/` interface), and find what kernel version added this interface.

The other sticking point is right now Liquidprompt looks at all sensors (specifically their `temp1_input`) and displays the highest one. This only reads one sensor. I'd like to keep that multi-sensor functionality, or even better, have one function to get the highest CPU temp, and one for the highest GPU temp. This might not be possible "out of the box".

rfc1036:
This is strictly Linux-only. The hwmon class has been introduced by the kernel a very long time ago, but drivers for specific hardware have been introduced in different releases.

The same algorithm which chooses the highest value reported can be implemented, but I would still want to be able to configure monitoring just a specific sensor. I do not have a system with a discrete GPU so I do not know if they report temperature using an hwmon class driver or what else.

Rycieos:
With the news in #633, I want to get this working more than ever.

This appears to be the documentation for this kernel interface:
https://www.kernel.org/doc/Documentation/hwmon/sysfs-interface
I can't find any info on first supported version.

It all looks quite straightforward, except this bit:
> Some chips [...] report the temperature measurement as a voltage. Converting this voltage back to a temperature [...] must occur in user space.

This is troublesome for two reasons:
1. That means we need to convert, and if the kernel can't do it, what are the chances a shell script can?
2. I don't see anything in the docs describing how to tell if the interface is returning a temperature or a voltage.

Ignoring that issue, I see two other problems:
1. Your implementation doesn't have a default value. I want every data function to work out-of-the-box, without configuration.
2. Your implementation doesn't support reading from more than one input.

Here is my suggestion:
```bash
_lp_temp_hwmon() {
    [[ -e /sys/class/hwmon/ ]] || return

    local file temp highest_temp=0
    for file in /sys/class/hwmon/*/temp*_input; do
        IFS= read temp <"$file"
        (( temp > highest_temp )) && highest_temp=$temp
    done
    temperature=$((highest_temp / 1000))
}
```

I realize this removes the ability for a user to not read some inputs. I'm open to suggestions. Or here's another idea:
```bash
# Done when reading config:
# If array not already set
[[ -z ${LP_TEMP_INPUT_FILES[@]+x} ]] && LP_TEMP_INPUT_FILES=( /sys/class/hwmon/*/temp*_input )
...
_lp_temp_hwmon() {
    [[ -e /sys/class/hwmon/ ]] || return

    local file temp highest_temp=0
    for file in "${LP_TEMP_INPUT_FILES[@]}"; do
        IFS= read temp <"$file"
        (( temp > highest_temp )) && highest_temp=$temp
    done
    temperature=$((highest_temp / 1000))
}
```

Let's get some discussion going on this.

Rycieos:
I took a look at [psutil](https://pypi.org/project/psutil/), the leading Python library for reading sensors, and here is how they do it on Linux:

https://github.com/giampaolo/psutil/blob/a5eeaadb7fd07b855d7e9a98659ce102ab904ee4/psutil/_pslinux.py#L1239-L1260

So it seems that we could read:
```bash
LP_TEMP_INPUT_FILES=(
    /sys/class/hwmon/hwmon*/temp*_input
    # CentOS has an intermediate /device directory:
    /sys/class/hwmon/hwmon*/device/temp*_input
    /sys/devices/platform/coretemp.*/hwmon/hwmon*/temp*_input
    # Older, fallback option
    /sys/class/thermal/thermal_zone*/temp
)
```

This gets more tricky, as by default, if the shell doesn't find any matches for the glob, it returns the string:
```bash
$ printf '%q\n' "${LP_TEMP_INPUT_FILES[@]}"
/sys/class/hwmon/hwmon0/temp1_input
/sys/class/hwmon/hwmon1/temp1_input
/sys/class/hwmon/hwmon2/temp1_input
/sys/class/hwmon/hwmon\*/device/temp\*_input
/sys/devices/platform/coretemp.\*/hwmon/hwmon\*/temp\*_input
/sys/class/thermal/thermal_zone0/temp
```

But the good news is that psutil ignores the possibility of hwmon/temp returning a voltage instead of Celsius, so I guess we can too?

PR #624, opened 2020-10-02
sysfs-based battery status (proof of concept)

rfc1036:
This is a have a proof of concept implementation of battery monitoring without calling other binaries. It works fine as is, but it needs to be somehow integrated in liquidprompt and I do not know how you would want to do this.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -467,6 +467,7 @@ _lp_require_tool BZR bzr
 if [[ "$LP_OS" = Darwin ]]; then
     _lp_require_tool BATT pmset
 else
+    false && \
     _lp_require_tool BATT acpi
 fi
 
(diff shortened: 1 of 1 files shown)


Rycieos:
Sorry for the delay.

This is good stuff. There are some minor tweaks I would want to get this to fit in with the rework state of Liquidprompt, but I'll make those myself.

The bigger issue here is compatibility.
1. When/what version was this kernel interface added?
2. Did it ever change?
3. Can we depend on `/sys/class/power_supply/BAT0/status` always being `Discharging` when discharging, or did a different version ever have `discharging`?
4. Did the above ever have extra stuff besides one word?
5. What is the magical `BAT0` in the path, and will that ever differ between machines?
6. What about multiple batteries?

If the answers are:

1. More than 20 years ago
2. No
3. Yes we can
4. No
5. No it won't
6. This is an interface over all batteries together

Then this is good to go. But I'm guessing some of those answers are different, and we need to figure out what that means for this data function.

If you can track down some of those answers, we can get closer to merging this.

Rycieos:
My questions above still stand. And we should probably add an interface like what we have for temperature ([`_lp_temperature()`](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/functions/data.html#lp_temperature) and [`__lp_temp_detect()`](https://liquidprompt.readthedocs.io/en/v2.0.0-rc.1/functions/internal.html#lp_temp_detect)) so we can support more than one backend.

todorowww:
I want to add my $0.02 to this discussion.

For some weird reason, on my system, it's `BAT1`, instead of `BAT0`. Status returns `Discharging`, so that's consistent.

Also, since I have multiple batteries reported by ACPI, here, I have `ACAD` (type `Mains`) `BAT1`, `hidpp_battery_2` and `hidd_battery_3` for my keyboard and mouse batteries.

Rycieos:
Closing as the work in #689 has resolved this.

PR #627, opened 2020-10-26
Display current mercurial bookmark, if any

milouse:
I would like to propose this new version of #403. Instead of using sed around a mercurial command, it directly use the output of a well crafted mercurial command, what seems to be a bit more logical? I may have missed some things, as I’m not an heavy user of mercurial…

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -990,8 +990,12 @@ _lp_hg_branch()
 
     local branch
     branch="$(hg branch 2>/dev/null)"
-
-    (( $? == 0 )) && _lp_escape "$branch"
+    if [ $? -eq 0 ]; then
+        local bookmark
+        bookmark="$(hg bookmarks -ql . 2>/dev/null)"
+        [ "z$bookmark" != 'z' ] && branch="$branch/$bookmark"
+    fi
+    _lp_escape "$branch"
 }
 
 # Set a color depending on the branch state:


Rycieos:
I won't be merging exactly as is (since this section is [completely changed in the upcoming rework](https://github.com/nojhan/liquidprompt/blob/358f38c81b047b5341ec6479b67d41570b4ee405/liquidprompt#L1009-L1021) (I'm working on it as I type this)), but I want this as a discussion point.

I'm trying to make it possible (not required) to display all VCS providers in the same way. Git has historically had the most information dumped into the prompt, while others have nothing except the branch name. Not all VCSs can provide the detail that Git can, but most can get close.

Anyway, this `"$branch/$bookmark"` format is different than what we are doing for Git right now. Not saying that we can't make changes to how Git is displayed, but I want to think it through. Here is what I am doing for branch display:
https://github.com/nojhan/liquidprompt/blob/358f38c81b047b5341ec6479b67d41570b4ee405/liquidprompt#L1936-L1944

I haven't finished thinking about how to return `hg` tags. Just spitballing here, maybe we can have `_lp_hg_tag()` return a matching bookmark instead of a tag? Thoughts?

milouse:
In fact hg tags and bookmarks are very different things. Tags are more or less the same thing than for git: a fixed point in the history. Bookmarks are more or less the same thing than named branch. They are used in a similar context. The difference is that by default, bookmarks are not pushed on the server. They can be used to begin a work on top of a shared named-branch on your machine. In mercurial, branches must be seen as a permanent data bind to any commit added to it. Contrary to git, branches must not be seen as disposable label put on a given path at a given time. That’s why bookmarks may be interesting to avoid polluting the branch namespace with temporary works.

That’s why it makes sense to display both branch and bookmark in the same time: it shows you from where you start (branch) and along which path you are walking (bookmark).

Rycieos:
Thanks for the explanation. I've cherry-picked this in my current rework as https://github.com/nojhan/liquidprompt/pull/630/commits/83e52da1448486a37ffe43e52cbf61dffad2d880. Does that look good?

If you want to test what I've been working on it's all on the `rework` branch.

Unrelated, I took a look at your https://github.com/milouse/liquidprompt/commit/2cfda5d545011e3bb027bdedcadaafbeb8bbe970 commit, and I don't think I want to pull that. That `--numstat` flag was added June 2017, which is really too new, and I'm not convinced it is safer. Any file with a newline in the name followed by a number would break this (ex `foobar\n10`).

Rycieos:
Cherry-picked as f4636e66455a80586f20bb1ea9624a15299cea58 and now in [v2.0.0-beta.1](https://github.com/nojhan/liquidprompt/releases/tag/v2.0.0-beta.1). Thanks for the implementation!

Added as [`_lp_hg_bookmark()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.1/functions/data/vcs.html#lp_hg_bookmark), and the use of that data function is in [`_lp_vcs_details_color()`](https://liquidprompt.readthedocs.io/en/v2.0.0-beta.1/functions/theme.html#lp_vcs_details_color).

PR #628, opened 2020-10-26
Feature/rvm support

milouse:
Reopen #435, this time rebased on master.


Rycieos:

milouse:
Sorry for the delay. I finally take time to rebase / update this MR toward the new version.

I never write any rst file before, thus you may have to check twice what I write in the documentation files.

Thank you for your patience.

liquidprompt
```diff
@@ -1133,6 +1135,28 @@ _lp_python_env_color() {
...
+_lp_ruby_env() {
+    (( LP_ENABLE_RUBY_VENV )) || return 2
+
+    local ret
+
+    if command -v rvm-prompt >/dev/null ; then
+        __lp_escape "$(eval rvm-prompt $LP_RUBY_RVM_PROMPT_OPTIONS)"
+        lp_ruby_env=$ret
+    elif command -v rbenv >/dev/null ; then
+        __lp_escape "$(rbenv version | sed -e 's/ (set.*$//' -e 's/^ruby-//')"
```
Rycieos, line 1147: Running `sed` is slow. Since this looks like a simple pattern removal, could this be done with parameter expansion instead?
```suggestion
        lp_ruby_env=$(rbenv version)
        lp_ruby_env=${lp_ruby_env%% (set*}
        __lp_escape "${lp_ruby_env#ruby-}"
```

Rycieos (changes requested):
This is looking good, thanks for the work to fix it up! The docs look great.

A few small issues with getting the data out of the tools. I will admit that I don't use Ruby, so I don't have either of these tools installed, but I will try to actually test before merging.

liquidprompt
```diff
@@ -1133,6 +1135,28 @@ _lp_python_env_color() {
     lp_python_env_color="[${LP_COLOR_VIRTUALENV}${lp_python_env}${NO_COL}]"
 }
 
+_lp_ruby_env() {
+    (( LP_ENABLE_RUBY_VENV )) || return 2
+
+    local ret
+
+    if command -v rvm-prompt >/dev/null ; then
+        __lp_escape "$(eval rvm-prompt $LP_RUBY_RVM_PROMPT_OPTIONS)"
```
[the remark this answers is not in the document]
milouse: See my above comment to your comment https://github.com/nojhan/liquidprompt/pull/628#discussion_r608785556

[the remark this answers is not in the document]
milouse: I saw that during my migration. In fact `rvm-prompt` args need to be separated. However, storing them in a  unique variable make any call to `rvm-prompt` to fail as it seems to be enable to split them correctly. The only (ugly) way I found yet was this eval thing. Historically it was not a problem as the old implementation of liquidprompt was doing an eval sometime later in the stack.

However, you are right we should avoid this eval thing. I’ll try to see if it works when the variable is directly a shell array.

liquidprompt
milouse: I must admit I never use rbenv myself. This line comes directly from the first comment of the related issue. I’ll try to review how rbenv works and see if I can improve it.

liquidprompt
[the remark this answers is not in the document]
Rycieos: > I’ll try to see if it works when the variable is directly a shell array.

Oh, yeah, a shell array is the right way to do this. Not the IFS hack that I came up with.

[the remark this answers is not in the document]
milouse: Should be fixed by f53782d7a0adc07d9ed9eda2de383e3168d65f62

liquidprompt
milouse: I take your proposal and commited it in b5fe8d6d5b558fd163155fd68dc721213bfbd74e

I kept only the removal of the ` (set…)` part of the output, as the `ruby-` indication may be usefull (to discriminate against jruby, truffle ruby…) and more homogeneous with default RVM output.

milouse:
Should be ok this time? :)

liquidprompt
```diff
@@ -1133,6 +1134,31 @@ _lp_python_env_color() {
     lp_python_env_color="[${LP_COLOR_VIRTUALENV}${lp_python_env}${NO_COL}]"
 }
 
+_lp_ruby_env() {
+    (( LP_ENABLE_RUBY_VENV )) || return 2
+
+    local ret
+
+    if command -v rvm-prompt >/dev/null ; then
+        [ -z "${LP_RUBY_RVM_PROMPT_OPTIONS-}" ] && LP_RUBY_RVM_PROMPT_OPTIONS=(i v g s)
```
Rycieos, line 1143: This is a config option, so it should be set to a default after line 218.

Detecting an empty or not set array is tricky across all of the shell versions we support (see https://stackoverflow.com/a/61551944/8338372), see line 226 for an example of how to do it correctly. This is more complex, since if the array is not set, we want to give it a default value, not leave it as an empty array. I think this is what we should do:

On line 219, where setting the default:
```bash
    LP_RUBY_RVM_PROMPT_OPTIONS=( ${LP_RUBY_RVM_PROMPT_OPTIONS[@]+"${LP_RUBY_RVM_PROMPT_OPTIONS[@]}"} )
    [[ ${#LP_RUBY_RVM_PROMPT_OPTIONS[@]} == 0 ]] && LP_RUBY_RVM_PROMPT_OPTIONS=(i v g s)
```
And then on line 1144, where using it, we shouldn't need unset protection (but do need quotes):
```bash
        __lp_escape "$(rvm-prompt "${LP_RUBY_RVM_PROMPT_OPTIONS[@]}")"
```

liquidprompt
```diff
@@ -1133,6 +1134,31 @@ _lp_python_env_color() {
...
+        __lp_escape "$(rvm-prompt $LP_RUBY_RVM_PROMPT_OPTIONS)"
+        lp_ruby_env=$ret
+    elif command -v rbenv >/dev/null ; then
+        local rbenv_ver=$(rbenv version)
```
Rycieos, line 1147: If doing `local` and assignment on the same line, Zsh 5.0.X will break without quotes (see [Shell Standards Wiki page](https://github.com/nojhan/liquidprompt/wiki/Shell-standards)).
```suggestion
        local rbenv_ver="$(rbenv version)"
```

liquidprompt
```diff
@@ -1133,6 +1134,31 @@ _lp_python_env_color() {
     lp_python_env_color="[${LP_COLOR_VIRTUALENV}${lp_python_env}${NO_COL}]"
 }
 
+_lp_ruby_env() {
+    (( LP_ENABLE_RUBY_VENV )) || return 2
+
+    local ret
+
+    if command -v rvm-prompt >/dev/null ; then
```
Rycieos, line 1142: I forgot before now, sorry. We need a `_lp_require_tool` section for this, to prevent running this function every prompt if the user does not have one of these tools since this feature is enabled by default. Except `_lp_require_tool` can't be used, as there are two valid Ruby providing commands.

Throw this section after line 497:
```bash
    if (( LP_ENABLE_RUBY_VENV )) \
      && ! command -v rvm-prompt \
      && ! command -v rbenv ; then
        LP_ENABLE_RUBY_VENV=0
    fi
````

**Review by Rycieos — CHANGES_REQUESTED** [review]
Some more issues popped up, and a thing I forgot about earlier.

File: liquidprompt (hunk 3 above)
Comment (milouse) [reply liquidprompt:1143]: I’m really not a crack at shell expansion. But is there a difference between:

```shell
# Initializing the array with maybe external content, then testing against length of the array
 LP_RUBY_RVM_PROMPT_OPTIONS=( ${LP_RUBY_RVM_PROMPT_OPTIONS[@]+"${LP_RUBY_RVM_PROMPT_OPTIONS[@]}"} )
[[ ${#LP_RUBY_RVM_PROMPT_OPTIONS[@]} == 0 ]] && LP_RUBY_RVM_PROMPT_OPTIONS=(i v g s)
```
and

```shell
# Testing against the existence of the variable and setting it to the default value if not (as seen on line 301 for LP_COLORMAP)
[[ -z "${LP_RUBY_RVM_PROMPT_OPTIONS-}" ]] && LP_RUBY_RVM_PROMPT_OPTIONS=(i v g s)
```

The second seems more quick/readable than the first for me… and in the case of rvm prompt options we don’t want to merge anything, but just have default value or nothing. What do you think?

liquidprompt
milouse: I just pushed a proposal in 72c01b1afb064814236bf29ca7c8aee8400d1cdb. What do you think of it?

liquidprompt
Rycieos: Yes, there is a difference. The second will evaluate to true if the array has elements and the first element is the empty string. (The reason is that shell is dumb, and referencing an array var without the `[@]` syntax gets the first element of the array.) That is unlikely for this instance, but possible.

It is very likely on the color map instance though, and I'm surprised that no one has reported a bug there. I guess no one is customizing their color map; this is the second fatal bug I have seen on this feature. Good find, I'll be fixing that. You can open a Github issue if you want Github points :smile:

liquidprompt
Rycieos: Yes, that looks good. The only issues are:

1. Delete line 506. Setting no value is pointless if the feature is disabled.
2. Use `[[` instead of `[` everywhere.

liquidprompt
Rycieos: I did some further digging, and it seems there is no way (supported by Bash 3.2+ and Zsh 5.0+) to get the number of elements in an array while being `set -u` safe. That is:
1. `${#var[@]}` does not work (errors when `var` is not set).
2. `${#var[@]-0}` does not work (syntax error).
3. `${var[@]+${#var[@]}}` does not work (returns nothing instead of `0` when the array is unset or empty).

Which leaves our only option to `var=( ${var[@]+"${var[@]}"} )` (setting the array to itself, or if it was unset, setting it to an empty array), because then we *know* it is set, and can then simply do a `[[ ${#var[@]} == 0 ]]` check to see if it is empty.

milouse: Thank you for your explanations, and let’s do this.

milouse:
We should be good now :)

Rycieos:
Yup, that does look good. I don't want to merge 10 commits, could you squash down to 1 or 2 commits? If I do it, Github won't mark the PR as merged.

milouse:
Done

docs/config.rst
```diff
@@ -537,6 +537,29 @@ Features
...
+
+   Display the currently activated RVM_ or RBENV_ virtual environment.
+
+   See also: :attr:`LP_RUBY_RVM_PROMPT_OPTIONS` and
+   :attr:`LP_COLOR_VIRTUALENV`.
+
+   .. _RVM: https://rvm.io/
+   .. _RBENV: https://github.com/rbenv/rbenv
+
+.. attribute:: LP_RUBY_RVM_PROMPT_OPTIONS
```
Rycieos, line 552: This section should be alphabetically sorted. It doesn't need to come directly after the ENABLE section since they will be hyperlinked to each other.

Rycieos (changes requested):
Sorry, obviously didn't look closely enough last time. These should be the last two changes, if you just want to amend the commit.

liquidprompt
```diff
@@ -1133,6 +1151,30 @@ _lp_python_env_color() {
     lp_python_env_color="[${LP_COLOR_VIRTUALENV}${lp_python_env}${NO_COL}]"
 }
 
+_lp_ruby_env() {
+    (( LP_ENABLE_RUBY_VENV )) || return 2
+
+    local ret
+
+    if [[ "$_LP_RUBY_VENV_PROGRAM" = "rvm" ]] ; then
+        __lp_escape "$(rvm-prompt ${LP_RUBY_RVM_PROMPT_OPTIONS[@]})"
```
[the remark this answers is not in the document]
milouse: It seems wrong to me as the whole thing is already quoted with double quotes :/

Can you confirm this fix?

docs/config.rst
milouse: Sorted

liquidprompt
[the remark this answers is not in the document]
Rycieos: Yes, that is correct. Shell is weird, and modern shells can handle quotes inside of quotes. The `"$( )"` section treats the entire command inside like a command not inside of any quotes, so without internal quotes, `rvm-prompt ${LP_RUBY_RVM_PROMPT_OPTIONS[@]}` would undergo word splitting after evaluation, so `LP_RUBY_RVM_PROMPT_OPTIONS=('a long string')` would execute as `rvm-prompt 'a' 'long' 'string'`.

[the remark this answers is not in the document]
milouse: Oh, I didn’t know. thanks again for all your explanations. It should be good now.

Rycieos (approved):
Looks great! This is the kind of patch that I love to merge. Thanks for your work over the years to get this patch merged in.

PR #633, opened 2020-11-20
Ignore errors during sensors

svenstaro:
Due to a [recent CVE](https://nvd.nist.gov/vuln/detail/CVE-2020-12912), Linux now disables some modern sensor interfaces starting with 5.9.8. This in effect makes the output of sensors contain many of these:
```
ERROR: Can't get value of subfeature energy17_input: Kernel interface error
ERROR: Can't get value of subfeature energy18_input: Kernel interface error
ERROR: Can't get value of subfeature energy19_input: Kernel interface error
ERROR: Can't get value of subfeature energy20_input: Kernel interface error
ERROR: Can't get value of subfeature energy21_input: Kernel interface error
ERROR: Can't get value of subfeature energy22_input: Kernel interface error
ERROR: Can't get value of subfeature energy23_input: Kernel interface error
ERROR: Can't get value of subfeature energy24_input: Kernel interface error
ERROR: Can't get value of subfeature energy25_input: Kernel interface error
ERROR: Can't get value of subfeature energy26_input: Kernel interface error
ERROR: Can't get value of subfeature energy27_input: Kernel interface error
ERROR: Can't get value of subfeature energy28_input: Kernel interface error
```
I think the most useful approach to combat this is to use ignore the errors in the output of `sensors`.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1511,7 +1511,7 @@ if (( LP_ENABLE_TEMP )); then
         # Only the integer part is retained
         local -i i
         local IFS=$' \t\n'
-        for i in $(sensors -u |
+        for i in $(sensors -u 2>/dev/null |
                 sed -n 's/^  temp[0-9][0-9]*_input: \([0-9]*\)\..*$/\1/p'); do
             (( $i > ${temperature:-0} )) && temperature=i
         done


Rycieos:
We have had many suggestions exactly like this, but this seems to be special circumstances.

Thanks for the CVE link. I don't see any link to a changelog for lmsensors or Linux detailing the change they made though. (I'm also surprised a change happened already when the CVE is still being investigated). Can you provide links to any changelogs?

Probably the right answer here is to suppress errors normally, but not on startup, and if an error happens, print an extra help message describing how to suppress the error messages inside lmsensors.

Rycieos:
This appears to be the Linux changelog:
https://cdn.kernel.org/pub/linux/kernel/v5.x/ChangeLog-5.9.8

But it only talks about power, not temperature. Could we ask lmsensors for only temp info?

svenstaro:
I don't think it's possible to ask `sensors` only for temperature stuff but maybe I'm missing someting.

Rycieos:

Issue #554, opened 2018-08-20
declare -A doesn't work in the default bash on macOS


augmentedfourth:
https://github.com/nojhan/liquidprompt/blob/6f5af893a3aba29d54005366ba88c3b10c72bcb6/liquidprompt#L226

macOS High Sierra ships with bash 3.2.57, which doesn't understand the `-A` option to `declare`.

I can fix it by setting my default shell to Homebrew bash (4.4.23), but that's a little frustrating.

Rycieos:
I found this hard to believe, but I dug into the change logs: https://tiswww.case.edu/php/chet/bash/CHANGES

```
This document details the changes between this version, bash-4.0-alpha,
and the previous version, bash-3.2-release.
...
ii. The shell provides associative array variables, with the appropriate
    support to create, delete, assign values to, and expand them. (declare -A)
```

We claim to support bash >= 3.2.0:
https://github.com/nojhan/liquidprompt/blob/eda83efe4e0044f880370ed5e92aa7e3fdbef971/liquidprompt#L59-L61

This is important, as Apple won't ship a newer bash than 3.2 for licensing reasons.

However, this change is not in master (yet); it is in commit 3f5af30c257644191fc5a9df08386ed62e853087 on @nojhan's `develop` branch. This is not the officially supported branch. But it brings up a good point.

@nojhan, this change would force us to drop support for bash 3.2, which would force us to drop support for MacOS. Are you planning on merging these develop changes to master?

Rycieos:
Maybe the `devel` version should follow `master` instead? It depends on how @nojhan is planning on doing branches in the future.

On that note, it's been time for a long time for a new official release.

augmentedfourth:
I forked the homebrew packaging discussion to #555 so this thread can get back on topic.

@nojhan, are you planning to deprecate bash 3.2 (and with, it all of macOS, unless people figure out how to manually upgrade bash) by introducing `declare -A` in 1e8a2a19 (and moving it in 3f5af30c)? I hope not. How can we fix this?

augmentedfourth:
Note also that the inclusion of `LP_ENABLE_ERR` in `develop` (added in both ee44113 and e6a4271, oddly, then merged in 8e9bc92) is basically the same change I proposed in 0f4bd8f under #544 (addressing #543). That PR was rejected due to the opinion that @dolmen didn't want to add any new options to `.liquidpromptrc` (and there was an alternate way to fix that particular issue with a custom theme).

[...]

PR #638, opened 2020-12-17
escape IFS character

rotoni:
Shell: bash 3.2.57 (1) -release (x86_64-apple-darwin20)
Operating system: macOS Big Sur
Liquidprompt version: 298693e1e3c091d79aba5d5adf302800766cd8ae

When I updated the liquidprompt, I got an error, so I fixed it.

Error messages:
```
⌁95;charging% [raina_otoni:~] 258 $
-bash: ((: 95;charging: syntax error: invalid arithmetic operator (error token is ";charging")
-bash: ((: 95;charging: syntax error: invalid arithmetic operator (error token is ";charging")
-bash: ((: 95;charging: syntax error: invalid arithmetic operator (error token is ";charging")
```

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1955,7 +1955,7 @@ case "$LP_OS" in
     _lp_battery() {
         (( LP_ENABLE_BATT )) || return 5
         local percent batt_status
-        IFS=; read -r lp_battery batt_status <<<"$(pmset -g batt | sed -n 's/^ -InternalBattery.*[[:space:]]\([0-9]*[0-9]\)%; \([^;]*\).*$/\1;\2/p')"
+        IFS=';' read -r lp_battery batt_status <<<"$(pmset -g batt | sed -n 's/^ -InternalBattery.*[[:space:]]\([0-9]*[0-9]\)%; \([^;]*\).*$/\1;\2/p')"
         case "$batt_status" in
             charged | "")
             return 4


Rycieos:
Oops, my mistake. Thanks for finding this, and for the patch!

In case you care, your commit is with an email address that Github does not recognize, so Github won't give you credit for the commit. You can either add that email address to your Github profile, or fix the email address in the commit and force push the amended commit. I'll wait to merge this until I hear from you.

The reason I did not catch this mistake is that I do not own a Mac. We have an old [pmset simulator](https://github.com/nojhan/liquidprompt/blob/298693e1e3c091d79aba5d5adf302800766cd8ae/pmset-simulator#L9), but it has not been touched in 6 years. I recently have been working on a [test script](https://raw.githubusercontent.com/nojhan/liquidprompt/b699dea7aec3b081292becf52fa1899fe82c3c8b/tools/external-tool-tester.sh) to collect outputs from users' tools like `pmset` so we can build integration tests to catch mistakes like this. Would you mind running it on your machine? Run it like this:
```bash
./external-tool-tester.sh > output.txt
```
If you could run it both while charging and while on battery, it would be even better. You can upload the output file to [Gist](https://gist.github.com/), [Pastebin](https://pastebin.com/), or similar. That would be a huge help to fixing issues on MacOS like these.

augmentedfourth:
@rotoni how are you getting liquidprompt working in the built-in bash 3? I've been running bash 4+ on my Mac for a couple of years: https://github.com/nojhan/liquidprompt/issues/554

Rycieos:
@augmentedfourth do note that we still support Bash 3.2 for that exact reason (that MacOS still ships with that version, although the lastest MacOS has zsh as the default shell). And I can reproduce this bug on Bash 4+ as well as 3.2.

As I commented on the issue you just linked, that bug was only on the `develop` branch, which should not be used. This issue is almost certainly about the current `master` branch.

augmentedfourth:
Yeah, I had a weird configuration that was using an old copy of liquidprompt in one of my environments rather than the one I installed via Homebrew. Very weird, and my own fault. Using the built-in macOS bash I'm seeing the same problem fixed here.

rotoni:
> Oops, my mistake. Thanks for finding this, and for the patch!
> 
> In case you care, your commit is with an email address that Github does not recognize, so Github won't give you credit for the commit. You can either add that email address to your Github profile, or fix the email address in the commit and force push the amended commit. I'll wait to merge this until I hear from you.
> 
> The reason I did not catch this mistake is that I do not own a Mac. We have an old [pmset simulator](https://github.com/nojhan/liquidprompt/blob/298693e1e3c091d79aba5d5adf302800766cd8ae/pmset-simulator#L9), but it has not been touched in 6 years. I recently have been working on a [test script](https://raw.githubusercontent.com/nojhan/liquidprompt/b699dea7aec3b081292becf52fa1899fe82c3c8b/tools/external-tool-tester.sh) to collect outputs from users' tools like `pmset` so we can build integration tests to catch mistakes like this. Would you mind running it on your machine? Run it like this:
> 
> ```shell
> ./external-tool-tester.sh > output.txt
> ```
> 
> If you could run it both while charging and while on battery, it would be even better. You can upload the output file to [Gist](https://gist.github.com/), [Pastebin](https://pastebin.com/), or similar. That would be a huge help to fixing issues on MacOS like these.

Ok, I have uploaded to Gist.
https://gist.github.com/rotoni/62ff2d34a4b662e3ed5bc69d8bada304

rotoni:
> @augmentedfourth do note that we still support Bash 3.2 for that exact reason (that MacOS still ships with that version, although the lastest MacOS has zsh as the default shell). And I can reproduce this bug on Bash 4+ as well as 3.2.
> 
> As I commented on the issue you just linked, that bug was only on the `develop` branch, which should not be used. This issue is almost certainly about the current `master` branch.

Yes, I'm using master branch.

Rycieos:
Thanks again! I added the output you posted as test cases on the new `pmset` tests: 9e205f51db459443e5c2ead0efa76f6a33c47c24

PR #639, opened 2020-12-17
Update liquidprompt-dist for a local rcfile

augmentedfourth:
This may or may not be useful, but wanted to propose the idea. And the implementation is probably sub-optimal.

I use a file sync to keep my `.liquidpromptrc` the same on all my hosts, but on a couple of the individual machines I want to change/add config, so I added this.

is this worth merging/suggesting to users? Or maybe `liquidprompt` itself should source multiple configs? So I could sync `~/.liquidpromptrc` as the "main" config but include local, subsidiary configs from `~/.config/liquidpromptrc`? I know that changes things a lot from how it's documented, though.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -31,6 +31,15 @@ start from this file for your config::
 
     cp ~/liquidprompt/liquidpromptrc-dist ~/.config/liquidpromptrc
 
+In the event that you synchronize your configuration file across multiple
+computers, or if you have an ``/etc/liquidpromptrc`` system-wide from which
+you'd like to make minor deviations in an individual user account, you can
+augment the primary config to add in any local modifications using lines such
+as these::
+
+    LOCAL_RCFILE=$HOME/.liquidpromptrc.local
+    [ -f "$LOCAL_RCFILE" ] && source "$LOCAL_RCFILE"
+
 .. note::
    The example config file does not include every config option, and the
    comments describing the options are less verbose than the descriptions on
--- a/liquidpromptrc-dist
+++ b/liquidpromptrc-dist
@@ -181,4 +181,11 @@ LP_ENABLE_ERROR=1
 # Ex: ("/root" "/home/me/large-remove-svn-repo")
 LP_DISABLED_VCS_PATHS=()
 
+# Use a local liquidpromptrc if it exists.
+# Can be helpful if you sync your primary config across machines, or if
+# there's a system-wide config at /etc/liquidpromptrc from which you'd
+# like to make only minor deviations.
+#LOCAL_RCFILE=$HOME/.liquidpromptrc.local
+#[ -f "$LOCAL_RCFILE" ] && source "$LOCAL_RCFILE"
+
 # vim: set et sts=4 sw=4 tw=120 ft=sh:


augmentedfourth:
If you want to call this a "personalized implementation detail" and kill it on those grounds as "not for everyone," that would make sense too. Just thought it might be worth considering.

augmentedfourth:
I was refactoring my personal rcfile against the distribution one today, which is why I thought of it.

Rycieos:
> Or maybe `liquidprompt` itself should source multiple configs? So I could sync `~/.liquidpromptrc` as the "main" config but include local, subsidiary configs from `~/.config/liquidpromptrc`? I know that changes things a lot from how it's documented, though.

I'm open to the idea. I have thought the same thing before. Figuring out the correct order might be hard.

> on a couple of the individual machines I want to change/add config, so I added this.

I am curious what you configure differently per machines. Since Liquidprompt disables tools when they are not available, I have not found myself needing to do that yet.

I do not hate this patch. I would say it should probably be commented out, like the example at the top of the file is. Or maybe it is more appropriate to add some better description to the [documentation](https://liquidprompt.readthedocs.io/en/latest/config.html) explaining how the config file is simply sourced, and other files could be sourced inside of that file too.

augmentedfourth:
Honestly, one of the main things I initially used this for was to include `LP_ENABLE_BATT=0` on systems without batteries. It seems, on my 3 or 4 most-used LP-enabled hosts, I might not have a need for this anymore. But that's not to say that suggesting the idea might not be useful to someone else.

Rycieos:
Coming back to this as we near the end of the v2.0 merge window.

I like this patch's general idea. I would prefer a note in the documentation (somewhere around config.rst line ) over this change in the example rc file, but I would accept it if it is commented out.

Either way, add more clarity to the description, explaining the reasoning, mainly that it can be used to make the configuration specific for differences between machines:
```bash
# Use a local liquidpromptrc for specific configuration per machine if it exists.
#LOCAL_RCFILE=$HOME/.liquidpromptrc.local
#[ -r "$LOCAL_RCFILE" ] && source "$LOCAL_RCFILE"
```

docs/config.rst
```diff
@@ -31,6 +31,15 @@ start from this file for your config::
...
     cp ~/liquidprompt/liquidpromptrc-dist ~/.config/liquidpromptrc
 
+In the event that you synchronize your configuration file across multiple
+computers, or if you have an ``/etc/liquidpromptrc`` system-wide from which
+you'd like to make minor deviations in an individual user account, you can
+augment the primary config to add in any local modifications using lines such
+as these::
+
+    LOCAL_RCFILE=$HOME/.liquidpromptrc.local
+    [ -f "$LOCAL_RCFILE" ] && source $LOCAL_RCFILE
```
Rycieos, line 41: Quoting vars inside `[[` is not needed. It is needed in `[`. It is also needed when using the variable as an argument. This (and the same thing in the other file) should either be:
```suggestion
    [ -f "$LOCAL_RCFILE" ] && source "$LOCAL_RCFILE"
```
or
```suggestion
    [[ -f $LOCAL_RCFILE ]] && source "$LOCAL_RCFILE"
```
Won't be a problem unless someone has a weird char or space in their rcfile path, but it is possible.

Rycieos (changes requested):
Looks good, just one quoting issue.

Rycieos (approved):
Looks good, thanks!

PR #645, opened 2021-01-11
Feature/vcs staging

### LINKED ISSUE #644: add _lp_git_staged_lines and _lp_git_staged_files for themes
### Description

Hello, i customize a power line theme and i has surprise by the base commend git for _lp_git_uncommitted_lines : is the addition of diff in working and stage.

maybe add _lp_git_staged_lines and _lp_git_staged_files for themes

### How will this be useful?

Can use this for show diff line/file in working ant staging zone in same time

### Example prompt

image

PS : with a little  help i can make a PR

Rycieos:
Adds `git_staged_files()` and `git_staged_lines()`.

Also includes a commit to add caching to the output of the `git diff --shortstat` commands, to speedup calls to both `*_lines()` and the related `*_files()` git calls on the same prompt build. I'm not completely sold on it yet, but it does work without speed reductions.

Fixes #644

### DIFF
--- a/docs/functions/data/vcs.rst
+++ b/docs/functions/data/vcs.rst
@@ -129,6 +129,26 @@ this.
 
    .. versionadded:: 2.0
 
+.. function:: _lp_vcs_staged_files() -> var:lp_vcs_staged_files
+
+   Returns ``true`` if any staged files exist in the repository. In other words,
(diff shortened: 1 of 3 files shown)


Orgams:
I have no remarks, it's ok for me ^^

PR #650, opened 2021-01-24
Make load display actual load instead of adjusted

### LINKED ISSUE #530: CPU usage can't be computed from load
Shell: bash/zsh
Operating system: Linux
Liquid Prompt version (tag, commit): master 9c80396

/proc/loadavg contain load but liquidprompt show CPU percentage. There is not a way to transform load into the percentage. The load is traditionally defined in queue theory as an average number of processes waiting for execution. Imagine this hypothetical example. You have 100 processes scheduled at the start of each second and each process takes 10ms to execute and one CPU to run it. Then your load would be 50 but your CPU utilization would be 100%. Then you take one process running non-stop. Your CPU utilization would be 100% but your load would be 0 or 1 depending if you include running process into load calculation which Linux does. But I'm not sure. Anyway, the load could be way over 1 on single CPU system and CPU utilization could never be over 100%. Unfortunately, this is exactly what I see using liquidpromp, CPU over 100% shown in white. It is not right. Correct way how to compute CPU utilization is to read the first line of /proc/stat

Something like
```
$ read _ user nice system idle iowait irq softirq steal guest guest_nice eol < /proc/stat
$ old_cpu_sum=$(( user + nice + system + idle + iowait + irq + softirq + steal + guest + guest_nice ))
$ old_cpu_idle=$idle
# after some while, command, next prompt rendering, ...
$ read _ user nice system idle iowait irq softirq steal guest guest_nice eol < /proc/stat
$ new_cpu_sum=$(( user + nice + system + idle + iowait + irq + softirq + steal + guest + guest_nice ))
$ echo $(( 100 - (old_cpu_idle - idle) * 100 / (old_cpu_sum-new_cpu_sum) ))%
29%
```
And now this is right CPU usage value.

Rycieos:
The load average display has been notoriously not helpful. It is scaled by the CPU count, and then multiplied by 100 to make it apear like a percentage. But load average and CPU utilization have almost nothing to do with each other.

This change displays the actual load average, while keeping the threshold and color calculations the same. This means the functionality is no different, just that the display has changed formats.

To make the configuration options LP_LOAD_THRESHOLD less confusing, it too has been changed to be an actual load value, no longer "centiload". Old values are still supported, which is done by checking for a period ('.') in the string. So for this historical reason, all new values must have a period in them. Which should not be a problem, any useful value would be a float.

Change the greater-than comparison to a greater-than-or-equal-to, since it is a little more intuitive to what would be expected from the config option.

Since I was at it, I added a new config option LP_LOAD_CAP, to configure the range for color values output by _lp_color_map() for load averages. It too takes a load average float value. As it does not have the historical reason above, it does not accept "centiload" values.

The implementation details are long. Floating point math does not exist in shells (which I suspect is the reason that "centiload" was used in the first place). To make it possible to compare, the values are scaled up by 100 just like before. They are not shown to the user. 100 scaling should be plenty for the precision we need.

To do the scaling, __lp_floating_scale() is added, which as a standalone function, can be unit tested.

And for good measure, the lp_load return value is passed through __lp_escape(), just in case.

Resolves #530

Special thanks to Tore Anderson (@toreanderson) for helping dig through the issues here.


Rycieos:
Links to live documentation changes:
https://liquidprompt--650.org.readthedocs.build/en/650/config.html#LP_LOAD_CAP
https://liquidprompt--650.org.readthedocs.build/en/650/config.html#LP_RUNTIME_THRESHOLD
https://liquidprompt--650.org.readthedocs.build/en/650/functions/data.html#lp_load
https://liquidprompt--650.org.readthedocs.build/en/650/functions/internal.html#lp_floating_scale

toreanderson:
LGTM, except for one thing I already pointed out in #530: the default `LP_LOAD_CAP` value of `1` is too low.

It ends up colouring load averages where the machine still is idle 10% of the time with the highest «danger red» colour. All load averages above that 10% idle point is shown with the exact same «danger red» colour, so the colour communicates exactly zero information as to how (over)loaded the machine really is. Screenshot below on a 4-CPU machine. A load average of 2.40 is nothing to worry about for such a machine as it is idle 40% of the time, yet it is shown with in red: 

bilde

A default value of `LP_LOAD_CAP=2` yields a more useful colour palette, at this point the first red colour shows up exactly when the machine reaches 0% idle:

bilde

`LP_LOAD_CAP=3` would actually be quite good too, as it even shows a green value and saves the most "scary" reds for the really high load situations:

bilde

Rycieos:
> the default `LP_LOAD_CAP` value of `1` is too low.

I left it as that, since that was the previous value. I am now convinced that something higher would be a better default.

Normally I would say that we shouldn't change something that was one way before, as the user won't realize why it changed. But since users can now configure it, why not make the default slightly more sane. I guess I just want something that "makes sense", but since load average has no upper bound, it is hard to point to any specific number and say "makes sense". I guess I like `2` more than `3`, since (to me at least) if the number is high enough to add to my prompt, it is "scary" enough to make it not green.

Also I just read the [Linux man page](https://man7.org/linux/man-pages/man5/procfs.5.html), and it claims the first number is a 1 minute average. I guess I need to fix the documentation. I need to verify if that is consistent across all Unixes, not just Linux.

toreanderson:
> if the number is high enough to add to my prompt, it is "scary" enough to make it not green.

In that case, you might want to consider increasing the default `LP_LOAD_THRESHOLD` value as well (to something like 0.8 or 0.9), because a load average of 0.6 is nowhere near "scary" (after all, it means «idle 40% of the time»).

Rycieos:
> a load average of 0.6 (per CPU) is nowhere near "scary" (after all, it means «idle 40% of the time»).

But that's exactly the point, it does not mean that. All that load average means is how many processes are waiting for something. On most Unixes, that means waiting for CPU. On Linux, it also means waiting for I/O.

Theoretically, a machine could have 100% CPU utilization while having 0.0 load average. Or the reverse, it could have crazy high load average while having virtually 0% CPU utilization.

In reality, they often are usually somewhat related, but no correlations can accurately be drawn.

All that said, I don't want to change the default anyway; without a very good reason to, I don't want to change how the default prompt works.

Issue #646, opened 2021-01-12
Question: Why does 'master' get tags for WIP versions?


atifraza:
I used to clone the repository instead of relying on distribution packages, because this provided the most up-to-date stable version of liquidprompt.

Recently, I was setting up a new system with the same script, and instead of a stable release, I ended up with a beta release when cloning the *master* branch.

My question is, why not develop and test the new versions on a *dev* branch?
I hope I did not miss something in the previous discussions regarding this aspect of development work.

Rycieos:
This is a valid question and discussion point.

For a brief history: I took over as the lead maintainer of the project in May of 2020. Before May of 2016, this project did have a `develop` branch that was used as expected. You can see the removal of this standard in this commit: 9184ce7ab5598b68c1e793bfb44f2583bf9c3ea0. On removal, there was no clear answer as to if `master` is "stable" as well as being the development branch.

My understanding of general Git branching standards is this: `master` (or the new standard of `main`) is simply the default branch, and carries no further meaning like "stable", "development", or "latest". If you see a project with a branch named `master` or `main`, I think one should not make assumptions on what purpose that branch has.

That said, obviously Git has a concept of a default branch (aka repo HEAD). This is the branch that gets checked out by default when cloning.

Slightly related, I was planning on using the future Github tool to rename `master` to `main` when they launch the tool to make that painless.

Lastly, Liquidprompt has a "no breaking changes" policy, meaning anything merged should not break any existing use cases. So as for the for the [134 commits](https://github.com/nojhan/liquidprompt/compare/v1.12.1...master) on `master` since `v1.12.1`, non of those commits have *planned* breakage. That said, yes, I have marked "beta" tags on those commits. I expect there to be small bugs as development goes.

All that to say: yes, I agree. This project should (probably) have both a "stable" branch and a "development" branch. There are two related questions along with that:
1. Which should be default? (the branch that is checked out when cloning the project.)
2. Which should have the name `master` (or `main` in the future)?

The first question is not simple. Most users have come to expect that the default branch is tracking development. Also, if we change the default branch, that does not change the currently checked out branch for any user that has already cloned the repo. But I agree that tracking development is probably not the best idea.

The second question is easier. Probably the default branch should keep the default name.

This is a user focused topic, so I want to hear feedback on this.

atifraza:
Thanks for the prompt response.

[...]

PR #654, opened 2021-01-31
Zinit setup documentation

johnhamelink:
This change adds instructions on how to load liquidprompt using `zinit
load`, then provides an additional config which makes use of zinit's "ice modifiers" to:

 - ~Only load the prompt if the shell is marked as interactive (The 'i' in `$-`): `load'[[ $- = *i* ]]'`~
 - ~Reset the prompt after loading the snippet: `wait'!'`~
 - Hide the default loading message: `lucid`
 - Don't change the directory when loading the plugin: `nocd`

After some testing, it became clear that the most idomatic configuration was probably not the best default for most users, and so it was simplified.

It also makes use of `zinit light` instead of `zinit load`, which is faster because it does not attempt to monitor or profile the execution of the plugin.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -35,6 +35,15 @@ following line in your ``.zshrc`` after activating Antigen::
 
    antigen bundle nojhan/liquidprompt
 
+Installation via Zinit
+----------------------
+
+To install via `Zinit <https://github.com/zdharma/zinit>`_, simply add the
+following lines in your ``.zshrc`` after activating Zinit::
+
+    zinit ice ver"stable" lucid nocd
+    zinit light nojhan/liquidprompt
+
 Dependencies
 ============
 


docs/install.rst
```diff
@@ -35,6 +35,22 @@ following line in your ``.zshrc`` after activating Antigen::
 
    antigen bundle nojhan/liquidprompt
 
+Installation via Zinit
+----------------------
+
+To install via `Zinit <https://github.com/zdharma/zinit>` _, simply add the following line in your ``.zshrc`` after activating Zinit::
```
Rycieos, line 41: There should be no space between the '`' and '_', or the link does not work. You can see this in the compiled docs, linked in the build jobs.

Rycieos, line 41: Format these lines to max of 80 characters.

docs/install.rst
```diff
@@ -35,6 +35,22 @@ following line in your ``.zshrc`` after activating Antigen::
...
+
+   zinit load nojhan/liquidprompt
+
+Zinit provides an additional `Ice Modifier <http://zdharma.org/zinit/wiki/INTRODUCTION/#some_ice-modifiers>` _ syntax, which can be used to improve the startup time of your terminal::
+
+   zinit ice load'[[ $- = *i* ]]' \
+             wait'!' \
+             lucid \
+             nocd \
+             atload'!__lp_set_prompt'
```
Rycieos, line 51: I haven't tested this. I'm assuming that this does work? I see two possibilities:
1. The `light` command makes the file load, but not execute. Hence the atload part, telling it what command to run in the prompt.
2. The atload part is duplicating definition, so _lp_set_prompt is run twice. Liquidprompt does after all setup Zsh all by itself correctly when not used in a plug-in manager.

I'm guessing it is the latter. I'll need to do a bit of testing to dissuade my worries. I don't want to add something to the docs that isn't completely correct. I'll try to do some testing later tonight.

Rycieos (changes requested):
Looks good! Thanks for reworking this PR. Just two small issues, and a question.

docs/install.rst
johnhamelink: > The light command makes the file load, but not execute. Hence the atload part, telling it what command to run in the prompt.

`light` doesn't change whether the file loads or not, it just changes what `zinit` does while it's loading, that is to say: it doesn't provide information in the `zinit report` command, which is useful for profiling.  [You can read about that here.](https://zdharma.org/zinit/wiki/INTRODUCTION/#basic_plugin_loading).

> The atload part is duplicating definition, so _lp_set_prompt is run twice. Liquidprompt does after all setup Zsh all by itself correctly when not used in a plug-in manager.

The `ice` part of the snippet is telling `zinit` to defer execution, and when it's ready to execute, the `__lp_set_prompt` function is evaluated, so I don't believe it is run twice. The advantage of doing things this was is that the terminal becomes usable immediately - even before liquidprompt is ready.

Rycieos: The issue is that when the Liquidprompt file is sourced, it configures Zsh to run that function. So if you also tell zinit to do that also, will it run twice?

johnhamelink: @Rycieos I'm not sure how definitive this is, but I added an echo to the first line of `__lp_set_prompt` and then started a new terminal, and the echo only appears once. Is that helpful?

docs/install.rst
```diff
@@ -35,6 +35,25 @@ following line in your ``.zshrc`` after activating Antigen::
 
    antigen bundle nojhan/liquidprompt
 
+Installation via Zinit
+----------------------
+
+To install via `Zinit <https://github.com/zdharma/zinit>`_, simply add the
+following line in your ``.zshrc`` after activating Zinit::
+
+   zinit load nojhan/liquidprompt
```
Rycieos, line 44: I just remembered, mentioned in https://github.com/nojhan/liquidprompt/issues/646#issuecomment-765661569, we have a `stable` branch, tracking stable releases (since `master` is the development branch). The docs were updated to recommend users install the stable branch, so I would like this too as well.
```suggestion
   zinit load ver"stable" nojhan/liquidprompt
   
Or, if you want to use the development (non-stable) branch::

   zinit load nojhan/liquidprompt
```
It seems to be that simple with Zinit?

Rycieos (changes requested):

docs/install.rst
johnhamelink: Hi again @Rycieos,

> It is not faster. Doing only zinit light nojhan/liquidprompt is the same speed, the lazy loading doesn't help (at least on my machine).

The performance improvement comes as a result of its asynchronous loading - if you have liquidprompt in combination with other plugins which all would otherwise load at startup, the cumulative effect of using zinit provides a noticeable performance improvement (anecdotally!)

> Because it is lazy, I can see it flash for a second as the default prompt displays then is replaced.

Yes, this is absolutely a valid point

> It is way more configuration. Why do we need to recommend all that when the simple command works just as well?

My intention was to produce the most idiomatically correct implementation of the config (it's the one I use myself) to make the most of zinit's featureset. If someone truly wants to use zinit in the way I am, I guess we've probably produced enough of a papertrail with this PR that they'll find what they're looking for - even if it's not in the docs, so on that level I don't really mind much :)

> Loading this way, Liquidprompt displays an exit code 1 on the first prompt. With only a load or light command, it does not.

Huh, that *is* a good point - and not something I noticed myself. I wonder why...

Yep, I think you make a salient case for dropping it, and I agree that the last one in particular is a showstopper. I'll remove it in my next commit!

Rycieos: Belay that order!

The `ver` command does not work as I described it. I just tested it, it seems that it needs to be part of an `ice` command? So unless you can figure out the correct way to use it as part of a `load` or `light`, I think we need an `ice` to target the stable branch.

johnhamelink: Aye aye cap'n :smile: 

Yes you are right, it was my intention to drop the lazy/async aspects of the ice modifier stuff only!

Rycieos: Just realized I forgot the other reason not to do that: Telling users to reference `__lp_set_prompt()` is not a good idea; it's a private function. While there are no plans to change that specific part of the internal API, it is a part that is not guaranteed to not change between versions, and I don't want users depending on that.

Obviously I can't stop power users like yourself from making it work to get lazy loading working, but at least I can prevent the average user from doing that.

Though maybe this should be a place for enhancement. It would be nice for Liquidprompt to support lazy/async loading, at least for Zsh. I would probably need some help though.

docs/install.rst
```diff
@@ -35,6 +35,15 @@ following line in your ``.zshrc`` after activating Antigen::
 
    antigen bundle nojhan/liquidprompt
 
+Installation via Zinit
+----------------------
+
+To install via `Zinit <https://github.com/zdharma/zinit>`_, simply add the
+following line in your ``.zshrc`` after activating Zinit::
```
Rycieos, line 42: Two lines now:
```suggestion
following lines in your ``.zshrc`` after activating Zinit::
```

Rycieos (changes requested):
This looks good. Sorry it took so long to settle on something.

All I need now is a squash, no need for 6 commits. I can do it myself, but then Github doesn't realize that I merged your commits, so it marks the PR as "closed" instead of "merged". If you don't care about that, let me know and I'll do the squashing.

johnhamelink:
@Rycieos no problem, I'll squash & force-push :) It's all good RE rigour - there's a reason why liquidprompt has been rock solid the entire time I've used it!

Rycieos (approved):
Looks good! Thanks for being flexible.

Issue #651, opened 2021-01-26
Replace all shell escape sequences


Rycieos:
This is a catch-all issue, documenting the changes that need to happen before these dependent issues can be fixed:

* Add running command to terminal title: #609
* Support Ksh: #608 
* Add a newline if the prompt is too long: #599 
* FQDN hostname is not really FQ: #472
* Allow setting date and time format: #281

To make this work, all escape sequences must be replaced with real strings. The escape sequences in Liquidprompt are:

* `\u`/`%n`: username, in `_lp_username()`
* `\h`/`%m`: hostname, in `_lp_hostname()`
* `\H`/`%M`: FQDN, in `_lp_hostname()`
* `\t`/`%*`: time, in `_lp_time()`
* `\$`/`%(!.#.%%)`: prompt mark, in `_lp_smart_mark()`

Working directory escape sequences have already been replaced as part of v2.0.

PR #663, opened 2021-03-26
bash: don't touch the promptvars option

cmm:
Rationale:

* In general, it is not proper for liquidprompt to change shell
  options that it does not absolutely have to change.

* LP_PS1_{PREFIX,POSTFIX} are a thing, and are in fact an effective
  means to combine liquidprompt with other prompt-touching logics
  (such as emacs-vterm).  So it is not good to prevent those other
  logics from actually working.

This commit restores the bash version of __lp_escape to what it did
prior to 62f0270 (escaping dollar signs in addition to backslashes)
and removes all interference with the promptvars shell option.

NOTE: I don't know much about shells other than bash, so left the
parallel logic for zsh as is.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -54,7 +54,8 @@ if test -n "${BASH_VERSION-}"; then
     # Must be used for all strings injected in PS1 that may comes from remote sources,
     # like $PWD, VCS branch names...
     __lp_escape() {
-        ret="${1//\\/\\\\}"
+        local arg="${1//\\/\\\\}"
+        ret="${arg//\$/\\\$}"
     }
 elif test -n "${ZSH_VERSION-}" ; then
     # Check for recent enough version of zsh.
@@ -2884,7 +2885,6 @@ prompt_on() {
         LP_OLD_PS1="$PS1"
         if (( _LP_SHELL_bash )); then
             LP_OLD_PROMPT_COMMAND="${PROMPT_COMMAND-}"
-            _LP_OLD_SHOPT="$(shopt -p promptvars)"
         else # zsh
             LP_OLD_PROMPT_COMMAND=""
             _LP_ZSH_PROMPT_THEME=""
@@ -2916,8 +2916,6 @@ prompt_on() {
             declare -g +x PROMPT_COMMAND
         fi
 
-        # Disable parameter/command expansion from PS1
-        shopt -u promptvars
         PROMPT_COMMAND=__lp_set_prompt
         (( LP_DEBUG_TIME )) && PROMPT_COMMAND="time $PROMPT_COMMAND" || true
     else # zsh
@@ -2943,7 +2941,6 @@ prompt_on() {
 prompt_off() {
     PS1=$LP_OLD_PS1
     if (( _LP_SHELL_bash )); then
-        eval "$_LP_OLD_SHOPT"
         PROMPT_COMMAND="$LP_OLD_PROMPT_COMMAND"
     else # zsh
         add-zsh-hook -d precmd "$_LP_ZSH_HOOK"
@@ -2956,7 +2953,6 @@ prompt_off() {
 prompt_OFF() {
     PS1="$_LP_MARK_SYMBOL "
     if (( _LP_SHELL_bash )); then
-        shopt -u promptvars
         PROMPT_COMMAND="$LP_OLD_PROMPT_COMMAND"
     else # zsh
         add-zsh-hook -d precmd "$_LP_ZSH_HOOK"


Rycieos:
This is actually coming at a good time.

Due to the work I'm doing for #651, all prompt escapes are being removed anyway. See my WIP branch [title-command](https://github.com/nojhan/liquidprompt/tree/title-command).

But that branch, by itself and also combined with this patch, has a problem: the extra backslashes show up in the title, as the title is changed to no longer be printed as part of PS1. While normally rare, combined with this patch, every title would show `\$`. That is unacceptable IMO, so we need a fix. Any ideas? The only idea I have is to pass around two strings from each data source: one for the prompt, and one for the title. Gross.

Note that Github doesn't recognize your commit email, so it isn't attributing the commit to you. You might want to fix that.

cmm:
Well, I don't think that turning data source outputs from just strings into {raw, ps1-escaped} pairs across the codebase is especially gross.  It does look like more work, I'll grant you that -- shell programming is *awkward* -- but there's nothing wrong with the approach (and plenty right, in fact).

(Thank you for alerting me to the e-mail problem, should be rectified now)

Rycieos:
> 
> 
> Well, I don't think that turning data source outputs from just strings into {raw, ps1-escaped} pairs across the codebase is especially gross. It does look like more work, I'll grant you that -- shell programming is _awkward_ -- but there's nothing wrong with the approach (and plenty right, in fact).

Well thanks for the reinforcement! I'll get working on a proof-of-concept for that then, see if any other weird things come up. If it looks good, then there should be no obstacles to getting this merged.

cmm:
now that I've taken a look at the code, this seems even simpler: currently the escaping is done too early, and is thus baked into the data.  it does not have to be this way --  escaping could be postponed until the consumption of those data in the FOO_color() functions.

you could even postpone it further: compose the PS1 not as a flat string at first, but as 2 arrays, one of strings and the other of indices into the first one which are the elements that need escaping.  then collect the first array into the final PS1, escaping whatever needs it along the way.

Rycieos:
> this seems even simpler: currently the escaping is done too early, and is thus baked into the data.  it does not have to be this way --  escaping could be postponed until the consumption of those data in the FOO_color() functions.


That was my thought, yes. I had considered that while back, but rejected the idea because it meant theme authors would have more work and understanding to do. But it is probably the right choice.


> you could even postpone it further: compose the PS1 not as a flat string at first, but as 2 arrays, one of strings and the other of indices into the first one which are the elements that need escaping.  then collect the first array into the final PS1, escaping whatever needs it along the way.

I don't understand. Could you explain or show what you mean?

I won't have a chance to work on this today, so if you are impatient you could try and get something working off my title-command branch.

cmm:
that's going to be bad news for theme authors, probably, yes.

> if you are impatient you could try and get something working off my title-command branch.

I'm actually not in any hurry (my setup is patched up enough to work with 2.0.2), but I'll see what I can do anyway :)

Rycieos:
I made a WIP commit on [`wip/raw-data-string`](https://github.com/nojhan/liquidprompt/tree/wip/raw-data-string): 929851bdda4f0361e6539f4a9d2d6084bbca8c1b. It has some things I don't like about it. Or rather, a lot of things. I really don't like it at all right this moment, but maybe it will grow on me.

But it does work. If you wouldn't mind taking a look at it and letting me know what you think.

cmm:
I've made an attempt to understand the code and mostly just got lost :/.  Doesn't look like I'll be much help to you here, but maybe the random questions and observations that have arisen during the attempt could be of some nebulous help (as in, there is not much point in you spending time and effort _answering_ them, but perhaps their nature could be of help in understanding the _sort_ of things people might find baffling about liquidprompt.  I mean, unless I'm unique, which I doubt):

- Default theme seems to be split between `liquidprompt` and `liquid.theme`, what's the reason for that?
- Default theme simply copies the prompt (stripped of formatting) into the title string (ISTR there's an issue already about that?), which does not look very useful.  And the title is set using that string in `__lp_set_prompt()` (which is theme-independent.  or is it?), but the powerline theme overwrites the title string with a minimalistic version in the `_lp_powerline_theme_directory()` function (of all places).  Or does it overwrite anything?  What happens before what?
- There is no naming convention to tell apart functions that build up data from those that perform observable effects.  I found this extremely confusing.
- The "command in title" feature (one of those cool features that you don't know you _want_ until you see it!) does not seem to be implemented in either included theme, or I guess I just failed to activate it?
- Are there actual legacy themes in the wild that need to be supported indefinitely?  Cannot their users/authors be told to either adapt the themes or refrain from upgrading liquidprompt (I mean, people who can create a theme for this thing don't exactly seem like they need to be coddled)?  It's not like anyone is forcing anyone to upgrade?  And if I have one takeaway from all this that I'm pretty confident about, it would be that there's more than enough deprecation code and docs already, and it's not like the code base is too simple for what it does...

Sorry about being useless here and thanks for your time!

Rycieos:
> 
> * Default theme seems to be split between `liquidprompt` and `liquid.theme`, what's the reason for that?

`liquid.theme` is a relic; it serves as an example of how to configure Liquidprompt and how to use substitute characters on terminals that do not support UTF-8 (which is very rare in 2021). It could probably do with an updated note at the top saying as such.

>   * Default theme simply copies the prompt (stripped of formatting) into the title string (ISTR there's an issue already about that?), which does not look very useful.  And the title is set using that string in `__lp_set_prompt()` (which is theme-independent.  or is it?), but the powerline theme overwrites the title string with a minimalistic version in the `_lp_powerline_theme_directory()` function (of all places).  Or does it overwrite anything?  What happens before what?

Yes, it is not the most useful thing. But 1) change is "bad" and 2) no one has suggested a better solution.

[`__lp_set_prompt()`](https://liquidprompt.readthedocs.io/en/stable/functions/internal.html#lp_set_prompt) is theme-independent, yes. The [function documentation](https://liquidprompt.readthedocs.io/en/stable/functions.html) I think does a good job at defining which functions are data, default theme, and otherwise.

The powerline theme(s) define their own title, yes. They do it minimally because that theme is newer, and I made that choice. Those themes are as much an example as they are usable, so any other theme author can decide what they want to put in the title.

There is no "title overwriting", because each theme must set a title, or it defaults to just the name of the shell. Or a user can override all of that with [`lp_title()`](https://liquidprompt.readthedocs.io/en/stable/functions/public.html#lp_title).
 
>   * There is no naming convention to tell apart functions that build up data from those that perform observable effects.  I found this extremely confusing.

I refer again to the function documentation. But yes, the function names are not obvious. This project is old, and while I have recently attempted to fix its large issues, I have also tried to break as little backwards compatibility as possible, which means not renaming everything.

That said, if you find the documentation lacking, please open an issue; I would love to clear it up.

>   * The "command in title" feature (one of those cool features that you don't know you _want_ until you see it!) does not seem to be implemented in either included theme, or I guess I just failed to activate it?

That part is theme agnostic while the contents of the title are not. Probably why it doesn't show up is that you haven't enabled [`LP_ENABLE_TITLE`](https://liquidprompt.readthedocs.io/en/stable/config.html#LP_ENABLE_TITLE), which is disabled by default. Or maybe you just aren't using the right branch? It is only on the `title-command` and `wip/raw-data-string` branches right now.

>   * Are there actual legacy themes in the wild that need to be supported indefinitely?  Cannot their users/authors be told to either adapt the themes or refrain from upgrading liquidprompt (I mean, people who can create a theme for this thing don't exactly seem like they need to be coddled)?  It's not like anyone is forcing anyone to upgrade?  And if I have one takeaway from all this that I'm pretty confident about, it would be that there's more than enough deprecation code and docs already, and it's not like the code base is too simple for what it does...

Legacy themes, not widespread, no. For individual users, yes. I had considered (and was testing a POC) doing a total breaking change for Liquidprompt v2 that would kill the old "themes" from v1 (that I now refer to as "default theme templates"). But dolmen convinced me that we have too many users that would just leave the project if an update broke their setup instead of fixing their configs. So v2 now supports the old templates as well as the newer themes (that are much more powerful). The other option was the create Liquidprompt v2 as a separate project, but that would split the userbase, which I do not want to do.
 
> Sorry about being useless here and thanks for your time!

No problem at all. I'm always happy to answer questions, because I love seeing people be interested in the project.

The 8 commits on the [`wip/raw-data-string`](https://github.com/nojhan/liquidprompt/tree/wip/raw-data-string) are really blocking me here, I want to be happy with any patch before I merge it, and I'm just not happy yet. That's why I haven't committed anything new in the past few weeks...

cmm:
I think I figured out what was gnawing at me about all this (I'm probably still missing something basic that makes this invalid, but anyway): there is no need to keep raw values of every data source.

This is because titles and prompts are different critters and have different purposes: a prompt (from liquidprompt's point of view, anyway -- but I think it's correct or I wouldn't be here) is a sort of up-to-the-second mini-status of the host; a window title, on the other paw, is only a way to quickly tell "what's in this window/tab" (which is why putting load/temperature/battery/time in the title makes zero sense, unless you arrange for all that to update on its own like the hardstatus in screen does, and that's obviously _way_ out of liquidprompt scope).

And as long as you keep title content minimal (either non-configurable at all, as now, or configurable _separately_ from prompt content), you can restrict it to (at most) hostname/user/path/command and deal with these 4 things ad-hoc (i.e. basically just arrange to keep unescaped path and command around -- the rest are fine as is, and there's no need for a hash/dollar/VC mark either).  I.e. the present default theme approach of replicating the whole prompt into title (and having to contort the code trying to deal with the formatting problems due to that approach) is just wrong and only complicates things.

Rycieos:
You are probably right. I'll need to think about it more, but it does make sense.

I think the reason that the whole prompt was dumped as the title (back in v1.2, 11473777cabeee763302e2ef9a4602613156ae90 in 2013) was because they wanted to avoid passing around extra strings to save as a title. It was the simple solution. But now that there are so many more data sources, passing around two copies of each string is way more complicated. Now it makes sense to strip the title down to essentials.

However, I would still like to leave it possible to configure. Especially since I added that feature in v2.0, and removing it now would be a breaking change. I won't pass around tons of extra data for sources that the default theme doesn't put in the title, but other themes will still have the option to rearrange and manipulate data before it goes into the title (which is the whole point of themes).

I'll think on this a bit more, but I do think you are right.

Rycieos:
I worked on a patch based on what you described, branch [title-escape-fix](https://github.com/nojhan/liquidprompt/tree/title-escape-fix) (a23af79232fc3b716dc54bc3927b3e05bd777189).

I'm pretty happy with it. It's a small change, but it's big enough to fix the biggest issues. The big change is that the default theme's title is no longer the entire prompt, but only `[user@host:path] $`. I had made the Powerline themes use that as well, so even better to get some consistency. Thanks for the feedback to get me on the right track.

The 8 commits on the [title-escape-fix](https://github.com/nojhan/liquidprompt/tree/title-escape-fix) branch I'm very happy with, and would like to merge into master as part of v2.1. I opened PR #665, if you would like to review and test it. It needs to be merged before this PR can be merged.

Rycieos:
> NOTE: I don't know much about shells other than bash, so left the parallel logic for zsh as is.

I dug into it for Zsh, and found this [in the docs](https://linux.die.net/man/1/zshmisc) (section *Expansion Of Prompt Sequences*):
* If the *PROMPT_BANG* option is set, a '!' in the prompt is replaced by the current history event number. A literal '!' may then be represented as '!!'. 
* If the *PROMPT_PERCENT* option is set, certain escape sequences that start with '%' are expanded. Many escapes are followed by a single character, although some of these take an optional integer argument that should appear between the '%' and the next character of the sequence. More complicated escape sequences are available to provide conditional expansion. A '%' may be represented as '%%'.
* If the *PROMPT_SUBST* option is set, the prompt string is first subjected to parameter expansion, command substitution and arithmetic expansion. (Can be worked around by escaping '$' as '\\$', and '\`' as '\\\`').
 
The issue is that when those options are turned off, the escaped version is printed litterally. Which means we only need and can do the substitution when the option is currently on. Possible, but annoying. Don't worry about implementing this for Zsh as part of this PR, I can do that.

Bash does not do that (from my testing); a '$var' is printed the same as a '\\$var' when `shopt -u promptvars`. *However*, it does not work when using '\`' escapes:

```bash
$ shopt -s promptvars
$ func() { echo foo; }
$ PS1='`func`> '
foo> PS1='\`func\`> '
`func`> shopt -u promptvars
\`func\`> 
```

Which means two things for this PR:
1. While it works correctly for escaping '$' characters, it does not for '\`'.
2. Since you can't always escape '\`', it looks like you will need to detect what the status of the *promptvars* option is, and only escape '\`' if it is on.

liquidprompt
```diff
@@ -54,7 +54,8 @@ if test -n "${BASH_VERSION-}"; then
     # Must be used for all strings injected in PS1 that may comes from remote sources,
     # like $PWD, VCS branch names...
     __lp_escape() {
-        ret="${1//\\/\\\\}"
+        local arg="${1//\\/\\\\}"
+        ret="${arg//\$/\\\$}"
```
Rycieos, line 58: Using a check like this (see [the docs](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html)) works perfect in my testing. And doing the check every time is actually faster than storing a check in a variable and checking the variable later (who knew?)
```bash
$ func() { for i in {1..1000000}; do if (( LP_ENABLE_SHORTEN_PATH )); then :; fi; done; }
$ time func
real   	0m4.375s
user   	0m4.276s
sys    	0m0.082s

$ func() { for i in {1..1000000}; do if shopt -q promptvars; then :; fi; done; }
$ time func
real   	0m3.924s
user   	0m3.877s
sys    	0m0.034s
```
```suggestion
        ret="${1//\\/\\\\}"
        if shopt -q promptvars ; then
            ret="${ret//\$/\\\$}"
            ret="${ret//\`/\\\`}"
        fi
```
This change (amended or squashed to one commit) plus a rebase to latest master seems like all the changes needed to get an approval from me.

Rycieos (changes requested):

Rycieos:
I'm merging this as is, because we need this feature for #672. The fixes I suggested above I implemented in a1d0a54, and I also made some similar changes for Zsh in 58969b2.

Thanks for the PR!

PR #666, opened 2021-04-14
Fix `_lp_git_bookmark()` cmd not found msgs (brewed bash, macOS x86)

zbeekman:
On macOS (10.15.7, x86) with brewed Bash (5.1.4(1)-release) and certain
Bash options set, command not found error messages are printed when
`_lp_vcs_bookmark()` is invoked as the TEST-COMMAND of an if statement.
This is rather annoying. The output is printed on stdout. This commit
fixes the issue by discarding any output created by the if TEST-COMMAND.

Here are my Bash settings for completeness & reproducibility:
```
$ set -o
allexport       off
braceexpand     on
emacs           on
errexit         off
errtrace        on
functrace       off
hashall         on
histexpand      on
history         on
ignoreeof       off
interactive-comments    on
keyword         off
monitor         on
noclobber       off
noexec          off
noglob          off
nolog           off
notify          off
nounset         off
onecmd          off
physical        off
pipefail        off
posix           off
privileged      off
verbose         off
vi              off
xtrace          off
```

Screenshot of the issue on master branch (as of 145f146e9cb4fb3c30a22e6c35529120f4650a28):

image

Screenshot showing issue is now resolved in this PR:

image

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1522,10 +1522,10 @@ _lp_vcs_details_color() {
     if _lp_vcs_branch; then
         branch="$lp_vcs_branch"
 
-        if _lp_vcs_bookmark; then
+        if _lp_vcs_bookmark > /dev/null; then
             branch+=": $lp_vcs_bookmark"
         fi
-    elif _lp_vcs_bookmark; then
+    elif _lp_vcs_bookmark > /dev/null; then
         branch="$lp_vcs_bookmark"
     elif _lp_vcs_tag; then
         branch="tag: $lp_vcs_tag"


zbeekman:
I'll try to keep my eye on CI, but please do ping me if there's anything else I can do to help get this resolved/merged! Thanks!

Rycieos:
stderr is already being nulled one function call deeper:
https://github.com/nojhan/liquidprompt/blob/145f146e9cb4fb3c30a22e6c35529120f4650a28/liquidprompt#L1607-L1609

But your patch is redirecting stdout, not stderr.

I set my Bash options to what yours are, and tested on Bash-4.4 and Bash-5.1, but I cannot reproduce this:
```bash
$ set -o errtrace
$ set -o
allexport       off
braceexpand     on
emacs           on
errexit         off
errtrace        on
functrace       off
hashall         on
histexpand      on
history         on
ignoreeof       off
interactive-comments    on
keyword         off
monitor         on
noclobber       off
noexec          off
noglob          off
nolog           off
notify          off
nounset         off
onecmd          off
physical        off
pipefail        off
posix           off
privileged      off
verbose         off
vi              off
xtrace          off
$ _lp_vcs_bookmark
$ echo $?
127
```
Are you sure that you are getting an error message printing to stdout when you call `_lp_vcs_bookmark` in a Git repo?

zbeekman:
Hmmmm, which version of Bash are you using? 

```
$ echo ${BASH_VERSION}
5.1.4(1)-release
```

Perhaps the output could be coming from the homebrew [command-not-found tap](https://github.com/homebrew/homebrew-command-not-found)?

```
brew tap homebrew/command-not-found
```

The problem with the current setup is that, yes, as you note, stderr is being redirected in the [child function](https://github.com/nojhan/liquidprompt/blob/master/liquidprompt#L1607) getting called, but the if clauses are printing this error message on stdout, and the child function writes output to stdout for VCS systems that have a notion of a bookmark, so you can't simply throw away stdout in the child/helper function. Redirecting stdout of the if test clause solved this quirk for me, and I don't see a downside to it other than a trivial increase in code length/complexity.

zbeekman:
> Are you sure that you are getting an error message printing to stdout when you call `_lp_vcs_bookmark` in a Git repo?

Yes, I am sure because when I'm on my branch from the PR it goes away and on master it comes back.

Running some further experiments, I think I can confirm that the [homebrew/homebrew-command-not-found](https://github.com/homebrew/homebrew-command-not-found) tap is responsible for this. If I untap this, and log out and back in again then I get command-not-found error messages on stderr, but then after re-tapping and sourcing .bashrc they appear on stdout:

``` bash
# after untapping and logout & back in
east06:~ ibeekman$ _lp_git_bookmark
-bash: _lp_git_bookmark: command not found
east06:~ ibeekman$ echo $?
127
east06:~ ibeekman$ _lp_git_bookmark 2> /dev/null

east06:~ ibeekman$ brew tap homebrew/command-not-found
==> Tapping homebrew/command-not-found
# ...
east06:~ ibeekman$ source .bashrc
# ...
east06:~ ibeekman$ _lp_git_bookmark > /dev/null
east06:~ ibeekman$ _lp_git_bookmark 2> /dev/null
_lp_git_bookmark: command not found
east06:~ ibeekman$ echo $?
127
```

Notice the leading `-bash:` disappears after tapping homebrew/homebrew-command-not-found and the error output switches to stdout.

IMO, while this is/could be considered a bug in Homebrew, I still think it's worthwhile to send stdout in the if test command to /dev/null to prevent stuff like this from happening.

Rycieos:
> 
> 
> which version of Bash are you using?

```
5.1.0(1)-release
```

> Perhaps the output could be coming from the homebrew [command-not-found tap](https://github.com/homebrew/homebrew-command-not-found)?

Looks like it:

https://github.com/Homebrew/homebrew-command-not-found/blob/9b28312f90822986c171c49be3d2afe014d77c3f/handler.sh#L23-L24
https://github.com/Homebrew/homebrew-command-not-found/blob/9b28312f90822986c171c49be3d2afe014d77c3f/handler.sh#L34-L35
https://github.com/Homebrew/homebrew-command-not-found/blob/9b28312f90822986c171c49be3d2afe014d77c3f/handler.sh#L41

Try reporting a bug with them? They are directing the error message to stderr on the lines for Zsh, so I don't see a reason that they shouldn't or can't copy what Bash does by default and print to stderr. You can try adding `>&2` to the end of the lines I linked above and seeing if it fixes the problem for you.

> The problem with the current setup is that, yes, as you note, stderr is being redirected in the [child function](https://github.com/nojhan/liquidprompt/blob/master/liquidprompt#L1607) getting called, but the if clauses are printing this error message on stdout, and the child function writes output to stdout for VCS systems that have a notion of a bookmark, so you can't simply throw away stdout in the child/helper function. Redirecting stdout of the if test clause solved this quirk for me, and I don't see a downside to it other than a trivial increase in code length/complexity.
> 
> IMO, while this is/could be considered a bug in Homebrew, I still think it's worthwhile to send stdout in the if test command to /dev/null to prevent stuff like this from happening.

Liquidprompt does not return data on stdout anymore, all data is passed by variables. Changing the above line 1608 to also redirect stdout to null would fix your problem (but would also need to be done for all `_lp_vcs_*()` functions, not just bookmark). It doesn't need to be done a level up, and I would prefer it, since `_lp_vcs_details_color()` is a theme function, and other user created themes would supply their own functions and wouldn't know to null stdout.

*However*, since this is an issue with homebrew-command-not-found, I would rather see it fixed there. In a normal Bash or Zsh environment, no commands inside of `_lp_vcs_bookmark()` should print anything to stdout. I think nulling stdout in this case would set a bad precedent: that we need to null all function calls in case the user's shell was configured strangely and would print things to stdout.

I'm closing this as not a Liquidprompt issue. Sorry that this isn't want you wanted, but I think this is a clear case of a shell plugin breaking expected behavior (that is, an error message should be printed on stderr).

zbeekman:
I'll try my luck with a Homebrew PR

PR #668, opened 2021-05-01
Add hyperlinks for directories in path

### LINKED ISSUE #659: Hyperlinks on path elements in graphical session
I guess most of users spend their time in the terminal under a graphical session, thus having access to the mouse and graphical apps. It happens to me quite often to start a graphical file browser from within the terminal, to do something that is more easily done in it (like image browsing, complex files selection, etc.). As of now, you have to write down the command starting the browser to actually start it from the current directory. It would be easier if it was just one click away from where the user is: the terminal.

Fortunately, there is a set of OSC escape sequences which can add an hyperlink into a text and which is now quite supported by terminal emulators (see [[1]](https://github.com/mintty/mintty/wiki/CtrlSeqs#hyperlinks)[[2]](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)).

For example, this hides the `file:///home/nojhan/code/` URL within the `code` text, which is then displayed with a dotted underline:
`printf '\e]8;;file:///home/nojhan/code/\e\\code\e]8;;\e\\\n'`

The path formatting function (`_lp_path_format`) should have an option to insert `file://` URLs behind each path element.

Rycieos:
Add a _lp_create_link() function that turns a URL and text into a operating system command sequence for a clickable hyperlink in the
terminal. See https://github.com/mintty/mintty/wiki/CtrlSeqs#hyperlinks for details.

Add sections of the current path as hyperlinks, with the "file://" protocol.

Remaining questions:

- [ ] Should links be added for other objects in the prompt? I can't think of any good ideas, other than VCS branches linking to the remote site copy of the branch. Or maybe, since Python virtualenvs IDs are nothing more than their path, they could be done the same as the path sections.
- [ ] Should there be `ENABLE` config options for each type of link?
- [ ] Should it be enabled or disabled by default?
- [ ] Should it be scoped to controlled by themes, or by Liquidprompt internals? (this current implementation uses it only in internals, not at the default theme level)
- [ ] Needs docs written.
- [ ] Needs specific test coverage.

Resolves #659

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -954,11 +954,34 @@ _lp_color_map() {
 # Generate a terminal hyperlink from a URL and text.
 _lp_create_link() {  # url, text
     (( LP_ENABLE_HYPERLINKS )) || return 2
+    lp_link="$_LP_OPEN_ESC"$'\E]8;;'"${1}"$'\E'"${_LP_BACKSLASH}${_LP_CLOSE_ESC}${2}${_LP_OPEN_ESC}"$'\E]8;;\E'"${_LP_BACKSLASH}$_LP_CLOSE_ESC"
+}
 
-    local ret
-    __lp_escape "$1"
+# Add a link to a path element, with protocol depending on the connection.
+_lp_create_link_path() { # path
+    local path="$1"
+    if (( LP_ENABLE_HYPERLINKS )); then
+        _lp_connection
 
-    link="$_LP_OPEN_ESC"$'\E]8;;'"${1}"$'\E'"${_LP_BACKSLASH}${_LP_CLOSE_ESC}${2}${_LP_OPEN_ESC}"$'\E]8;;\E'"${_LP_BACKSLASH}$_LP_CLOSE_ESC"
+        if [[ "$lp_connection" == "ssh" ]]; then
+            # SSH connection needs the `sftp` protocol to reach the remote host.
+            local client_ip client_port server_ip server_port
+            IFS=" " read -r client_ip client_port server_ip server_port <<<"$SSH_CONNECTION"
+            local username=${USER:-${USERNAME:-${LOGNAME-}}}
+            _lp_create_link "sftp://${username}@${server_ip}:${client_port}/${PWD}/" "$path"
+            lp_link_path="$lp_link"
+
+        elif [[ "$lp_connection" == "su" || "$lp_connection" == "lcl" ]]; then
+            # Reach local host with the `file` protocol.
+            _lp_create_link "file://${PWD}/" "$path"
+            lp_link_path="$lp_link"
+
+        else # no link for telnet.
+            lp_link_path="$path"
+        fi
+    else # No link.
+        lp_link_path="$path"
+    fi
 }
 
 # Return true if the input is the name of a function.
@@ -1351,7 +1374,7 @@ _lp_path_format() {
     lp_path=
     lp_path_format=
 
-    local ret link
+    local ret lp_link_path
 
     local lp_pwd_tilde
     __lp_pwd_tilde
@@ -1377,15 +1400,11 @@ _lp_path_format() {
         __lp_escape "$lp_path"
 
         if [[ $display_path == "$vcs_root_directory" ]]; then
-            lp_path_format="${vcs_root_format}${ret}"
+            _lp_create_link_path "${ret}"
+            lp_path_format="${vcs_root_format}$lp_link_path"
         else
-            lp_path_format="${last_directory_format}"
-        fi
-
-        if _lp_create_link "file://${PWD}/" "$ret"; then
-            lp_path_format+=$link
-        else
-            lp_path_format+=$ret
+            _lp_create_link_path "${ret}"
+            lp_path_format="${last_directory_format}$lp_link_path"
         fi
 
         return
@@ -1510,11 +1529,9 @@ _lp_path_format() {
         fi
 
         if (( ! is_shortening )); then
-            if _lp_create_link "file://${current_path/#"~/"/"$HOME/"}/" "$ret"; then
-                lp_path_format+=$link
-            else
-                lp_path_format+=$ret
-            fi
+
+            _lp_create_link_path "$ret"
+            lp_path_format+="$lp_link_path"
 
             if [[ -n $path_to_proccess && ( $current_path != "/" || $separator != "/" ) ]]; then
                 if [[ $current_path != "/" ]]; then
@@ -2219,6 +2236,7 @@ _lp_hostname() {
         local ret
         __lp_escape "$lp_hostname"
         lp_hostname=$ret
+
     else
         return 2
     fi
@@ -2239,14 +2257,25 @@ _lp_hostname_color() {
     fi
 
     if _lp_hostname; then
+
         case "$lp_connection" in
         lcl)
             lp_hostname_color+="@${LP_COLOR_HOST}${lp_hostname}${NO_COL}"
             ;;
         ssh)
+            local client_ip client_port server_ip server_port hostname=
+            # client_* are unused
+            # shellcheck disable=SC2034
+            IFS=" " read -r client_ip client_port server_ip server_port <<<"$SSH_CONNECTION"
+            local username=${USER:-${USERNAME:-${LOGNAME-}}}
+            if _lp_create_link "ssh://${username}@${server_ip}:${server_port}" "$lp_hostname"; then
+                hostname="$lp_link"
+            else
+                hostname="$lp_hostname"
+            fi
             # If we want a different color for each host
             (( LP_ENABLE_SSH_COLORS )) && LP_COLOR_SSH="$LP_COLOR_HOST_HASH"
-            lp_hostname_color+="@${LP_COLOR_SSH}${lp_hostname}${NO_COL}"
+            lp_hostname_color+="@${LP_COLOR_SSH}${hostname}${NO_COL}"
             ;;
         su)
             lp_hostname_color+="@${LP_COLOR_SU}${lp_hostname}${NO_COL}"
--- a/liquidprompt
+++ b/liquidprompt
@@ -43,6 +43,7 @@ if test -n "${BASH_VERSION-}"; then
 
     _LP_FIRST_INDEX=0
     _LP_PERCENT='%'    # must be escaped on zsh
+    _LP_BACKSLASH='\\' # must be escaped on bash
 
     # Escape the given strings
     # Must be used for all strings injected in PS1 that may comes from remote sources,
@@ -89,6 +90,7 @@ elif test -n "${ZSH_VERSION-}" ; then
 
     _LP_FIRST_INDEX=1
     _LP_PERCENT='%%'
+    _LP_BACKSLASH='\'
 
     __lp_escape() {
         ret="${1//\\/\\\\}"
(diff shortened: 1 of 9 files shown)


nojhan:
> Should links be added for other objects in the prompt? I can't think of any good ideas, other than VCS branches linking to the remote site copy of the branch.

The username may be used as a link toward $HOME.
The hostname may be a `sftp://` link (maybe useless for localhost, thus fallback to `file://`?).

> Or maybe, since Python virtualenvs IDs are nothing more than their path, they could be done the same as the path sections.

Yes,

> Should there be ENABLE config options for each type of link?

Definitely yes. For instance, many terminal emulator have feature allowing to add a term link on text matching a regexp, in order to run any command (I added one to Terminator myself, so as to be able to open a file in my editor at a line/column, just by clicking on a compiler output). Such a feature may come in the way, hence the necessary declutching option.

> Should it be enabled or disabled by default?

I would still say enabled by default, or else many users are not going to discover it.

> Should it be scoped to controlled by themes, or by Liquidprompt internals? (this current implementation uses it only in internals, not at the default theme level)

It's difficult to say, but I would tend to see themes as a first step toward finding common modules that would fit as internals. So if it's easy to make a modular and generic feature within the path function (which seems already quite modular, at least I was able to use it in my own theme easily), it's OK to put it in there, else, maybe consider making it a theme first, then see how to make it a common feature.

nojhan (approved):
Awesome.

Rycieos:
> I don't think you can, actually - a PR is identified in-url by an id number, which is not the branch name. That mapping would have to be done (AFAIK) server-side.

True. The best we could do is link to a search, like for this PR:

https://github.com/nojhan/liquidprompt/pulls?q=is%3Apr+7470b5fa566fda8da939c2767417a7d36e8ce05a

But we could always link to the branch, and Github might link to it?

https://github.com/nojhan/liquidprompt/tree/feature/url-links

Or link to the branch search, which would show a matching PR if it exists:

https://github.com/nojhan/liquidprompt/branches/all?query=feature%2Furl-links

nojhan:
Thinking about that, it may be tricky to have links toward any git-related website, given that not everybody use Github (at least Gitlab is quite a competitor, not even talking about other options).

I guess we need to:

1. Parse the `git branch -vv` command, to get the remote, then the `git remote show` command, to get the URL. Then we need to bet that the remote URL follows the same scheme for HTTP (which seems to be a fairly common case, but nothing guaranteed). However, that only links to the repository, not the branch. For the branch we would need to:
2. Let the user configure their own `LP_GIT_LINK_BRANCH_TEMPLATE="https://github.com/nojhan/liquidprompt/tree/%s"` (and co.).

Rycieos:
> Thinking about that, it may be tricky to have links toward any git-related website, given that not everybody use Github

Exactly, and guessing is all but guaranteed to fail at some point. I could see implementing your second idea with templates, but that would greatly reduce the amount of users that would configure to enable it.

nojhan:
A compromise would be to provide a default to Github, or to add the most known templates as comments in the shipped preset.

nojhan:
I implemented something (cherry-picking the aforementioned code to be up-to-date with the current master) on [feat+url-links](https://github.com/nojhan/liquidprompt/tree/feat+url-links).

nojhan:
I'm quite loosing track of my work, so if that's OK, I will force-push [feat+url-links](https://github.com/nojhan/liquidprompt/tree/feat+url-links) on this PR (it contains @Rycieos' commit, and this PR was too old to be merged anyway). Then it will be easier to review.

nojhan:
@Rycieos I cannot assign you a review on a PR you started, but I certainly hope you will request some changes :smile:

liquidprompt
```diff
@@ -772,6 +775,42 @@ _lp_color_map() {
...
 }
 
+# Generate a terminal hyperlink from a URL and text.
+_lp_create_link() {  # url, text
+    (( LP_ENABLE_HYPERLINKS )) || return 2
+
+    local ret
+    __lp_escape "$1"
+
+    link="$_LP_OPEN_ESC"$'\E]8;;'"${1}"$'\E'"${_LP_BACKSLASH}${_LP_CLOSE_ESC}${2}${_LP_OPEN_ESC}"$'\E]8;;\E'"${_LP_BACKSLASH}$_LP_CLOSE_ESC"
```
Rycieos, line 785: Something is off here. Why run a string through `__lp_escape()` and then not use the return value?

Either `$1` or `$2` should be escaped, or maybe both... Or neither? We need to track down where the escaping is suppose to happen.

liquidprompt
```diff
@@ -1138,10 +1177,13 @@ _lp_path_format() {
         __lp_escape "$lp_path"
 
         if [[ $display_path == "$vcs_root_directory" ]]; then
-            lp_path_format="${vcs_root_format}${ret}"
+            _lp_create_link_path "${vcs_root_format}${ret}"
+            lp_path_format="$lp_link_path"
```
Rycieos, line 1181: You can't pass the formatting into the linker. Or at least you shouldn't.
```suggestion
            _lp_create_link_path "$ret"
            lp_path_format="${vcs_root_format}${lp_link_path}"
```

liquidprompt
```diff
@@ -1912,14 +1961,21 @@ _lp_hostname_color() {
...
 
     if _lp_hostname; then
+
         case "$lp_connection" in
         lcl)
             lp_hostname_color+="@${LP_COLOR_HOST}${lp_hostname}${NO_COL}"
             ;;
         ssh)
+            local hostname=
+            if _lp_create_link "ssh://${lp_hostname}" "$lp_hostname"; then
```
Rycieos, line 1971: Thinking out loud; we can find the IP address that we connected to with
```bash
IFS=" " read -r client_ip client_port server_ip server_port <<<"$SSH_CONNECTION"
```
So then we could make the link `"ssh://${server_ip}:${server_port}"` instead, which is probably more likely to work?

tests/test_utils.sh
```diff
@@ -694,6 +699,27 @@ function test_path_format_last_dir() {
...
+  LP_ENABLE_HYPERLINKS=1
+  LP_PATH_VCS_ROOT=1
+
+  typeset lp_path lp_path_format
+
+  _lp_path_format '{format}'
+  assertEquals "root directory" '/' "$lp_path"
+  assertContains "root directory formatting" "$lp_path_format" '{format}/'
+
+  # FIXME How to test OSC-8 escape sequences?
```
Rycieos, line 720: Not easy. I suppose the only real test is to build a string based on the spec (again) to test against.

Rycieos (commented):
I can't requests changes on my own PR either 😑  But I can submit a review.

Looking pretty good. I need to go over the whole thing again in depth before I can feel confident with it, but these suggestions are at least a start.

liquidprompt
nojhan: Probably neither of them should be escaped? If I want to add formatting in the text, I for sure don't want it to be escaped. And if the URL is escaped, it has the potential to break something. I will go for no escape at all.

nojhan:
As of now, path elements are linked toward `$PWD`, because only the path label is passed to `_lp_create_link_path`, but ideally, the absolute path down to the current label would be passed, so that we can build up a link to this specific directory. However, `_lp_path_format` being such a complex function, I am not at ease at adding such a feature there.

nojhan:
@Rycieos Probably needs a review.

liquidprompt
```diff
@@ -1112,7 +1149,7 @@ _lp_path_format() {
     lp_path=
     lp_path_format=
 
-    local ret
+    local ret lp_link
```
Rycieos, line 1152:
```suggestion
    local ret lp_link_path
```

tests/test_utils.sh
```diff
@@ -694,6 +699,40 @@ function test_path_format_last_dir() {
...
+  _lp_create_link "$url" "$label"
+  local link="$_LP_OPEN_ESC"$'\E]8;;'"${url}"$'\E'"${_LP_BACKSLASH}${_LP_CLOSE_ESC}${label}${_LP_OPEN_ESC}"$'\E]8;;\E'"${_LP_BACKSLASH}$_LP_CLOSE_ESC"
+  assertEquals "typical OSC-8 escape sequence" "$lp_link" "$link"
+
+  url="https://test&eacute;.io/?var=1&dbg=2"
+  label="liquid\tlink"
+  _lp_create_link "$url" "$label"
+  link="$_LP_OPEN_ESC"$'\E]8;;'"${url}"$'\E'"${_LP_BACKSLASH}${_LP_CLOSE_ESC}${label}${_LP_OPEN_ESC}"$'\E]8;;\E'"${_LP_BACKSLASH}$_LP_CLOSE_ESC"
+  assertEquals "typical OSC-8 escape sequence with complex text" "$lp_link" "$link"
+}
```
Rycieos, line 734: Good tests, but also needs one to test the path links embedded in the formatted path.

liquidprompt
```diff
@@ -772,6 +775,40 @@ _lp_color_map() {
...
+            local client_ip client_port server_ip server_port
+            IFS=" " read -r client_ip client_port server_ip server_port <<<"$SSH_CONNECTION"
+            local hostname=$(hostname -f 2>/dev/null)
+            local username=${USER:-${USERNAME:-${LOGNAME-}}}
+            _lp_create_link "sftp://${username}@${server_ip}:${client_port}/${PWD}/" "$path"
+            lp_link_path="$lp_link"
+
+        elif [[ "$lp_connection" == "su" || "$lp_connection" == "lcl" ]]; then
+            # Reach local host with the `file` protocol.
+            _lp_create_link "file://${PWD}/" "$path"
```
Rycieos, line 801: > As of now, path elements are linked toward `$PWD`, because only the path label is passed to `_lp_create_link_path`, but ideally, the absolute path down to the current label would be passed, so that we can build up a link to this specific directory. However, `_lp_path_format` being such a complex function, I am not at ease at adding such a feature there.

Yeah, that is a bit of a problem. I agree, this function is terribly complex, so don't worry about it. I'll re-familiarize myself with it and fix this part up later when I also tackle #761

Rycieos (commented):
Request for changes

Sorry for the silence recently; work got busy, and then the holidays.

In general looking good. I want to see a bit better test coverage, since `lp_path_format` is so complicated and impossible to understand (sorry).

Ignore the failing jobs, they are fixed with commits c8722f2ac57aaaf68d720fe5ef25cb11feccf468 and 6c403a01704867fc45a2558ab1d7abbf109ea57b. A rebase would fix the failures.

nojhan:
@Rycieos: Good for another review.

tests/test_utils.sh
```diff
@@ -694,6 +699,63 @@ function test_path_format_last_dir() {
...
+  assertEquals "typical OSC-8 escape sequence" "$expected_link" "$lp_link"
+
+  url="https://test&eacute;.io/?var=1&dbg=2"
+  label="liquid\tlink"
+  expected_link="$_LP_OPEN_ESC"$'\E]8;;'"${url}"$'\E'"${_LP_BACKSLASH}${_LP_CLOSE_ESC}${label}${_LP_OPEN_ESC}"$'\E]8;;\E'"${_LP_BACKSLASH}$_LP_CLOSE_ESC"
+  _lp_create_link "$url" "$label"
+  assertEquals "typical OSC-8 escape sequence with complex text" "$expected_link" "$lp_link"
+
+  local pathword="home"
+  typeset PWD="/home/nojhan"
```
Rycieos, line 736: Why mix `local` and `typeset`? In Bash and Zsh they mean the same thing.

The rest of the tests use `typeset` (because when I wrote them I was hoping we could get to Ksh support, but I gave that up, and Ksh does not support `local`), so just use that.

Rycieos (commented):
Looks good. One easy tests issue.

This has conflicts with your prompt_comparison file, you will need to rebase to `master` to fix those before I can merge.

tests/test_utils.sh
nojhan: Just copy/pasted without thinking about it. Which one do you prefer?

Rycieos: I'd say `typeset`, to be consistent in the tests.

nojhan:
GG :beers:

PR #671, opened 2021-05-30
Add missing leading _ to some _lp_line_count

milouse:
It seems `_lp_line_count` was renamed to `__lp_line_count` (doubling the leading `_`) without a full renaming amount the source. Thus some old usage of the old name still exists in mercurial, fossil or SVN functions.

This MR fixes all these function call.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -2005,7 +2005,7 @@ _lp_hg_head_status() {
 _lp_hg_stash_count() {
     local shelves count
     shelves="$(\hg shelve --list 2>/dev/null; printf x)"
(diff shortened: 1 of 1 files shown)


Rycieos:
Yikes. It is true that at some point `__lp_line_count()` was named `_lp_line_count()`, but the first instance of it in the real history, a314677b8031804130c69de94d4604e9c319575a, it is named `__lp_line_count()`. This is obviously an artifact of many many rebasing changes before the rework merge. That and not having any tests over the VCS functions. Just goes to show despite how far the test coverage has come, we still have a long way to go.

Thanks for the PR!

PR #677, opened 2021-06-28
Unify project name as "Liquid Prompt" in docs (spelled as two words)

ottok:
On the front page README.md the project name is spelled "Liquid Prompt" as two words. This PR suggests to use that spelling across all project documentation.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -4,9 +4,9 @@ Config Options
 .. contents::
    :local:
 
-Almost every feature in Liquidprompt can be turned on or off using these config
-options. They can either be set before sourcing Liquidprompt (in ``.bashrc`` or
-``.zshrc``), or set in a Liquidprompt config file.
+Almost every feature in Liquid Prompt can be turned on or off using these config
+options. They can either be set before sourcing Liquid Prompt (in ``.bashrc`` or
+``.zshrc``), or set in a Liquid Prompt config file.
 
 .. note::
    Config variables set in a config file take precedence over variables set in the
@@ -26,7 +26,7 @@ The config file is searched for in the following locations:
 
 The first file found is sourced.
 
-Liquidprompt ships with an example config file, ``liquidpromptrc-dist``. You can
+Liquid Prompt ships with an example config file, ``liquidpromptrc-dist``. You can
 start from this file for your config::
 
     cp ~/liquidprompt/liquidpromptrc-dist ~/.config/liquidpromptrc
@@ -127,7 +127,7 @@ General
 
    * **truncate_chars_from_path_left**: Truncates characters from the start of
      the path, showing consecutive directories as one shortened section. E.g. in
-     a directory named ``~/MyProjects/Liquidprompt/tests``, it will be shortened
+     a directory named ``~/MyProjects/liquidprompt/tests``, it will be shortened
      to ``...prompt/tests``. The shortened mark is :attr:`LP_MARK_SHORTEN_PATH`.
    * **truncate_chars_from_dir_right**: Leaves the beginning of a directory name
      untouched. E.g. directories will be shortened like so: ``~/Doc.../Office``.
@@ -344,7 +344,7 @@ Features
       example, Docker doesn't inherit anything unless explicitly told to.
       Singularity in many configurations inherits most variables but shell
       functions and zsh hooks might not make it in.  For full functionality,
-      liquidprompt may need to be sourced inside the child container.
+      ``liquidprompt`` may need to be sourced inside the child container.
 
    See also: :attr:`LP_COLOR_CONTAINER`.
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,17 +1,17 @@
-.. Liquidprompt documentation master file, created by
+.. Liquid Prompt documentation master file, created by
    sphinx-quickstart on Wed Nov 18 11:00:53 2020.
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Welcome to Liquidprompt's documentation!
-========================================
+Welcome to Liquid Prompt's documentation!
+=========================================
 
-Liquidprompt is an adaptive prompt for Bash & Zsh that gives you a nicely
+Liquid Prompt is an adaptive prompt for Bash & Zsh that gives you a nicely
 displayed prompt with useful information when you need it. It does this with a
 powerful theming engine and a large array of data sources.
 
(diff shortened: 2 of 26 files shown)


Rycieos:
In my opinion, I think it makes sense as one word. The project ID is also one word; if it was two words, I would think it would be `liquid-prompt`. I would rather see any instances of "Liquid prompt" changed to "Liquidprompt" than this.

But @nojhan named it originally, so what would you like to see?

augmentedfourth:
Also note that the package name in homebrew (and I think other distribution mechanisms) is a single word, un-capitalized... just like the file name of the script itself. I'll defer to @nojhan as well, but as someone who's used this tool for over a decade "Liquid Prompt" just looks jarring.

nojhan:
That's a thought question. Do someone have access to some SEO tools to check which name leads to the largest search engine share (out of existing hits)?

Rycieos:
> Do someone have access to some SEO tools to check which name leads to the largest search engine share

I'm no SEO expert, but some quick digging leads to the following:
1. "Liquid prompt" has more than 10 times the "hits" than "Liquidprompt" does, but that is not a fair comparison. The space means it is two separate search terms, and you get a ton of false positive hits.
2. "Liquid prompt" hits mostly on blog posts about the project (so many blog posts, wow).
3. "Liquidprompt" hits mostly on documentation and packaging indexes.
4. "Liquidprompt" hits on pages with only the string "Liquid prompt" in the page.

Based on this, I think:
* Because of 4, searching with the string "liquidprompt" will hit both forms, while "liquid prompt" won't always.
* Documentation is probably the first thing we want users to find.
* Because of 1, we don't want users to find unrelated pages.

I think that "Liquidprompt" is better for SEO.

nojhan:
Seems like a good point to me :)

ottok:
I'll wait for a week for further feedback and if this is the decision, I'll revert the PR to do the change for all current Liquid Prompt into Liquidprompt.

dolmen:
As someone who has contributed to this mess, my preference was Liquid Prompt as the project name and `liquidprompt` as the command to invoke it.

But as a former maintainer, consider my vote as neutral if votes are involved.

ottok:
When looking at pages that Google finds on "liquidprompt" they are all pages that reference the package name, not the project name. As soon as you start to get results about blog posts that decribe the project, everybody seem to write about "Liquid Prompt" – two separate words, and both capitalized. I agree with that and I think it is also better English (though I am not a native speaker). One of the search results with "liquidprompt" even yields "Liquid Prompt", because that blog post is so popular.

Thus, I leave my PR leaning on "Liquid Prompt".


image

vs

image

ottok:
I also checked [Google Trends](https://trends.google.com/trends/explore?date=all&q=liquid%20prompt,liquidprompt). The name "Liquid Prompt" is more popular than spelled together, but the search volume is so low that the difference is not significant.

image

Rycieos:
Sorry for the radio silence on this. I'm not sure what the path forward should be, since we have a few on each side of the camp, and @nojhan seems to be fine either way. Ideally we would get some other opinions on this, but I realize that leaving this sitting open forever is not ideal either.

ottok:
Don't worry, leaving the issue open is fine to me, no hurry here :)

In this case taking some time to collect feedback and think about it is a
good thing.

>

nojhan:
Looking at the search engine results, I think "Liquid Prompt" would better match what the actual users are using to refer to the project.

This is a key point in my opinion, because we are facing competitors that have less interesting features and design, and still surprisingly seem to be more popular. We definitely need to reach the users *in their own words*.

Also, it emphasizes one of the key feature of the design, which is to actually be *liquid*. It is probably worth it to use the adjective as a keyword, and I think having it separated helps. One could even use "*the* Liquid Prompt", so as to emphasize even more.

It doesn't seem like a big deal to have the command/repository keeping the one-word version, I think most of people would understand, especially if the official website use the two-words version.

docs/theme/included/powerline.rst
```diff
@@ -291,9 +291,9 @@ See the `Powerline Fonts installation docs`_ for help.
 Configuration
 =============
 
-Liquidprompt Configuration
+Liquid Prompt Configuration
 --------------------------
```
Rycieos, line 295: Line needs expanding to match header.

docs/theme/included/powerline.rst
```diff
@@ -62,9 +62,9 @@ See the `Powerline Fonts installation docs`_ for help.
...
```
Rycieos, line 66: Line needs expanding to match header.

docs/theme/included/alternate_vcs.rst
```diff
@@ -49,9 +49,9 @@ A demo of what disabling the configuration options might look like:
 Configuration
 =============
 
-Liquidprompt Configuration
+Liquid Prompt Configuration
 --------------------------
```
Rycieos, line 53: Line needs expanding to match header.

docs/index.rst
```diff
@@ -1,17 +1,17 @@
-.. Liquidprompt documentation master file, created by
+.. Liquid Prompt documentation master file, created by
    sphinx-quickstart on Wed Nov 18 11:00:53 2020.
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Welcome to Liquidprompt's documentation!
+Welcome to Liquid Prompt's documentation!
 ========================================
```
Rycieos, line 7: Line needs expanding to match header.

docs/config.rst
```diff
@@ -125,7 +125,7 @@ General
 
    * **truncate_chars_from_path_left**: Truncates characters from the start of
      the path, showing consecutive directories as one shortened section. E.g. in
-     a directory named ``~/MyProjects/Liquidprompt/tests``, it will be shortened
+     a directory named ``~/MyProjects/Liquid Prompt/tests``, it will be shortened
```
Rycieos, line 128: This isn't a explicit name reference, but a directory path. Probably should leave this one as it was, even though it is only an example.

Rycieos (changes requested):
Ok, sounds like a close enough consensus then. I agree, let's unify the name.

@ottok, if you could fix these small issues I've pointed out, and maybe squash and rebase, I'll merge this in.

ottok:
Sure, I can refresh to use "Liquid Prompt" everywhere as agreed today.

ottok:
Fixed all items commented and rebased on latest master.

Rycieos (approved):
Thanks @ottok!

PR #679, opened 2021-07-06
Add AWS Vault support

elerch:
Extension to #496 to add [AWS Vault](https://github.com/99designs/aws-vault) support. AWS Vault does not set the AWS_PROFILE/AWS_DEFAULT_PROFILE variables, but does add AWS_VAULT. Since this is philosophically the same as showing an AWS profile on the prompt, I simply added AWS_VAULT to the detection chain.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -271,11 +271,14 @@ Features
    :type: bool
    :value: 1
 
-   Display the current value of :envvar:`AWS_PROFILE` or
-   :envvar:`AWS_DEFAULT_PROFILE`. These variables are used to switch between
-   configuration profiles by the `AWS CLI`_.
+   Display the current value of :envvar:`AWS_PROFILE`,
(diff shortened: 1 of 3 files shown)


Rycieos (changes requested):
Thanks for the PR! This is a good change, but we need documentation to reflect it.

You need to add details to the `docs/config.rst` file around line 270 (rendered docs here:
https://liquidprompt.readthedocs.io/en/latest/config.html#LP_ENABLE_AWS_PROFILE).
Make sure to add the link you put in the PR description so users can learn more about AWS Vault.

Also add the new env var to the `docs/functions/data.rst` file about line 114. (rendered docs: https://liquidprompt.readthedocs.io/en/latest/functions/data.html#lp_aws_profile).

elerch:
Thanks! I added the doc changes. Let me know if there is anything else you think is needed.

Rycieos (approved):
Looks good, thanks again!

PR #681, opened 2021-08-12
Add Kubernetes context default namespace

Rycieos:
Add data gathering for the configured default **namespace** of the current `kubectl` context. Display that data as a string directly after the context string, separated by a colon.

See [Setting the namespace preference](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#setting-the-namespace-preference) in the Kubernetes docs.

Example:
```[⎈cluster:my-namespace]```

While this is a separate chunk of data from the context, it is actually part of the context definition, and so it makes sense to display it along with the context.

While most Kubernetes objects are subject to the strict DNS name restrictions, kubectl contexts never have their name server-side, and so do not have the same restrictions. This means a context name can include any characters, meaning we have no safe way to parse it out of a data string. I wanted to use:

```bash
line=$(kubectl config view --minify --output 'jsonpath={.current-context}{" "}{..namespace}')
IFS=' ' read kubernetes_context kubernetes_namespace <<<"$line"
```

But since spaces can be in a context name, this would break.

CC: @imsky @mschuett @lyoshenka @ismith

Questions:
1. Is this data you want to see? I know it is for me, but do others want it?
2. Would you want this displayed in a different way?
3. Should the delimiter (`:`) be configurable? Especially since `kubectl` does not limit what characters can be in context names, so a colon could be part of a name.
4. Can you think of a better way to get this data? I really want to do it in one call instead of two, but I can't find a way to do it safely since a context can contain whatever character we pick for a delimiter.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -417,7 +417,24 @@ Features
 
    .. _`context`: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/
 
-   See also: :attr:`LP_DELIMITER_KUBECONTEXT`,
+   See also: :attr:`LP_ENABLE_KUBE_NAMESPACE`,
+   :attr:`LP_DELIMITER_KUBECONTEXT_PREFIX`,
+   :attr:`LP_DELIMITER_KUBECONTEXT_SUFFIX`,
+   :attr:`LP_COLOR_KUBECONTEXT`,
+   and :attr:`LP_MARK_KUBECONTEXT`.
+
+   .. versionadded:: 2.1
+
+.. attribute:: LP_ENABLE_KUBE_NAMESPACE
+   :type: bool
+   :value: 0
+
+   Display the current `Kubernetes <https://kubernetes.io/>`_ default
+   `namespace`_ in the current context.
+
+   .. _`namespace`: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#setting-the-namespace-preference
+
+   See also: :attr:`LP_ENABLE_KUBECONTEXT`,
    :attr:`LP_DELIMITER_KUBECONTEXT_PREFIX`,
    :attr:`LP_DELIMITER_KUBECONTEXT_SUFFIX`,
    :attr:`LP_COLOR_KUBECONTEXT`,
--- a/docs/functions/data.rst
+++ b/docs/functions/data.rst
@@ -47,7 +47,7 @@ Battery
 Development Environment
 -----------------------
 
-.. function:: _lp_kubernetes_context() -> var:lp_kubernetes_context
+.. function:: _lp_kubernetes_context() -> var:lp_kubernetes_context, var:lp_kubernetes_namespace
 
    Returns ``true`` if a Kubernetes context is found.
    Returns the Kubernetes context name or the first name component.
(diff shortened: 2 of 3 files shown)


ismith:
> 1. Is this data you want to see? I know it is for me, but do others want it?

I hadn't actually realized the default namespace was configurable in a context! So, yes, this does seem potentially desirable, but I lack experience to support that intuition. My own experience with k8s has mostly used namespaces to separate `kube-system` from apps whose code I (or my org) have written, and cluster users have had full permissions within a given cluster, and so it's not been necessary to segment contexts this way. I know that other organizations do use namespaces for cognitive + ACL isolation, in which case I imagine this becomes more salient.

> 3. Should the delimiter (:) be configurable?

This seems like an easy yes - it costs little in complexity.

> 2. Would you want this displayed in a different way?
> 3. [...]
> 4. Can you think of a better way to get this data? I really want to do it in one call instead of two, but I can't find a way to do it safely since a context can contain whatever character we pick for a delimiter.

The available options for display seem to be:
| kubernetes_context | kubernetes_namespace | notes
----------------------------|-----------------------------------|---------
| off | off | trivial case, nothing displayed |
| on | off | common (and status quo) case - I suspect most users, like me, didn't realize default namespace is configurable |
| on | on | 
| off | on | plausible, but i suspect less common than the context-only case |

It seems unlikely that in the on/on case, you'd want to display these separately. In which case, it's not really necessary to have separate variables for context and namespace, right? Just set the `jsonpath` according to:
- which delimiter is configured
- which of the above 3 (disregarding the off/off case) cases is enabled

And thus, you'd have one variable (`kubernetes_string` - possibly there's a better name) containing any of:
- `cluster`
- `cluster:my-namespace`
- `my-namespace`
depending on the `jsonpath`.

Rycieos:
> 
> 
> I hadn't actually realized the default namespace was configurable in a context!

If I wasn't clear, I am configuring this default namespace all the time, changing it to be whatever namespace I am currently working in. That is why I find the display useful. 

> It seems unlikely that in the on/on case, you'd want to display these in separate locations. In which case, it's not really necessary to have separate variables for context and namespace, right?

Huh, that is a really good point. I really like the idea, except that the newish design of Liquidprompt is that the data layer will return unedited, unopinionated data, and the theme layer (which can be completely replaced by the user) can arrange it however it wants.

Which means whatever config variable is used for that delimiter is a default theme option, and the data layer shouldn't touch it. I suppose a structure like this:

```bash
_lp_kubernetes_context() { # delimiter_string
  delim=${1-:}
  ...
  lp_kubernetes_context="${context}${delim}${namespace-}"
}

_lp_kubernetes_context_color() {
  _lp_kubernetes_context "$LP_KUBECONTEXT_DELIMITER"
  ...
}
```
could work, but that is making the assumption that every theme would want the namespace to immediately follow the context (which is exactly what you said).

However, you did get me thinking. While it is true that a context can have any character in it, a namespace cannot, since it lives as a reference in the k8s apiserver. I could split on the _last_ delimiter found, maybe something like:
```bash
line=$(kubectl config view --minify --output 'jsonpath={.current-context}{"/"}{..namespace}')
kubernetes_context=${line%/*}
kubernetes_namespace=${line##*/}
```
I think that should work nicely.

PR #692, opened 2022-01-05
feature: add the dotmatrix theme

nojhan:
The _dotmatrix_ theme is basically the look of "powerline", but done right on the user experience side.
The main idea is to highlight what's important, instead of displaying too many things because fancy colors everywhere are fun.
    
To achieve that, the dotmatrix theme has been very carefully designed:
- it uses four modules: host/path, machine state, environment, version control.
- it uses a consistent visual grammar (using 3 separators and 3 colors).
- it follows good practices for color blindness.

Screenshot:
demo_dotmatrix

The full rendering of the modules is summarized in the [design document](https://github.com/nojhan/liquidprompt/raw/theme%2Bdotmatrix/themes/dotmatrix/liquidprompt_dotmatrix_design.odt).
I've used this prompt for months and it seems ready for a beta release.

The current code uses a similar approach to the one used for the powerline theme, but needs refactoring.


themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,895 @@
...
+        DOTMATRIX_HOST_LOCAL=${POWERLINE_HOST_LOCAL:-"⌂"}
+        DOTMATRIX_USER_LOCAL=${DOTMATRIX_USER_LOCAL:-"♟"}
+        DOTMATRIX_VCS=${DOTMATRIX_VCS:-"╟┘"}
+        DOTMATRIX_MARK=${DOTMATRIX_MARK:-"■"}
+        DOTMATRIX_LINE=${DOTMATRIX_LINE:-"━"}
+        DOTMATRIX_PLUG=${DOTMATRIX_PLUG:-"▛"}
+        DOTMATRIX_SOFT=${DOTMATRIX_SOFT:-"╱"}
+        DOTMATRIX_MEAN=${DOTMATRIX_MEAN:-"▞"}
+        # Separator arrays                                (left center right 
+        [[ -z ${DOTMATRIX_LINK[@]+x} ]] && DOTMATRIX_LINK=("▙" " "   "▜")
```
Rycieos, line 32: I know it's verbose, but array checking needs to be done with `${array[@]+"${array[@]}"}` to support `set -u`. See [the wiki](https://github.com/nojhan/liquidprompt/wiki/Shell-standards#arrays). And if you are checking for a not set array to set a default, it needs to be this more complex form.

```suggestion
        DOTMATRIX_LINK=( ${DOTMATRIX_LINK[@]+"${DOTMATRIX_LINK[@]}"} )
        [[ ${#DOTMATRIX_LINK[@]} == 0 ]] && DOTMATRIX_LINK=("▙" " "   "▜")
```

themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,895 @@
...
+            side+="$dotmatrix_field"
+        else
+            __dotmatrix_make_field "$lp_time" "${DOTMATRIX_COLOR_DARK[@]}"
+            side+="$dotmatrix_field"
+        fi
+    fi
+
+    # Merge side in header
+    local side_as_text="$(_lp_as_text "${side}")"
+    local head_as_text="$(_lp_as_text "${header}")"
```
Rycieos, line 295: I hope we can come up with a better way to do this. `_lp_as_text()` not a good function. Two subshells and two `sed` calls are not fast...

themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,895 @@
...
+    # FIXME within SSH, COLUMNS seems to be broken
+    local head_width=$((${COLUMNS}-${#side_as_text}-${#head_as_text}))
+    #-6)) # Why -6? I have no idea.
+    if [[ $head_width -lt 0 ]] ; then
+        head_width=0
+    fi
+
+    local spaced_side+=$(printf "%.0s " $(seq ${head_width}))
```
Rycieos, line 303: Two more subshells, not good. I bet this would be faster as a for loop:
```bash
local i
for ((i=0; i < head_width; i++)); do
   spaced_side+=" "
done
```

Rycieos, line 303: This whole "put a section of the prompt on the right (or "backwards") side of the terminal" is a cool feature. I think we should add it to the main `liquidprompt` file as a utility function. Something like `__lp_opposite_align_prompt()`.

Zsh has this functionality built in, so we could just wrap that to save compute time on Zsh.

Rycieos (changes requested):
This looks nice! I'll run with this theme for a few days, see what I think.

I haven't dug too much into the code, as there is a lot, and you said you were going to refactor. But what I have looked at I have liked.

Bunch of thoughts:
* I'd like to see the second line full bar be optional, so I can turn it off.
* I'd like an option to not display the right section if it is empty.
* If the dotmatrix variant is the default instead of cheveron, the screenshots should show that version.
* The theme needs documentation like the other themes. See the [Powerline docs](https://liquidprompt.readthedocs.io/en/latest/theme/included/powerline.html) for an example.

Also I noted the failed build, which wasn't caused by these changes, and fixed it in cfad0e3f94c579c882e9a4c9fadb80d29b16913f.

Rycieos:
> I'd like to see the second line full bar be optional, so I can turn it off.

I see now that it contains things like Python venvs. So it probably can't be optional.

But it is missing some newer features, like Kubernetes context.

nojhan:
Thanks for the review, I will fix those things (in a near future).

docs/conf.py
```diff
@@ -30,6 +30,7 @@
 # ones.
 extensions = [
     'sphinx_rtd_theme',
+    'sphinx.ext.autodoc'
```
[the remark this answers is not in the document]
nojhan: I confess I have no idea. It is now removed.

docs/index.rst
```diff
@@ -13,6 +13,31 @@ powerful theming engine and a large array of data sources.
...
+Liquid Prompt has been carrefuly designed to highlight the *relevant*
+information, at *the right time*.
+It does not clutter your prompt with useless eye candy and helps you detect
+important alerts right away.
+Moreover, Liquid Prompt is highly configurable: you can disable useless features
+and change any alert thresholds to suit your needs.
+On top of this, it comes with several themes, from the discrete default to the
+most colorful and fancy one, àla powerline.
+
+.. image:: demo_dotmatrix.png
```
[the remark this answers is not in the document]
nojhan: I strongly disagree with that. The home page (just as the README) is a crucial part for hooking new users.
However, I agree that big screenshot may render the page difficult to read.  Thus I moved them down, after all the other content, and added a demo of the *powerline* theme, to be fair.

liquidprompt
```diff
@@ -753,7 +753,7 @@ lp_terminal_format() {  # fg, bg, bold, underline, fallback_fg, fallback_bg
     (( LP_ENABLE_COLOR )) || return 2
 
     local af_color ab_color fg bg previous_af_color
-    fg=$1
+    fg=${1:-"-1"}
```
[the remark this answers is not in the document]
nojhan: Probably just a bad refactoring on my side, sorry for the noise. Now fixed (with a comment, for the next reader).

themes/dotmatrix/dotmatrix.theme
Rycieos: I did come up with a better way to do this. Use `__lp_strip_escapes()`. See commit 6085cd21b958c16853c3bb5ea7bfdb7daf11e2cf.

docs/index.rst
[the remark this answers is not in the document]
nojhan: TBC with PR #731

nojhan:
I think I did all the requested changes (at last).

Note that some shellchecks are not passing in powerline and some bypass does not work in LP, for which I plead not guilty.

themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,1053 @@
...
+    # if _lp_XXX ; then
+    #     inline+="${inline}${DOTMATRIX_ITEMS[0]}${DOTMATRIX_SPACE_LINE}${lp_XXX}${DOTMATRIX_SPACE_LINE}${DOTMATRIX_ITEMS[2]}"
+    # fi
+    local ret
+    __lp_strip_escapes "$inline"
+    local as_text="$ret"
+    inline_width=$((${COLUMNS}-${#as_text}))
+
+    # Fill up the remaining space.
+    inline+="${lp_terminal_format}$(printf "%.0s${DOTMATRIX_LINE}" $(seq ${inline_width}))"
```
Rycieos, line 405: This is still using the `printf '%.0s'` method instead of `_lp_fill()`.

liquidprompt
```diff
@@ -826,12 +828,44 @@ _lp_sb() {
     [[ -n "$1" ]] && echo -nE " $1 "
 }
 
+
+# Adds as much $3 character (e.g. spaces) between $1 and $2,
+# so as to make the resulting string the same width than the current terminal.
+_lp_fill() {
+
+    local left="$1" right="$2" space="$3"
+    [[ -z $3 ]] && space=" "
```
Rycieos, line 837: This is simpler, right?
```suggestion
    local left="$1" right="$2" space="${3:-" "}"
```

liquidprompt
```diff
@@ -826,12 +828,44 @@ _lp_sb() {
     [[ -n "$1" ]] && echo -nE " $1 "
 }
 
+
+# Adds as much $3 character (e.g. spaces) between $1 and $2,
+# so as to make the resulting string the same width than the current terminal.
+_lp_fill() {
```
Rycieos, line 834: Love this.

Just spitballing here, would it be useful to have a function/version where you input three strings, and one of them is centered in the prompt?

```
a bunch of stuff here ----------------- central stuff here ----------------- right stuff here
```
I'm not sure what would be better, centered between the side texts, or centered in the terminal window.

themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,1053 @@
...
+    local temperature=""
+    if _lp_temperature ; then
+        __dotmatrix_make_field "$lp_temperature" ${DOTMATRIX_COLOR_DARK[@]+"${DOTMATRIX_COLOR_DARK[@]}"}
+        temperature="${dotmatrix_field}${LP_MARK_TEMP}"
+    fi
+
+    __dotmatrix_make_sep_soft ${DOTMATRIX_COLOR_SEP_DARK[@]+"${DOTMATRIX_COLOR_SEP_DARK[@]}"}
+    # Non-bracketed variables are intentionals,
+    # so that if the variable is null, it does not count in the arguments array.
+    __dotmatrix_join_fields "$dotmatrix_sep" $battery $load $temperature
```
Rycieos, line 302: This is unsafe. You can't rely on `IFS` being anything sane. Quote these things here, and in `__dotmatrix_join_fields()` on line 913, first check if each element is the empty string (`[[ -z $f ]] && continue`).

themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,1053 @@
...
+    header+="${_DOTMATRIX_USER}"
+    header+="${_DOTMATRIX_MULTIPLEXER}"
+    header+="${_DOTMATRIX_HOSTNAME}"
+    header+="${_DOTMATRIX_FS}"
+    header+="${_DOTMATRIX_PATH} "
+
+    if _lp_dirstack; then
+        __dotmatrix_make_sep_plug ${DOTMATRIX_COLOR_LITE[@]+"${DOTMATRIX_COLOR_LITE[@]}"}
+        header+="${dotmatrix_sep}"
+        __dotmatrix_make_field $lp_dirstack ${DOTMATRIX_COLOR_LITE[@]+"${DOTMATRIX_COLOR_LITE[@]}"}
```
Rycieos, line 259: `$lp_dirstack` needs to be quoted here.

See https://github.com/nojhan/liquidprompt/wiki/Shell-standards#quoting
> There are almost no scenarios where quoting a variable is incorrect, so unless you know for sure that not quoting it is safe, just quote it!

themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,1053 @@
...
+__dotmatrix_make_field() { # string, (color array)
+    local lp_terminal_format string="${1-}"
+    [[ -z $string ]] && return 1
+
+    shift
+    local format="$@"
+    if [[ -z $format ]] ; then
+        lp_terminal_format -1
+    else
+        lp_terminal_format $format
```
Rycieos, line 927: This should not work. If it is working as you expect, it is a lucky bug, but I doubt it will work correctly on all shells.

`var="$@"` does _not_ make `var` an array. Much better to just call `lp_terminal_format "$@"`, that will work exactly as expected.

I would also skip the 
```bash
    if [[ -z $format ]] ; then
        lp_terminal_format -1
```
part, this is a private function to the theme, just enforce the method signature of always being explicit with what formatting you want.

themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,1053 @@
...
+    [[ -z $bg ]] && bg=-1
+    local fbg="${6-}"
+    [[ -z $fbg ]] && fbg=-1
+
+    sep="${DOTMATRIX_PLUG}"
+    if [[ "${_lp_last_ab_color-}" == "$bg" ]] ; then
+        sep="${DOTMATRIX_MEAN}"
+    fi
+
+    lp_terminal_format -3 $bg 0 0 -1 $fbg
```
Rycieos, line 951: Again, these need to be quoted. Try setting `IFS` to something like `$'\n\r \t1'` and see how much of this code breaks. (I tested it, it's a lot.)

Rycieos (changes requested):
This is getting pretty good. I like the look of this theme, quite polished. I love what you have been able to do with the theming engine; means it is moving in the right direction.

Quite a few issues with quoting though, and with usage of `$@`. Note that you didn't include this theme file in the `shellcheck.sh` script, otherwise Shellcheck would have caught all of those for you. Hopefully once you do that Shellcheck can help you fix those, but let me know if you need help.

docs/theme/included/dotmatrix.rst
```diff
@@ -0,0 +1,493 @@
...
+
+Design
+======
+
+Rationale
+---------
+
+The DotMatrix theme use a couple of design principles:
+
+* It is accessible to colorblind people, hence it uses:
```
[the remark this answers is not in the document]
nojhan: "Blue-yellow form" is (by far) the rarest condition, hence why it is recommended for accessibility. But of course, the user may override `DOTMATRIX_COLOR_*` variables.

It would probably be nice to have an option to automatically configure themes in that regard. Maybe a good use-case for a interactive theme configuration bot àla Powerlevel.

docs/theme/included/dotmatrix.rst
```diff
@@ -0,0 +1,493 @@
...
+- :attr:`LP_MARK_BATTERY`,
+- :attr:`LP_MARK_LOAD`,
+- :attr:`LP_MARK_TEMP`,
+- :attr:`LP_PS1_PREFIX`.
+
+The following options are **not honored**:
+
+- every `LP_COLOR_\*` options.
+- all the others `LP_MARK_\*` options not listed in the above section.
+- :attr:`LP_USER_ALWAYS` (at least a user icon is always displayed).
```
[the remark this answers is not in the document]
nojhan: On of DotMatrix’ core design is to use space between segments to carry information. For instance, the link between the "terminal type" and "user" segments actually indicate the remote connection type. The same goes for the link between the user and the host segment. If one remove the user segment, then the information carried by the spacing is ambiguous.
I tried to use small placeholders, but the result was not very readable. So at the end, the normal state is an icon, while the non-login user gets a name.

liquidprompt
nojhan: If not more readable, at least probably a little bit faster.

liquidprompt
nojhan: Definitely centered between the side texts.
I guess this may be useful for some hypothetical future theme.

themes/dotmatrix/dotmatrix.theme
nojhan: Very sorry… I did a round trip on the code and tried to be consistent.

nojhan:
Thanks again for the review, I hope it’s not too much.

Rycieos (changes requested):
> Thanks again for the review, I hope it’s not too much.

Nah, I love code reviews. And I haven't had any reviews to do at work lately, so this fills that void. And I wouldn't do this unless I wanted to.

Nice progress. Only small issues now, except that it doesn't seem to work at all in Zsh. When I try to load it, it just prints something cryptic and doesn't change the theme:
```
% lp_theme dotmatrix
lp_terminal_format=$'%{\C-[(B\C-[[m\C-[[30m\C-[[48;5;244m%}'
```
Maybe something to do with array indexes?

themes/dotmatrix/dotmatrix.theme
```diff
@@ -0,0 +1,1063 @@
...
+        local line_width=$((${COLUMNS:-80}-${#left_as_text}))
+
+        local line="$inline"
+        local i c f a b j
+        for ((i=0; i < line_width; i++)); do # For each character in the line.
+            local dash_colors=()
+            # for ((c=0; c < colors_nb-1; c++)); do # For each color parameter.
+            for ((c=0; c < 6; c++)); do # For each color parameter.
+                a="${DOTMATRIX_COLOR_LINE_START[c]}"
+                b="${DOTMATRIX_COLOR_LINE_END[c]}"
```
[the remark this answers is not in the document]
nojhan: Is there a portable way to handle the last index as well?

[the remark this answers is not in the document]
Rycieos: Just adding `_LP_FIRST_INDEX` to both 0 and 6 in the for loop should work.

Rycieos (changes requested):
Only one docs issue.

The Zsh issue is still happening. I've determined that the theme is being activated, but PS1 is not updating, and it also prints `lp_terminal_format=$'%{\C-[(B\C-[[m\C-[[30m\C-[[48;5;244m%}'` to the terminal.

Very strange, as when I turn on debugging, I see `PS1` be set to something, but if I print its value right after, it shows as the PS1 from the previous theme.

If I have some time tomorrow I'll try to do more to track it down.

nojhan:
This moved to its own repository: https://github.com/nojhan/lp-dotmatrix

PR #702, opened 2022-03-15
Add option LP_USE_SUDO_VALIDATE to change sudo credential testing

Mice7R:
Validation of sudo credentials is done by invoking a command (true).

This means that, if the user has sudo credentials, a root session is opened and then closed. Potentially generating two PAM messages plus the sudo line for true being executed as root:
```
sudo:     user : TTY=pts/4 ; PWD=/home/user/Projects/liquidprompt ; USER=root ; COMMAND=/usr/bin/true
sudo: pam_unix(sudo:session): session opened for user root(uid=0) by user(uid=1000)
sudo: pam_unix(sudo:session): session closed for user root
```
This makes filtering this messages rather complex. Specially the pam ones that cannot be distinguished form other sudo commands.

This commit adds an option to use the _validate_ command (sudo -v) to test for credentials. The difference is that **_validate_ does not generate any log on success**. ie. If the user has cached credentials.

In case the user does not have cached credentials, or not in sudoers, it  has the same behaviour as before: error line:
```
sudo:     user : a password is required ; TTY=pts/7 ; PWD=/home/user/Projects/liquidprompt; USER=root ; COMMAND=validate
sudo:     nobody : command not allowed ; TTY=pts/3 ; PWD=/home/user/Projects/liquidprompt ; USER=root ; COMMAND=validate
```
However, this lines are easy to filter out by searching COMMAND=validate

I've opted for adding LP_USE_SUDO_VALIDATE instead of replacing the default behaviour because I'm not sure if -v is on every system. 
The default value is 0: using the old behaviour.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -551,17 +551,7 @@ Features
       run ``sudo`` once each prompt. This is likely to make your sysadmin hate
       you.
 
-   See also: :attr:`LP_COLOR_MARK_SUDO`, :attr:`LP_USE_SUDO_VALIDATE`.
-
-.. attribute:: LP_USE_SUDO_VALIDATE
-   :type: bool
-   :value: 0
-
-   Use sudo validate command ``sudo -v`` to check for valid credentials.
-
-   :attr:`LP_ENABLE_SUDO` must be enabled for this to have any effect.
-
-   See also: :attr:`LP_ENABLE_SUDO`.
+   See also: :attr:`LP_COLOR_MARK_SUDO`.
 
 .. attribute:: LP_ENABLE_SVN
    :type: bool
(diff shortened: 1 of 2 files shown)


Rycieos (changes requested):
According to the Changelog:
https://www.sudo.ws/releases/changelog/
The `-v` flag is 27 years old, since the first version of `sudo`. And `sudo` is kept updated by all OSes since it is so critical to security.

My worry is why this was not done this way to begin with. Is there really no good reason not to use the `-v` flag? Seems perfectly safe to me.

I say remove the config option completely. Replace the current line with your suggestion, since it seems to be a straight improvement.

Rycieos:
Thanks for the PR!

PR #719, opened 2022-07-26
 Add Fig as an installation method

brendanfalk:
We recently listed on `liquidprompt` on [Fig Plugin Store](https://fig.io/plugins), so we'd love to have Fig displayed as a download method on your readme. Thanks so much and please let me know if you have any questions!

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -44,6 +44,14 @@ add the following lines in your ``.zshrc`` after activating Zinit::
     zinit ice ver"stable" lucid nocd
     zinit light nojhan/liquidprompt
 
+Installation via Fig
+----------------------
+
+Install `liquidprompt` with `Fig <https://fig.io/>`_ in just one click.
+
+.. image:: https://fig.io/badges/install-with-fig.svg
+   :target: https://fig.io/plugins/other/liquidprompt
+
 Dependencies
 ============
 


Rycieos:
Thanks for the PR. I gave Fig a try, it's a cool concept. I like the GUI config page for plugins.

This would be the first endorsement in our docs of a proprietary product, so I would be curious what @nojhan, @dolmen, or others think of that. I don't think I would reject it based off that, but I also will not merge this right away without that discussion first.

There is another issue though, in that I think your company is violating the terms of the [AGPLv3 license](https://github.com/nojhan/liquidprompt/blob/master/LICENSE) and/or copyright. (I am not an attorney, these are only my opinions). Violations as I see them:
1. You have copied our demo image to [your plugins repo](https://github.com/withfig/plugins/blob/e8bd94d1bc3c2a7819f40824b534b15f0495ef53/plugins/liquidprompt/images/3.png) in [this commit](https://github.com/withfig/plugins/commit/9a91211deee40413eb3efc3d9fc10d9e8fab7bd1). I am not sure if the AGPLv3 covers images for documentation purposes like this, but if so this is a violation by copying it and putting it under the MIT license. If not, then the original copyright covers it, which @nojhan owns as the creator of the image.
2. Similarly, you have copied much of our documentation to [your plugins repo](https://github.com/withfig/plugins/blob/e8bd94d1bc3c2a7819f40824b534b15f0495ef53/plugins/liquidprompt/index.ts). I again do not know if AGPLv3 covers textual documentation, but if it doesn't, it would fall under the original author's copyright, which is me.

How I would like to see this resolved:
1. Remove the image, and link to the original in this repo.
2. License the content in https://github.com/withfig/plugins/blob/e8bd94d1bc3c2a7819f40824b534b15f0495ef53/plugins/liquidprompt under the AGPLv3.

I welcome discussion. Especially a third party attorney chiming in 😁

brendanfalk:
@Rycieos You raise a great point here re AGPLv3. Admittedly, this is a complete oversight on our end. Very happy to discuss with you how we can make this work. Your proposal seems very reasonable and easy to do. 

I also want to check with some people more experienced in this space to make sure we aren't continuing to violate the license after these changes. Give me a day or two to learn more and I will get back.

nojhan:
Hi. I see no problem in mentioning proprietary products, as I see no problem in supporting features related to proprietary software or proprietary software using Liquidprompt.

About the license, @Rycieos is right : 

- As for me, the screenshot is AGPL. But even if it was not the case, it would be copyrighted by default, so an explicit authorization would have been needed. So in any cases, you should at least consider indicating the license.
- Given that the (actual state of) the liquidprompt directory in Fig makes no use of the liquidprompt content (if we forget about the screenshot being still there), I don't see that it should be licensed AGPL. Albeit this would be necessary if some of liquidprompt code/doc was there, of course. Note that I would not ask for removal, but rather explicit licensing.

Rycieos:
Good callout @milouse. If the CodeWhisperer wants to be included in our documentation, the are welcome to open a new PR.

PR #733, opened 2022-09-05
refactor the dev section, feat: adds cmake

nojhan:
- Adds the display of the main CMake cache variables (compiler, generator, build type).
- Refactor the development tool section
    - in a list that is easy to parse visually,
    - instead of the previous set of space-separated keywords, too close to the VCS section,
    - some of which had brackets (now removed).
    - Adds config var for opening, separating and closing the section.
- Adds the corresponding docs.
- Update the prompt comparison to add CMake.
    - Update the dev tools scale and others' marks.


liquidprompt
```diff
@@ -929,6 +935,27 @@ __lp_theme_zsh_complete() {
...
+    lp_hash_color="${color}${str}${NO_COL}"
+}
+
+_lp_join() {
+    # Join the $2... array with the $1 string.
+    local d=${1-}
+    local f=${2-}
+    if shift 2; then
+            lp_join="$(printf '%s' "$f" "${@/#/$d}")"
+    fi
```
Rycieos, line 956: A subshell is going to be slower than this:
```suggestion
    shift 2
    local IFS=""
    lp_join="${first}${*/#/$delimiter}"
```

liquidprompt
```diff
@@ -929,6 +935,27 @@ __lp_theme_zsh_complete() {
...
+}
```
Rycieos, line 957: Both of these functions need unit tests added.

liquidprompt
```diff
@@ -3311,61 +3363,67 @@ _lp_default_theme_prompt_data() {
...
 
     if _lp_aws_profile_color; then
-        LP_AWS_PROFILE=" $lp_aws_profile_color"
+        LP_AWS_PROFILE="$lp_aws_profile_color"
     else
         LP_AWS_PROFILE=
     fi
 
+    if _lp_cmake; then
+        LP_CMAKE="${lp_cmake_compiler}${LP_MARK_CMAKE}${lp_cmake_generator}${LP_MARK_CMAKE}${lp_cmake_buildtype}"
```
Rycieos, line 3414: This string building should not be done here, but in a `_lp_cmake_color()` function. Both to follow previous standards, and to keep reusable theming accessible to other themes.

liquidprompt
```diff
@@ -3390,10 +3448,20 @@ _lp_default_theme_prompt_template() {
...
-        PS1+="${LP_SCLS}${LP_AWS_PROFILE}${LP_CONTAINER}${LP_VENV}${LP_NODE_VENV}${LP_RUBY_VENV}${LP_TFSPACE}${LP_KUBECONTEXT}${LP_PROXY}"
+        local dev_all=("${LP_SCLS}" "${LP_AWS_PROFILE}" "${LP_CONTAINER}" "${LP_VENV}" "${LP_NODE_VENV}" "${LP_RUBY_VENV}" "${LP_TFSPACE}" "${LP_KUBECONTEXT}" "${LP_CMAKE}")
+        local dev_active=()
+        for i in "${dev_all[@]}"; do
+            if [[ "$i" ]]; then
+                dev_active+=("$i")
+            fi
+        done
+        if [[ ${#dev_active[*]} -gt 0 ]] ; then
+            _lp_join "${LP_MARK_DEV_MID}" "${dev_active[@]}"
```
Rycieos, line 3462: Need to declare `local lp_join`.

liquidprompt
```diff
@@ -3404,7 +3472,7 @@ _lp_default_theme_prompt_template() {
         PS1+="${LP_RUNTIME}${LP_ERR}${LP_MARK_PREFIX}${LP_COLOR_MARK}${LP_MARK}${LP_PS1_POSTFIX}"
 
         # Get the core sections without prompt escapes and make them into a title.
-        _lp_formatted_title "${LP_PS1_PREFIX}${LP_BRACKET_OPEN}${LP_USER}${LP_HOST}${LP_MARK_PERM}${lp_path-}${LP_BRACKET_CLOSE}${LP_MARK_PREFIX}${LP_MARK}${LP_PS1_POSTFIX}"
+        _lp_formatted_title "${LP_PS1_PREFIX}${LP_BRACKET_OPEN}${LP_USER}${LP_HOST}${LP_MARK_PERM}${lp_path-}${LP_BRACKET_CLOSE}${LP_MARK_PREFIX}${LP_SHLVL}${LP_MARK}${LP_PS1_POSTFIX}"
```
Rycieos, line 3475: What is the reasoning behind adding this to the title? The point was to only have the core sections; what is used as most default prompts.

I don't think you meant to put it here, maybe you thought you were moving it?

liquidprompt
```diff
@@ -3390,10 +3448,20 @@ _lp_default_theme_prompt_template() {
         # add title escape time, jobs, load and battery
         PS1="${LP_PS1_PREFIX}${LP_TIME}${LP_BATT}${LP_LOAD}${LP_TEMP}${LP_WIFI}${LP_JOBS}"
         # add user, host, permissions colon, working directory, and dirstack
-        PS1+="${LP_BRACKET_OPEN}${LP_USER}${LP_HOST}${LP_PERM}${LP_PWD}${LP_DIRSTACK}${LP_BRACKET_CLOSE}${LP_SHLVL}"
+        PS1+="${LP_BRACKET_OPEN}${LP_USER}${LP_HOST}${LP_PERM}${LP_PWD}${LP_DIRSTACK}${LP_BRACKET_CLOSE}${LP_PROXY}"
```
Rycieos, line 3451: If you are moving `LP_PROXY`, the docs need to be updated with the new order.

liquidprompt
```diff
@@ -3390,10 +3448,20 @@ _lp_default_theme_prompt_template() {
...
+        local dev_active=()
+        for i in "${dev_all[@]}"; do
+            if [[ "$i" ]]; then
+                dev_active+=("$i")
+            fi
+        done
+        if [[ ${#dev_active[*]} -gt 0 ]] ; then
+            _lp_join "${LP_MARK_DEV_MID}" "${dev_active[@]}"
+            PS1+=" ${LP_MARK_DEV_OPEN}${lp_join}${LP_MARK_DEV_CLOSE}"
+        fi
```
Rycieos, line 3464: This should be broken out into a function, maybe `_lp_dev_env_color()`.

It also should be assigned to a variable like `$LP_DEV_ENVS`, so those who want to use it this new way but in a different spot can do so. And good on keeping the original `$LP_SCLS` and such.

liquidprompt
```diff
@@ -1397,9 +1424,34 @@ _lp_kubernetes_context() {
...
 }
 
+_lp_cmake() {
+    (( LP_ENABLE_CMAKE )) || return 2
+    [[ ! -f CMakeCache.txt ]] && return 1
+
+    local cmake_compiler
+    IFS="=" read -r _ cmake_compiler <<<"$(grep COMPILER:FILEPATH CMakeCache.txt)"
+    cmake_compiler="${cmake_compiler##*/}" # Only the part after the last slash.
+    _lp_hash_color "$cmake_compiler"
```
Rycieos, line 1437: Needs `local lp_hash_color`.

liquidprompt
```diff
@@ -1397,9 +1424,34 @@ _lp_kubernetes_context() {
...
+    local cmake_generator
+    IFS="=" read -r _ cmake_generator <<<"$(grep CMAKE_GENERATOR:INTERNAL CMakeCache.txt)"
+    # Shorten: Makefiles -> Make, and Visual Studio -> VS.
+    cmake_generator="${cmake_generator/%Makefiles/Make}"
+    cmake_generator="${cmake_generator/#Visual Studio/VS}"
+    _lp_hash_color "$cmake_generator"
+    lp_cmake_generator="$lp_hash_color"
+
+    local cmake_buildtype
+    IFS="=" read -r _ cmake_buildtype <<<"$(grep CMAKE_BUILD_TYPE:STRING CMakeCache.txt)"
```
Rycieos, line 1449: This code is nice and clean, but calling a subshell 3 times and opening the same file also 3 times isn't the fastest.

I already did some crazy `grep` replacement in the python env:
https://github.com/nojhan/liquidprompt/blob/42dbe9ce08058fbb60d9bfef91625661c36d62b2/liquidprompt#L1235-L1245
Not saying you must, but something to consider.

liquidprompt
```diff
@@ -311,6 +309,13 @@ __lp_source_config() {
     LP_MARK_DIRSTACK="${LP_MARK_DIRSTACK:-"⚞"}"
     LP_MARK_SHLVL="${LP_MARK_SHLVL:-"└"}"
     LP_MARK_WIFI="${LP_MARK_WIFI:-"📶"}"
+    LP_MARK_DEV_OPEN="${LP_MARK_DEV_OPEN:-"<"}"
+    LP_MARK_DEV_CLOSE="${LP_MARK_DEV_CLOSE:-">"}"
```
Rycieos, line 313: Breaking out these marks is the right call.

I think that they should stick with the previous default of `[...]` though, I don't see a good reason to change it.

liquidprompt
```diff
@@ -929,6 +935,27 @@ __lp_theme_zsh_complete() {
...
 }
 
+_lp_hash_color() {
+    # compute the hash of the argument and color it with the corresponding color in:
+    # (green, yellow, blue, purple, cyan)
+    # Red (would be index 1) is not used here, as it should be reserved for alerts.
+    local str=${1-}
+    local cksum="$(printf '%s' "$str" | cksum)"
+    local hash="${cksum%%[$' \t']*}"
+    __lp_foreground_color "$(( 2 + hash % 5 ))"
```
Rycieos, line 945: By skipping red, you prevent this from being used for `__lp_hostname_hash()` (since we can't change that now without changing all hostname color mappings).

I get your logic, but maybe there is a way to make it work yet?

```suggestion
    __lp_foreground_color "$(( 2 + (hash - 1) % 6 ))"
```
This would keep previous colors the same, but change red to white. Hm, white probably isn't the best choice. Maybe it can't work, but I'll leave my thoughts here anyway.

docs/config.rst
```diff
@@ -476,6 +427,56 @@ Features
...
+     without the cluster and user parts.
+
+   If set to the empty string no truncating will occur (this is the default).
+
+   See also: :attr:`LP_ENABLE_KUBECONTEXT`,
+   :attr:`LP_DELIMITER_KUBECONTEXT_PREFIX`, :attr:`LP_COLOR_KUBECONTEXT`,
+   and :attr:`LP_MARK_KUBECONTEXT`.
+
+   .. versionadded:: 2.1
+
```
Rycieos, line 479: These are in alphabetical order to make them easier to find. They link to each other, so they don't need to follow their specific enable option.

Rycieos (changes requested):

Rycieos:
In general looks good, but I think it should be split into 2 PRs, one for the dev env changes, and one for adding CMake. Or at least two commits.

liquidprompt
nojhan: I think I may have added the cmake section, notice it was missing, added it, then realized it was the title and removed the cmake section.

liquidprompt
nojhan: It is faster to visually parse sections with different types of brackets.

Squared brackets are already used for the central part of the prompt. Parentheses are already used in the VCS section.

That leaves us with either `<…|…>` or `{…|…}`. I picked at random, but curly braces seems also good, especially for people being used to maths. I tested both for some time, and I find that `<…|…>` seems faster to spot and easier to parse, as it's less similar to the other sections.

liquidprompt
nojhan: I thought a bit about that, but red seems less a problem for the hostname, as I can't think of a "dangerous" host.
However, I can think of "dangerous" development environment states, that I would not want to be mistaken with other states.

So at the end I just let the hostname code as is.

liquidprompt
Rycieos: That's sound logic, you have convinced me.

We'll want to make sure we add that reasoning to the changelog and release notes.

docs/functions/util.rst
```diff
@@ -101,3 +101,18 @@ These functions are designed to be used by themes.
...
 
+.. function:: _lp_hash_color(str)
+
+   Colorize the given string with a color depending on its hash.
+   The color is chosen among: (green, yellow, blue, purple, cyan).
+   Note that the red color is not a candidate, as it should be reserved for alerts.
+
+   .. versionadded:: 2.2
+
+.. function:: _lp_join(delimiter, items...)
```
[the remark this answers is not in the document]
Rycieos: It should be in the signature, not the body description:
```suggestion
.. function:: _lp_join(delimiter, items...) -> var:lp_join
```
Same above.

docs/config.rst
```diff
@@ -318,6 +319,17 @@ Features
 
    See also: :attr:`LP_MARK_BZR`.
 
+.. attribute:: LP_ENABLE_CMAKE
+    :type: bool
```
Rycieos, line 323: Indentation in the rest of this file is 3 spaces (don't get me started on why, Sphinx started it that way and I didn't think to change it until too late).

liquidprompt
```diff
@@ -1261,7 +1288,7 @@ _lp_python_env() {
 _lp_python_env_color() {
     _lp_python_env || return "$?"
 
-    lp_python_env_color="[${LP_COLOR_VIRTUALENV}${lp_python_env}${NO_COL}]"
+    lp_python_env_color="${LP_COLOR_VIRTUALENV}${lp_python_env}${NO_COL}"
```
Rycieos, line 1291: Each one of these changes really should have a `.versionchanged: 2.2` note added to their documentation saying that they no longer include the `[...]`, as they are intended to be used with `_lp_dev_env_color()`.

liquidprompt
```diff
@@ -1397,9 +1424,53 @@ _lp_kubernetes_context() {
...
+    IFS="=" read -r _ cmake_compiler <<<"$(grep COMPILER:FILEPATH CMakeCache.txt)"
+    lp_cmake_compiler="${cmake_compiler##*/}" # Only the part after the last slash.
+
+    local cmake_generator
+    IFS="=" read -r _ cmake_generator <<<"$(grep CMAKE_GENERATOR:INTERNAL CMakeCache.txt)"
+    # Shorten: Makefiles -> Make, and Visual Studio -> VS.
+    cmake_generator="${cmake_generator/%Makefiles/Make}"
+    lp_cmake_generator="${cmake_generator/#Visual Studio/VS}"
+
+    local cmake_buildtype
```
Rycieos, line 1444: > SC2034 (warning): cmake_buildtype appears unused. Verify use (or export if used externally).

You have the wrong declaration here. Though you don't really need it at all, since you don't manipulate it after you set it.

liquidprompt
```diff
@@ -1397,9 +1424,53 @@ _lp_kubernetes_context() {
...
+    IFS="=" read -r _ lp_cmake_buildtype <<<"$(grep CMAKE_BUILD_TYPE:STRING CMakeCache.txt)"
+
+    [[ -z "${lp_cmake_compiler}${lp_cmake_generator}${lp_cmake_buildtype}" ]] && return 3
+
+    lp_cmake="${lp_cmake_compiler}${LP_MARK_CMAKE}${lp_cmake_generator}${LP_MARK_CMAKE}${lp_cmake_buildtype}"
```
Rycieos, line 1449: > SC2034 (warning): lp_cmake appears unused. Verify use (or export if used externally).

Yeah, it is unused. I think you should just delete this line altogether, and just return your 3 data strings and expect the theme to figure them out.

tests/test_utils.sh
```diff
@@ -710,6 +710,31 @@ function test_is_function {
...
+
+    _lp_join "+" ""
+    assertEquals "$lp_join" ""
+
+    typeset -a arr
+    arr=(1 2 3)
+    _lp_join "/" "${arr[@]}"
+    assertEquals "$lp_join" "1/2/3"
+}
+
```
Rycieos, line 736: You have the expected and actual arguments switched around in all of these tests, it should be `assertEquals "1/2/3" "$lp_join"`

liquidprompt
```diff
@@ -929,6 +935,27 @@ __lp_theme_zsh_complete() {
...
 
+_lp_hash_color() {
+    # compute the hash of the argument and color it with the corresponding color in:
+    # (green, yellow, blue, purple, cyan)
+    # Red (would be index 1) is not used here, as it should be reserved for alerts.
+    local str=${1-}
+    local cksum="$(printf '%s' "$str" | cksum)"
+    local hash="${cksum%%[$' \t']*}"
+    __lp_foreground_color "$(( 2 + hash % 5 ))"
+    local color="${_LP_OPEN_ESC}${af_color}${_LP_CLOSE_ESC}"
```
Rycieos, line 946: `__lp_foreground_color()` does not promise to return `$af_color`, that is why it is failing in the tests. It does this when color is disabled.

I think the correct fix is to check if `LP_ENABLE_COLOR` is enabled, or return right at the start of this function:
```bash
(( LP_ENABLE_COLOR )) || return 2
```
Obviously then this function will not always set `$lp_hash_color`, so you will need to protect where it is used.

liquidprompt
```diff
@@ -3376,6 +3475,13 @@ _lp_default_theme_prompt_data() {
         LP_VCS=
     fi
 
+    # Should be called after the dev env group above.
+    if _lp_dev_env_color ; then
+        LP_DEV_ENV="$lp_dev_env_color"
```
Rycieos, line 3480: This variable needs to be added to the `docs/theme/default.rst` file.

liquidprompt
```diff
@@ -3186,6 +3257,28 @@ _lp_container_color() {
     lp_container_color="«${LP_COLOR_CONTAINER}${lp_container}${NO_COL}»"
 }
 
+
+_lp_dev_env_color() {
```
Rycieos, line 3261: This function needs to be documented in the `docs/functions/theme.rst` file.

liquidprompt
```diff
@@ -1397,9 +1424,53 @@ _lp_kubernetes_context() {
...
+
+    local cmake_buildtype
+    IFS="=" read -r _ lp_cmake_buildtype <<<"$(grep CMAKE_BUILD_TYPE:STRING CMakeCache.txt)"
+
+    [[ -z "${lp_cmake_compiler}${lp_cmake_generator}${lp_cmake_buildtype}" ]] && return 3
+
+    lp_cmake="${lp_cmake_compiler}${LP_MARK_CMAKE}${lp_cmake_generator}${LP_MARK_CMAKE}${lp_cmake_buildtype}"
+}
+
+_lp_cmake_color() {
```
Rycieos, line 1452: This function needs to be documented in the `docs/functions/theme.rst` file.

liquidprompt
```diff
@@ -1397,9 +1424,53 @@ _lp_kubernetes_context() {
 _lp_kubernetes_context_color() {
     _lp_kubernetes_context || return "$?"
 
-    lp_kubernetes_context_color="[${LP_MARK_KUBECONTEXT}${LP_COLOR_KUBECONTEXT}${lp_kubernetes_context}${lp_kubernetes_namespace+:}${lp_kubernetes_namespace-}${NO_COL}]"
+    lp_kubernetes_context_color="${LP_MARK_KUBECONTEXT}${LP_COLOR_KUBECONTEXT}${lp_kubernetes_context}${lp_kubernetes_namespace+:}${lp_kubernetes_namespace-}${NO_COL}"
 }
 
+_lp_cmake() {
```
Rycieos, line 1430: This function needs to be documented in the `docs/functions/data.rst` file.

tests/test_utils.sh
```diff
@@ -710,6 +710,31 @@ function test_is_function {
   unalias not_my_function
 }
 
+function test_hash_color {
+    PS1="$ "
+    lp_activate --no-config # For having _lp_foreground
+    _lp_hash_color "Debug"
+    assertContains "$lp_hash_color" "Debug"
+    # FIXME How to test color?
```
Rycieos, line 718: Good question.

Since the tests can be run without color support (and in Github Actions will be), I think there is no valid test we can do to see if color was added. I think your test here is decent enough, as it will catch unset errors (like it already did).

Rycieos (changes requested):
Missing a lot of docs yet. Thanks for splitting the commits, sorry it will be a pain to amend them now.

Don't forget to make sure the tests are passing.

docs/config.rst
nojhan: It's a kind of super-power to be able to spot such discrepencies :open_mouth:

tests/test_utils.sh
nojhan: Out of curiosity, why does it matter? I would understand for `assertContains`, but for `assertEquals`?

Rycieos: Because the output of the test error labels each string incorrectly:

> ASSERT:expected:<A_B   C D EE> but was:<A_B _ _ C_ D _EE>

liquidprompt
```diff
@@ -1397,9 +1433,61 @@ _lp_kubernetes_context() {
...
+        elif [[ "$lp_cmake_buildtype" == "RelWithDebInfo" ]] ; then
+            lp_cmake_buildtype="${LP_COLOR_CMAKE_RWDI}${lp_cmake_buildtype}${NO_COL}"
+        elif [[ "$lp_cmake_buildtype" == "Debug" ]] ; then
+            lp_cmake_buildtype="${LP_COLOR_CMAKE_DEBUG}${lp_cmake_buildtype}${NO_COL}"
+        else
+            _lp_hash_color "$lp_cmake_buildtype"
+            lp_cmake_buildtype="$lp_hash_color"
+        fi
+
+        lp_cmake_color="${lp_cmake_compiler}${LP_MARK_CMAKE}${lp_cmake_generator}${LP_MARK_CMAKE}${lp_cmake_buildtype}"
```
Rycieos, line 1483: Why overwrite these variables? I think string construction will be cleaner, and will leave the original variables they way they were:

```suggestion
        _lp_hash_color "$lp_cmake_compiler"
        lp_cmake_color="${lp_hash_color}${LP_MARK_CMAKE}"

        _lp_hash_color "$lp_cmake_generator"
        lp_cmake_color+="${lp_hash_color}${LP_MARK_CMAKE}"

        if [[ "$lp_cmake_buildtype" == "Release" ]] ; then
            lp_cmake_color+="${LP_COLOR_CMAKE_RELEASE}${lp_cmake_buildtype}${NO_COL}"
        elif [[ "$lp_cmake_buildtype" == "RelWithDebInfo" ]] ; then
            lp_cmake_color+="${LP_COLOR_CMAKE_RWDI}${lp_cmake_buildtype}${NO_COL}"
        elif [[ "$lp_cmake_buildtype" == "Debug" ]] ; then
            lp_cmake_color+="${LP_COLOR_CMAKE_DEBUG}${lp_cmake_buildtype}${NO_COL}"
        else
            _lp_hash_color "$lp_cmake_buildtype"
            lp_cmake_color+="${lp_hash_color}${NO_COL}"
        fi
```

docs/functions/theme.rst
```diff
@@ -117,13 +120,47 @@ specific text and formatting may change.
       Return code matches data function.
       Return method changed from stdout.
 
+.. function:: _lp_cmake_color() -> var:lp_cmake_compiler, var:lp_cmake_generator, var:lp_cmake_buildtype
```
Rycieos, line 123: This is wrong, it returns `lp_cmake_color`
```suggestion
.. function:: _lp_cmake_color() -> var:lp_cmake_color
```

liquidprompt
```diff
@@ -929,6 +938,33 @@ __lp_theme_zsh_complete() {
...
+    lp_hash_color="${color}${str}${NO_COL}"
+}
+
+_lp_join() {
+    # Join the $2... array with the $1 string.
+    local delimiter=${1-}
+    local IFS=""
+    if (( _LP_SHELL_zsh )); then
+        shift 1
+        lp_join="${(pj/$delimiter/)*}"
```
Rycieos, line 960: Needs a Shellcheck ignore flag, as this is Zsh specific syntax
```suggestion
        # shellcheck disable=SC2296
        lp_join="${(pj/$delimiter/)*}"
```

liquidprompt
```diff
@@ -1397,9 +1433,61 @@ _lp_kubernetes_context() {
...
+    # Remove all white spaces.
+    lp_cmake_generator="${cmake_generator// /}"
+
+    IFS="=" read -r _ lp_cmake_buildtype <<<"$(grep CMAKE_BUILD_TYPE:STRING CMakeCache.txt)"
+
+    [[ -n "${lp_cmake_compiler}${lp_cmake_generator}${lp_cmake_buildtype}" ]] || return 3
+}
+
+_lp_cmake_color() {
+    (( LP_ENABLE_COLOR )) || return 2
```
Rycieos, line 1461: I don't think you want this check here. Most of our color functions work just fine with color disabled, a better name for them would be "formatted".

liquidprompt
Rycieos: I was only partially right. Turns out there are still some bugs/quirks with how LP handles non-color environments that I need to work out. I can fix them in the future, but for now to prevent issues and make the tests pass, make this change
```suggestion
    local color="${_LP_OPEN_ESC}${af_color-}${_LP_CLOSE_ESC}"
```

Rycieos (changes requested):
Looks pretty good, but still some docs issues.

You can squash and split into two commits whenever you want. If you don't now, I'll remind you before I merge.

liquidprompt
```diff
@@ -3310,62 +3417,68 @@ _lp_default_theme_prompt_data() {
...
 
+    if _lp_cmake_color; then
+        LP_CMAKE="$lp_cmake_color"
+    else
+        LP_CMAKE=
+    fi
+
     if _lp_runtime_color; then
-        LP_RUNTIME=" $lp_runtime_color"
+        LP_RUNTIME="$lp_runtime_color"
```
Rycieos, line 3475: This space should not have been removed. It now renders incorrectly.

liquidprompt
```diff
@@ -3310,62 +3417,68 @@ _lp_default_theme_prompt_data() {
...
     else
         LP_RUNTIME=
     fi
 
     if _lp_error_color; then
-        LP_ERR=" $lp_error_color"
+        LP_ERR="$lp_error_color"
```
Rycieos, line 3481: This space should not have been removed. It now renders incorrectly.

liquidprompt
```diff
@@ -3376,6 +3489,13 @@ _lp_default_theme_prompt_data() {
         LP_VCS=
     fi
 
+    # Should be called after the dev env group above.
+    if _lp_dev_env_color ; then
+        LP_DEV_ENV="$lp_dev_env_color"
```
Rycieos, line 3494: Shouldn't this lead with a space?
```suggestion
        LP_DEV_ENV=" $lp_dev_env_color"
```
I see that the space is added on line 3287, but I think it should be added here instead like the other functions do, so other themes using the function can get it without any opinionated spacing.

liquidprompt
```diff
@@ -3310,62 +3417,68 @@ _lp_default_theme_prompt_data() {
         LP_SHLVL=
     fi
 
-    if _lp_python_env_color; then
-        LP_VENV=" $lp_python_env_color"
+    if _lp_python_env_color ; then
```
Rycieos, line 3420: Undo this changed line.

liquidprompt
```diff
@@ -3310,62 +3417,68 @@ _lp_default_theme_prompt_data() {
...
     if _lp_software_collections_color; then
-        LP_SCLS=" $lp_software_collections_color"
+        LP_SCLS="$lp_software_collections_color"
     else
         LP_SCLS=
     fi
 
     if _lp_aws_profile_color; then
-        LP_AWS_PROFILE=" $lp_aws_profile_color"
+        LP_AWS_PROFILE="$lp_aws_profile_color"
```
Rycieos, line 3463: I've though about this more, and I don't think these changes are needed, or even correct.

By changing the values of `LP_AWS_PROFILE` and similar, we need to add lots of notes to the changelog and release notes, and those who want to use the "old" way will need to change their configs.

The better way to do it is on line 3277 use the values of the return variables:
```diff
-    local dev_all=("${LP_SCLS}" "${LP_AWS_PROFILE}" "${LP_CONTAINER}" "${LP_VENV}" "${LP_NODE_VENV}" "${LP_RUBY_VENV}" "${LP_TFSPACE}" "${LP_KUBECONTEXT}" "${LP_CMAKE}")
+    local dev_all=("${lp_software_collections_color-}" "${lp_aws_profile_color-}" "${lp_container_color-}" "${lp_python_env_color-}" "${lp_node_env_color-}" "${lp_ruby_env_color-}" "${lp_terraform_env_color-}" "${lp_kubernetes_context_color-}" "${lp_cmake_color-}")
```
(Probably want to wrap that line to make it readable.)

This also means we can change our code here to keep `LP_AWS_PROFILE` and similar exactly the same by adding our bracket wrapping in this location:
```bash
        LP_AWS_PROFILE=" ${LP_MARK_DEV_OPEN}${lp_aws_profile_color}${LP_MARK_DEV_CLOSE}"
```

liquidprompt
```diff
@@ -3310,62 +3417,68 @@ _lp_default_theme_prompt_data() {
...
     else
         LP_AWS_PROFILE=
     fi
 
+    if _lp_cmake_color; then
+        LP_CMAKE="$lp_cmake_color"
```
Rycieos, line 3469: Add a space to this one for consistency
```suggestion
        LP_CMAKE=" $lp_cmake_color"
```

liquidprompt
```diff
@@ -1397,9 +1433,61 @@ _lp_kubernetes_context() {
...
+    IFS="=" read -r _ cmake_generator <<<"$(grep CMAKE_GENERATOR:INTERNAL CMakeCache.txt)"
+    # Shorten: Makefiles -> Make, and Visual Studio -> VS.
+    cmake_generator="${cmake_generator/%Makefiles/Make}"
+    cmake_generator="${cmake_generator/#Visual Studio/VS}"
+    # Remove all white spaces.
+    lp_cmake_generator="${cmake_generator// /}"
+
+    IFS="=" read -r _ lp_cmake_buildtype <<<"$(grep CMAKE_BUILD_TYPE:STRING CMakeCache.txt)"
+
+    [[ -n "${lp_cmake_compiler}${lp_cmake_generator}${lp_cmake_buildtype}" ]] || return 3
```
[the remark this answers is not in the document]
Rycieos: This needs an answer before I can merge.

docs/theme/default.rst
```diff
@@ -182,65 +182,23 @@ default order if the user does not configure a different template.
...
-   .. versionadded:: 2.1
-
-.. attribute:: LP_VENV
+.. attribute:: LP_DEV_ENV
 
-   The current Python (or Conda) virtual environment. See
-   :attr:`LP_ENABLE_VIRTUALENV`.
-
-.. attribute:: LP_NODE_VENV
+   Sections related to development environments (see section `dev_env`_ below).
```
Rycieos, line 199: Either here or in the linked Dev Env section, there needs to be a note explaining that `LP_DEV_ENV` is the new way to do the same thing that all the others did before, and how if a user doesn't like the order they can use the old sections manually.

docs/functions/util.rst
```diff
@@ -101,3 +101,18 @@ These functions are designed to be used by themes.
...
+   Colorize the given string with a color depending on its hash.
+   The color is chosen among: (green, yellow, blue, purple, cyan).
+   Note that the red color is not a candidate, as it should be reserved for alerts.
+
+   .. versionadded:: 2.2
+
+.. function:: _lp_join(delimiter, items...) -> var:lp_join
+
+   Join all strings in items with the given deliminter.
+   Example: `_lp_join ", " "a" "b" "c"` will render `lp_join="a, b, c"`
```
Rycieos, line 115: Needs double backtics to render as code block.
```suggestion
   Example: ``_lp_join ", " "a" "b" "c"`` will render ``lp_join="a, b, c"``.
```

docs/config.rst
```diff
@@ -29,15 +29,15 @@ The first file found is sourced.
 Liquid Prompt ships with an example config file, ``liquidpromptrc-dist``. You can
 start from this file for your config::
 
-    cp ~/liquidprompt/liquidpromptrc-dist ~/.config/liquidpromptrc
+   cp ~/liquidprompt/liquidpromptrc-dist ~/.config/liquidpromptrc
```
Rycieos, line 32: Undo this change.

docs/config.rst
```diff
@@ -29,15 +29,15 @@ The first file found is sourced.
...
 
 In the event that you synchronize your configuration file across multiple
 computers, or if you have an ``/etc/liquidpromptrc`` system-wide from which
 you'd like to make minor deviations in an individual user account, you can
 augment the primary config to add in any local modifications using lines such
 as these::
 
-    LOCAL_RCFILE=$HOME/.liquidpromptrc.local
+   LOCAL_RCFILE=$HOME/.liquidpromptrc.local
```
Rycieos, line 40: Undo this change.

Rycieos (changes requested):
More nit picking. I keep finding stuff, sorry.

liquidprompt
nojhan: I would rather make the default config use brackets for open and close and space + double brackets for middle sep ("] ["). That way we get the old behavior.

I still think is a bad design, speaking ergonomics: it adds a lot of visual clutter and confusion with the path section. I'd rather have a good design and a cumbersome changelog (not so many people read it anyway) as a default.

We may also add a new theme, to get the old behavior just a command away.

liquidprompt
nojhan: It's in the dev section, so it should be handled by `LP_MARK_DEV_MID`.

liquidprompt
Rycieos: > I still think is a bad design

I agree, your new design is better. But we need to keep compatibility, which means these variables needs to stick around. I see these options:

1. Leave the variables as they are/were, but add the new `LP_DEV_ENVS` to PS1 in the default theme, and don't use `LP_AWS_PROFILE ` and similar.
2. Remove `LP_AWS_PROFILE ` and similar, but make a theme that uses them instead.

I don't see any reason to pick 2 over 1.

liquidprompt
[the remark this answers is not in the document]
nojhan: It was just to have different error code for different issues, in case the calling function would need it. I did not think there was some semantics in the number. What would be the benefit of having two different issues leading to the same error code?

liquidprompt
nojhan: Oh, I understand what you want, now, sorry for being slow. Yes, I think that’s definitely the way to go.

liquidprompt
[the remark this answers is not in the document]
Rycieos: I would argue that it is not two different issues, but two different paths to the same issue, being "no data available".

This is the standard for our data functions:

https://liquidprompt.readthedocs.io/en/stable/functions/data.html
> All data functions will return true (meaning return code 0) when they are both enabled and have data. They will return false (meaning return code 1) when they do not have data. Most will return 2 when they are disabled

I could see an argument that no `CMakeCache.txt` file is different from one existing but not containing the data we care about. But I think those semantics are too nuanced to try to indicate in a prompt. It is also not how the rest of our data functions work currently.

liquidprompt
```diff
@@ -3376,6 +3527,13 @@ _lp_default_theme_prompt_data() {
         LP_VCS=
     fi
 
+    # Should be called after the dev env group above.
+    if _lp_dev_env_color ; then
```
Rycieos, line 3531: Now that the sections are no longer used by default in the theme, Shellcheck is yelling about them. Add `# shellcheck disable=SC2034` the line above `_lp_default_theme_prompt_data() {` to fix that.

liquidprompt
```diff
@@ -929,6 +938,56 @@ __lp_theme_zsh_complete() {
...
+
+    local delimiter="$2"
+    shift 2
+    local fields=("$@")
+
+    declare -g -A lp_grep_fields # Global associative array.
+    local line key
+    while IFS='' read -r line ; do
+        for key in ${fields[@]+"${fields[@]}"} ; do
+            if [[ "$line" == "${key}${delitimer}"* ]] ; then
```
Rycieos, line 982:
```suggestion
            if [[ "$line" == "${key}${delimiter}"* ]] ; then
```

liquidprompt
```diff
@@ -1397,9 +1456,63 @@ _lp_kubernetes_context() {
...
+
+    _lp_grep_fields "CMakeCache.txt" "=" "CMAKE_C_COMPILER:FILEPATH"  "CMAKE_CXX_COMPILER:FILEPATH" "CMAKE_GENERATOR:INTERNAL" "CMAKE_BUILD_TYPE:STRING"
+
+    local cmake_c_compiler=${lp_grep_fields["CMAKE_C_COMPILER:FILEPATH"]}
+    lp_cmake_c_compiler="${cmake_c_compiler##*/}" # Only the part after the last slash.
+
+    local cmake_cxx_compiler=${lp_grep_fields["CMAKE_CXX_COMPILER:FILEPATH"]}
+    lp_cmake_cxx_compiler="${cmake_cxx_compiler##*/}" # Only the part after the last slash.
+
+    locafieldsmake_generator=${lp_grep_fields["CMAKE_GENERATOR:INTERNAL"]}
```
Rycieos, line 1475:
```suggestion
    local cmake_generator=${lp_grep_fields["CMAKE_GENERATOR:INTERNAL"]}
```

Rycieos (changes requested):
Super small things now (except for the Bash 3.2 issue). I'd be willing to chat about that one.

docs/config.rst
```diff
@@ -937,6 +966,38 @@ Marks
...
 
+.. attribute:: LP_MARK_DEV_CLOSE
+   :type: string
+   :value: ">"
+
+   Closing of the "development tools" section.
+
+   .. versionadded:: 2.2
+
+.. attribute:: LP_MARK_CMAKE
```
Rycieos, line 977: This block still not alphabetical.

tests/test_utils.sh
```diff
@@ -710,6 +710,51 @@ function test_is_function {
...
+    assertEquals "" "$lp_join"
+
+    typeset -a arr
+    arr=(1 2 3)
+    _lp_join "/" "${arr[@]}"
+    assertEquals "1/2/3" "$lp_join"
+}
+
+function test_grep_fields {
+    filename=$(mktemp)
```
Rycieos, line 738: Need to remove the temp file at the end of the test.

tests/test_utils.sh
```diff
@@ -710,6 +710,51 @@ function test_is_function {
...
+    echo -e "key1:value1\nkey2:value2" > "$filename"
```
Rycieos, line 739: `echo -e` is brittle, prefer `printf` everywhere.
```suggestion
    printf 'key1:value1\nkey2:value2' > "$filename"
```

tests/test_utils.sh
```diff
@@ -710,6 +710,51 @@ function test_is_function {
...
+    _lp_grep_fields "$filename" ":" "key1"
+    assertEquals ${lp_grep_fields[_LP_FIRST_INDEX+0]} "value1"
+    assertEquals ${#lp_grep_fields[@]} 1
+
+    echo -e "key1=value1\nkey2=value=key2=?\nkey3==val" > "$filename"
+    _lp_grep_fields "$filename" "=" "key1" "key2" "key3"
+    assertEquals ${lp_grep_fields[_LP_FIRST_INDEX+0]} "value1"
+    assertEquals ${lp_grep_fields[_LP_FIRST_INDEX+1]} "value=key2=?"
+    assertEquals ${lp_grep_fields[_LP_FIRST_INDEX+2]} "=val"
+    assertEquals ${#lp_grep_fields[@]} 3
```
Rycieos, line 755: Good test cases, I would also like to see one where the keys are in a different order in the source file than when they are passed in as arguments. Should be as simple as adding a `_lp_grep_fields "$filename" "=" "key3" "key1" "key2"` then the exact same asserts.

Also needs a test case for when the key is NOT in the file. The index should be unset, so just verifying that. (can test for unset value with `[[ -z ${array[i]+x} ]]`, so probably like `assertTrue "found match for non existent key:value" '[[ -z "${lp_grep_fields[_LP_FIRST_INDEX+1]+x}" ]]'`

docs/functions/util.rst
```diff
@@ -26,6 +26,16 @@ These functions are designed to be used by themes.
 
    The returned string is a fully escaped terminal formatting sequence.
 
+.. function:: _lp_grep_fields(filename, delimiter, keys...) -> var:lp_grep_fields
+
+   Parse the given filename for one key/value pairs of the form "<key><delimiter><value>"
+   (e.g. "this=that") on each line.
+   Sets an associative array containing the parsed values, for each key.
```
Rycieos, line 33: It no longer returns an associative array, so this doc needs updating.

Make sure to note that doing a for each loop over the results (like `for item in "${lp_grep_fields[@]}"`) is NOT safe, since missing keys will just be skipped, so an array that looks like
```bash
array[0]=foo
array[2]=bar
```
will expand to exactly those two items, and there will be no indication that index 1 is missing. So explicit indexing (and unset variable checking) is needed. As well as using the `_LP_FIRST_INDEX` offset.

Rycieos (changes requested):
Truly small stuff now. I'm confident if you fix these things we can get this merged. So if you want to prevent another review cycle, also squash back down to 2 commits while you are at it.

docs/config.rst
nojhan: :man_facepalming:

tests/test_utils.sh
nojhan: Interestingly, switching to `printf` introduced a very subtle bug: the `while read line` construction will not read the last line if the read file does not end with an explicit new line. To ensure to always read the file up to the very last line, one should use `while IFS='' read -r line || [[ -n "$line" ]]`. This may affect your `_lp_python_env` function as well.

Rycieos: Yes, that is a valid observation. There are a lot of utilities that assume all files end in a newline, but unless you can think of a drawback to assuming characters without a newline is a valid line, I say let's use the `while IFS='' read -r line || [[ -n "$line" ]]` form.

And yes, I was planning on converting `_lp_python_env()` to use this new search function once we had this merged.

docs/config.rst
Rycieos: To be fair, this isn't something you should need to worry about. Or even me for that matter. This should be something that is checked by the docs linter. Shouldn't be that hard for me to throw something together...

tests/test_utils.sh
```diff
@@ -710,6 +710,79 @@ function test_is_function {
...
+
+    # Empty file.
+    printf '\n' > "$filename"
+    _lp_grep_fields "$filename" ":=" "nokey"
+    assertTrue "found a non existing key" '[[ -z "${lp_grep_fields[_LP_FIRST_INDEX+0]+x}" ]]'
+
+    # Really empty file.
+    printf '\n' > "$filename"
+    _lp_grep_fields "$filename" ":=" "nokey"
+    assertTrue "found a non existing key" '[[ -z "${lp_grep_fields[_LP_FIRST_INDEX+0]+x}" ]]'
```
Rycieos, line 781: This test is the same as the above one?

liquidprompt
```diff
@@ -1397,9 +1459,63 @@ _lp_kubernetes_context() {
...
 }
 
+
+_lp_cmake() {
+    ((  LP_ENABLE_CMAKE  )) || return 2
+    [[ -f CMakeCache.txt ]] || return 1
+
+    _lp_grep_fields "CMakeCache.txt" "=" "CMAKE_C_COMPILER:FILEPATH"  "CMAKE_CXX_COMPILER:FILEPATH" "CMAKE_GENERATOR:INTERNAL" "CMAKE_BUILD_TYPE:STRING"
+
+    local cmake_c_compiler=${lp_grep_fields[_LP_FIRST_INDEX+0]}
```
Rycieos, line 1472: Need to protect all these if `set -u` is set.
```suggestion
    local cmake_c_compiler=${lp_grep_fields[_LP_FIRST_INDEX+0]-}
```
It looks like the rest of these functions will work "correctly" if these return nothing, though not optimized since it will hash and then color the empty string.

docs/functions/util.rst
```diff
@@ -26,6 +26,34 @@ These functions are designed to be used by themes.
...
+   Parse the given filename for one key/value pairs of the form "<key><delimiter><value>"
+   (e.g. "this=that") on each line.
+   Sets an array containing the parsed values, for each key in the same order the function was called.
+
+   .. code-block:: sh
+      :caption: Example of use
+
+      _lp_grep_fields "CMakeCache.txt" "=" "CMAKE_C_COMPILER:FILEPATH" "CMAKE_CXX_COMPILER:FILEPATH"
+      cmake_c_compiler=${lp_grep_fields[_LP_FIRST_INDEX+0]}
+      cmake_cxx_compiler=${lp_grep_fields[_LP_FIRST_INDEX+1]}
```
Rycieos, line 40: Good example, but these usages need to be protected for users using their shell in `set -u` mode.
```suggestion
      cmake_c_compiler=${lp_grep_fields[_LP_FIRST_INDEX+0]-}
      cmake_cxx_compiler=${lp_grep_fields[_LP_FIRST_INDEX+1]-}
```

Rycieos (changes requested):
This has to be the end of it. Looks good otherwise.

tests/test_utils.sh
nojhan: :man_facepalming:

Rycieos (approved):
Thanks for working hard on this! Finally approving and merging.

I rewrote your commits, mostly to change what was in each commit. I also had to fix a nasty bug only on Zsh 5.0 in how it handles the line `local var=()`. I added those details to the shell standards.

PR #734, opened 2022-09-19
Theme+neon

nojhan:
Neon screenshot

**feat(theme): adds "Neon"**
    
- The ``neon`` theme is an extension of the default theme.
- This prompt is a fully usable theme, designed to be easier
  to visually parse than the default theme.
- It is also an example of how to build a theme modifiyng
  colors and structure of the sections.
- The theme features three rows:
    - the first one hosting "secondary" information, in blue (jobs, sensors, dev context, etc.),
    - the second row displays "primary" information (user, path, VCS, etc.), in green,
    - the third one is dedicated to the prompt (runtime, error code, prompt mark), in cyan.
- Within each row, important notice are displayed in bold
  (and thus probably in a slightly lighter color, depending on your terminal configuration).

**feat(config): LP_MARK_JOBS_SEPARATOR**
    
- Was hard-coded as "/", keep this default and make it configurable.


Rycieos:
This looks nice. I am (usually) not a fan of multi-line prompts, but this one is pretty clean. I probably just like it because it is quite similar to default LP.

The big question that needs to be answered: **_How many themes do we want to merge into and maintain as part of Liquid Prompt?_**

My initial answer to this question was "**exactly one (1)**" when I built the theming engine. I built the Powerline theme mostly as a proof of concept, but also so Liquid Prompt would have a theme it shipped with. Then I realized that the basic Powerline theme missed out on a lot of data sources, so I built `powerline_full` to make it a real example theme. Then `alternate_vcs` came along because of a user request, and to have an example of a theme that modifies the default theme.

The issue is that each of these themes need to be kept updated along with Liquid Prompt. In the past 2 years since I added the new theming engine, the biggest mistake contributors have made is forgetting/not knowing about these themes and not updating them when adding new data sources.

To myself I had said "no more themes" because even just for myself keeping them updated was a pain (even though only the `powerline_full` theme really needed to be updated most of the time!) But now you want to merge in 3 new themes! (which is awesome, I'm not complaining)

I see 3 paths forward (in my order of preference):
1. Draw the line where it is, say "no more themes". Any new themes can be put in their own repo (you could make a `nojhan/liquidprompt-themes` repo, with a directory for each theme (in this case I would be happy to still code review for you)). We even have [a wiki page for user contributed themes](https://github.com/nojhan/liquidprompt/wiki/Themes), but it is so far empty. Documentation would be "external", but it would allow mainline Liquid Prompt development to not be blocked by keeping all the themes updated.
2. Merge in as many themes as we want, but allow them to become outdated. This would make it easy for users to find the themes, and for contributors to add features, since they could ignore the themes. Then those who care for their themes could make updates to them when they can. I would be worried about user experience, where users expect all themes to "just work" after install, and likely many would not.
3. Merge in some (or all) themes, and require that they are kept up-to-date with Liquidprompt. This would make it more difficult for contributors, especially first time contributors, as well as slow down normal development.

I'd like to hear your thoughts on this.

As for this PR, I see these issues after a quick glance:
1. The demo images should be generated with the `tools/theme-preview.sh` script, to show as many data sources as possible, and to keep the same data values between themes.
2. On that note, the `tools/theme-preview.sh` script does not show the newlines correctly. I think your error is on line 287, where you include the string "\n" in `PS1`. Bash will likely interpret this correctly, but not always. Better to include a literal newline (`$'\n'` is one way to do that).
3. The docs claim that "almost any modern font should work", but I found that to not be true. The forward slash character you are using (``, on line 122), is a private use Unicode character. A nerd font drew it correctly. (This command will find any private Unicode characters in a file: `grep -P '[\x{E000}-\x{F8FF}\x{F0000}-\x{FFFFD}\x{100000}-\x{10FFFD}]' themes/neon/neon.theme`.)
4. Switching to the theme works correctly, but when switching back to the default theme, many marks are not switched back! I only gave it a quick glance, but I think this is because you override variables like `LP_MARK_BRACKET_OPEN` so that you can reuse the functions of the default theme. This clearly does cut down on the amount of code duplication, but themes need to be self contained so that `lp_theme` will completely replace them.
5. It looks like `_lp_neon_theme_prompt()` duplicates a lot of `_lp_default_theme_prompt_data()`. Probably some way to reuse it.

nojhan:
It's fun you like it, because it's my least favorite one :)

My main motivation with adding (a few) themes is to reach for new users/contributors.
I find it a pity that LP is (quite objectively) the best prompt out there, and it's still the one with the smallest community. So my objective at the moment is to add what's missing to compete with the other prompt systems (you already saw my wishlist of feature requests, and one must add some way to add more dev-env-related sections, possibly as plugins, if that's too much of a burden to maintain.).

The maintenance problem you mention is a fair point, and I would not want to integrate as many themes as possible, but my point is that we are far from having this problem right now.
What I see now is that we have a very well thought theme engine that's completely invisible to potential contributors, because *we basically have no theme*. If we can't show off a various set of themes, it can only man we actually have no theme engine. Which would be a pity, because I find it to be the best theme engine among all the competitors, *by far* (I mean, look at the lack of variety in their themes, comparing to what I as able to do quite easily).

I am not a big fan of theming myself, and would rather stick to a carefully crafted setup (DotMatrix, for instance, on which I spent countless hours of fine tuning) than switching between dozens of themes. **BUT** of course one size does not fit all (as you and I are proof, liking very different prompts styles), and I can only observe that a large-enough set of themes is a *major* selling point for new users.

My preferred path to a compromise between marketing/features and maintenance would be to have *a few* included themes, showing all the major styles of themes, as diverse as possible, and as well crafted as possible. Any other theme that would not meet those criteria would be in a separate user-based repository, possibly with an easy-to-use import-from-Git mechanism to activate them. This does not prevent us from tagging them in the docs with something like "last updated in 2.1, may miss some new features".

So, how many themes would I suggest? I was initially planning to add at least one more theme with two variants, to match what's actually available on the market of prompt systems. That is, a "*ship" theme (using words as connectors between sections, and icons for every section), with a variant for a theme *à la* "Pure" (same, and very limited feature subset).
I think an additional "extravagant" theme would be good, but just to show-off.
That would made up for two "real" themes and four "marketing" ones:

1. Liquid theme (text-based, minimal colors, good classical UX).
2. DotMatrix (background-colored segments, *very* good UX, but atypical).
3. Powerline theme (background-colored segments, bad UX).
4. Neon theme (text-based, minimal color, fancy structure, average UX).
5. PureShip theme (TBD, word-based, fancy colored and icons, poor UX).
6. An extravagant theme (not sure how, but something that shows off crazy many-lines ASCII-art with rainbows and such. Very, very bad UX, but that's not the point).

I think with that, we cover what's available (and feasible for us), so users may compare and see that LP can actually be configured to match their needs. Ideally ending using one of the two first themes, if they take some time to think about it (I would probably write a doc page about UX to help them understanding).

BTW, if you fancy IRC and would want to chat, I am alone on `irc:libera.chat/#liquidprompt`.


[I'll came back at the PR later on, thanks for the review, already.]

nojhan:
I keep thinking about your (very valid) points, and on a second thought, maybe there is a better compromise:

- We make a command that allow to easily clone/install themes from Git, like `prompt_theme install nojhan/lp-neon`,
- We add a dynamic page to the website, that allows to automatically gather all git[hub|lab] repositories which feature a `liquidprompt-theme` tag, and display its screenshot.
- We move all themes (but the default) on separate repositories, configured accordingly.

That way, we avoid the maintenance nightmare you're mentioning, but we still have an easy access to theme installation, and we can show them on the home page.

nojhan:
This theme has been moved to its own repo: https://github.com/nojhan/lp-neon

PR #736, opened 2022-09-19
feat(themes): better powerline themes

nojhan:
Powerpuff screenshot

feat(themes): better powerline themes
    
- Refactor powerline colors:
    - cleaner gradient of color,
    - light/dark meta-sections alternance.
    - Adds a specific TAG color for `prompt_tag`.
    - Adds X11/Txt section.
- Rename `powerline_full` in `powerline_plus`.
- Backport `_lp_fill` from theme+dotmatrix.
- Add two sub-themes:
    - `power2lines`: display prompt after the colored line.
        - Adds meta-sections with gaps.
    - `powerpuff`: spread power2lines' sections on two rows and left/right aligned.
- Adds a "Themes overview" section for having all screenshots visible at once.


nojhan:
FTR, the *Powerline* theme would preferably be moved out of the included theme and get its own repo. I will make this PR there, once @Rycieos has moved the codebase.

nojhan:
Should one merge and then extract, or extract and then merge? I would rank this theme as important for v2.2, wherever it is hosted.

Rycieos:
I think the right order is extract and then merge. I've been hesitating since I don't know if I want to actually remove a theme from the project (breaking changes are bad), but I agree it is the right move, so I will make it happen (at some point) before v2.2

nojhan:
@Rycieos since think I added most (if not all) of the features I wanted for the 2.2, should we do something about those powerline themes?

Rycieos:
Yeah, now is probably the right time to do that. I'll try to get to it this week.

Rycieos:
I have release v2.2.0-rc.1. It's time to come back to this.

I don't want to remove the theme code, in case anyone is using it, I don't want to make backwards breaking changes.

That said, I do like the idea of splitting it off still. So here are my thoughts:
* Copy off the powerline and powerline_full themes into a separate repo.
* Move their docs.
* Add docs here describing the move and deprecating the versions here.
* You can make your PR against that repo.
* 
Thoughts?

nojhan:
Seems like a good plan to me.

Rycieos:
I have finished splitting this out. That took quite a bit, but I have a separate repo, with it's own docs. Please open this PR in https://github.com/liquidprompt/liquidprompt-powerline.

Let me know if you want me to wait to release v2.2 here until that PR is merged. I don't think that is necessary; we could release v1.0 there and v2.2 here, and then once your changes are ready, we could release v1.1 there.

nojhan:
I would rather wait that the main themes are aligned on the feature set of 2.2. I think this will make a more impactful release event, especially given that I am planning to market it broadly.

I also thought that it will be probably easier to understand for users if the themes follows the same versions than LP (i.e. being versioned 2.2.0 when they support LP 2.2.0). That way, you immediately know if the versions are aligned.

Should I also move my themes within the liquidprompt organization, or should we restrict them to a limited number?

Rycieos:
> I would rather wait that the main themes are aligned on the feature set of 2.2. I think this will make a more impactful release event, especially given that I am planning to market it broadly.

I am fine with that.
 
> I also thought that it will be probably easier to understand for users if the themes follows the same versions than LP (i.e. being versioned 2.2.0 when they support LP 2.2.0). That way, you immediately know if the versions are aligned.

This is a larger discussion that I think should be had. I've broken it out to https://github.com/liquidprompt/liquidprompt/discussions/799

> Should I also move my themes within the liquidprompt organization, or should we restrict them to a limited number?

I have no issue with you doing so. My only concern would be if we did not keep the themes up to date with changes in Liquid Prompt itself. I think anything in the org should be kept to the same quality as the main project. If you think you can do that, then go for it.

nojhan:
> I have no issue with you doing so. My only concern would be if we did not keep the themes up to date with changes in Liquid Prompt itself. I think anything in the org should be kept to the same quality as the main project. If you think you can do that, then go for it.

I am not sure I can commit to that... Let's keep them aside.

PR #737, opened 2022-10-02
doc(contributing): adds a technical checklist

nojhan:
Trying to build up on my last reviews, to avoid missing points.


Rycieos (approved):
Nice, looks great. These are details that we need for sure.

I made a few fixes, just formatting.

PR #738, opened 2022-10-06
feat(utils): adds the lp_fill function

nojhan:
Simulate right-aligned string part by filling the line.

(Extracted from the former DotMatrix PR)

### DIFF
--- a/docs/functions/util.rst
+++ b/docs/functions/util.rst
@@ -26,6 +26,41 @@ These functions are designed to be used by themes.
 
    The returned string is a fully escaped terminal formatting sequence.
 
+.. function:: _lp_fill(left, right, [fillstring], [splitends]) -> var:lp_fill
+
+   Adds as much *fillstring* (e.g. spaces) between *left* and *right*,
+   so as to make the resulting string the same width as the current terminal.
+
+   If *fillstring* is omitted, it defaults to one space.
+
+   If *fillstring* is a string with several characters
+   and *splitends* is 1 (the default),
+   then the final occurrence of *fillstring*
+   will have its end cut, so as to fit the terminal width.
+
+   If *fillstring* has multiple characters and *splitends* is 0,
+   some spaces will be inserted after the last occurrence of *fillstring*,
+   so as to match the exact width of the terminal.
+
+   *fillstring* may contains escaped sequences (such as colors).
+
+   .. note:: If *fillstring* have multiple characters and *splitends* is 1,
+             then **the last occurrence** of *fillstring*
+             is stripped of any escaped sequences
+             and only its printable characters are inserted.
+
+   If the available number of columns in the terminal is smaller than
+   the width of *left* and *right* combined, then
+   the function will return code 1 and set *lp_fill* to
+   *left* and *right*, concatenated.
+
+   For example, ``_lp_fill "Left part·" "·right part" "⣀⠔⠉⠢" 1`` will render
+   (in a terminal being 32 characters large):
+
+      Left part·⣀⠔⠉⠢⣀⠔⠉⠢⣀⠔⠉·right part
+
+   .. versionadded:: 2.2
+
 .. function:: _lp_grep_fields(filename, delimiter, keys...) -> var:lp_grep_fields
 
    Parse the given filename for one key/value pairs of the form
--- a/liquidprompt
+++ b/liquidprompt
@@ -992,6 +992,68 @@ _lp_grep_fields() {
     done <"$file"
 }
 
+# Adds as much $3 character (e.g. spaces) between $1 and $2,
+# so as to make the resulting string the same width as the current terminal.
+_lp_fill() {
+
+    local left="$1" right="$2" fillchars="${3:-" "}" splitends="${4:-1}"
+
+    # Compute the gap between left and right.
+    local ret
+    __lp_strip_escapes "$left"
+    local left_as_text="$ret"
+    __lp_strip_escapes "$right"
+    local right_as_text="$ret"
+    local gap_width=$((${COLUMNS:-80}-${#left_as_text}-${#right_as_text}))
+    if [[ $gap_width -lt 0 ]] ; then
+        lp_fill="${left}${right}"
+        return 1
+    fi
+
+    local filled=""
+    __lp_strip_escapes "$fillchars"
+    local fillchars_as_text="${ret}"
+    local fillchars_as_text_width=${#fillchars_as_text}
+    local nb_fillchars=$((gap_width/fillchars_as_text_width))
+    local i
+    for (( i=0; i < nb_fillchars; i++ )) ; do
+        filled+="$fillchars"
+    done
+
+    __lp_strip_escapes "$filled"
+    local actual_width=${#ret}
+    # If there is still a gap (i.e. we have an unaligned multi-character fillchars).
+    if [[ ${actual_width} -lt ${gap_width} ]] ; then
+        # User asked for splitends.
+        if [[ $splitends -ne 0 ]] ; then
+            # The last occurence is necessarily stripped from escaped sequences,
+            # or else we may have dangling sequences.
+            for (( i=_LP_FIRST_INDEX; i < $((${#fillchars_as_text}+_LP_FIRST_INDEX)); i++ )) ; do
+                # Get one single character.
+                if (( _LP_SHELL_zsh )) ; then
+                    filled+="${fillchars_as_text[i,i]}"
+                else
+                    filled+="${fillchars_as_text:i:1}"
+                fi
+                __lp_strip_escapes "$filled"
+                actual_width=${#ret}
+                if [[ ${actual_width} -ge ${gap_width} ]] ; then
+                    # Stop at this char if we're full.
+                    break
+                fi
+            done
+        else # User asked for no splitends.
+            for (( i=actual_width; i < gap_width; i++ )) ; do
+                # Fill with spaces.
+                filled+=" "
+            done
+        fi
+    fi
+
+    # shellcheck disable=SC2034
+    lp_fill="${left}${filled}${right}"
+}
+
 
 ##########################
 # Working Directory Path #
--- a/tests/test_utils.sh
+++ b/tests/test_utils.sh
@@ -694,6 +694,70 @@ function test_path_format_last_dir() {
   assertEquals "full directory formatting with multichar separator" "{l}pathname" "$lp_path_format"
 }
 
+function test_lp_fill {
+    local lp_fill
+
+    COLUMNS=80
+    _lp_fill "Left" "Right" " "
+    assertEquals "full width" 80 ${#lp_fill}
+
+    COLUMNS=3
+    _lp_fill "L" "R" "-"
+    assertEquals "single fill" "L-R" "$lp_fill"
+
+    _lp_fill "L" "R" ""
+    assertEquals "defaults to space" "L R" "$lp_fill"
+
+    COLUMNS=5
+    _lp_fill "L" "R" "~"
+    assertEquals "simple fill 5" "L~~~R" "$lp_fill"
+
+    COLUMNS=6
+    _lp_fill "L" "R" "+-"
+    assertEquals "multi-fill 6" "L+-+-R" "$lp_fill"
+
+    _lp_fill "L" "R" "123" 1
+    assertEquals "multi-fill 6 split" "L1231R" "$lp_fill"
+
+    _lp_fill "L" "R" "${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123" 0
+    assertEquals "multi-fill 6 with escape and no split" "L${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123 R" "$lp_fill"
+
+    _lp_fill "L" "R" "${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123" 1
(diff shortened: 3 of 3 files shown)


liquidprompt
```diff
@@ -929,6 +929,42 @@ __lp_theme_zsh_complete() {
     _describe 'theme' lp_theme_list
 }
 
+# Adds as much $3 character (e.g. spaces) between $1 and $2,
+# so as to make the resulting string the same width than the current terminal.
```
Rycieos, line 933:
```suggestion
# so as to make the resulting string the same width as the current terminal.
```

liquidprompt
```diff
@@ -929,6 +929,42 @@ __lp_theme_zsh_complete() {
...
+_lp_fill() {
+
+    local left="$1" right="$2" space="${3:-" "}"
+
+    if [[ ${#space} -gt 1 ]] ; then
+        # Only keep the first character.
+        space="${space:0:1}"
+    fi
+
+    # Merge side in header
```
Rycieos, line 943: I don't understand this comment. Did it refer to something specific in the dotmatrix theme?

tests/test_utils.sh
```diff
@@ -694,6 +694,34 @@ function test_path_format_last_dir() {
   assertEquals "full directory formatting with multichar separator" "{l}pathname" "$lp_path_format"
 }
 
+function test_lp_fill {
```
Rycieos, line 697: Should include a test with some Unicode characters, as well as some multi-byte Unicode characters and some extra wide ones.

I'm willing to bet that the extra wide ones will not work correctly. Not sure how to fix that, maybe it's not possible.

docs/functions/util.rst
```diff
@@ -26,6 +26,18 @@ These functions are designed to be used by themes.
...
    The returned string is a fully escaped terminal formatting sequence.
 
+.. function:: _lp_fill(left, right, [character]) -> var:lp_fill
+
+   Adds as much *character* (e.g. spaces) between *left* and *right*,
+   so as to make the resulting string the same width as the current terminal.
+
+   If `character` is ommitted, it defaults to space.
+
+   If `character` is a string with several characters,
```
Rycieos, line 36: Consistency:
```suggestion
   If *character* is omitted, it defaults to space.

   If *character* is a string with several characters,
```

Rycieos (changes requested):
Obviously the tests need to pass.

liquidprompt
nojhan: Yes, some rest of former vocabulary before I switched to the (obviously) simple left/right.

tests/test_utils.sh
nojhan: On my terminal emulator, everything is "fine": wide characters are just overlapping each others and hiding the beginning of the next one.

If we allow for multiple filling characters, this can be mitigated, but a user would have to add extra space to the right part by herself and the spacing may be randomly inconsistent. I'll try that.

Rycieos: Ah, I didn't think of Unicode characters as the fill string. I meant for the left or right strings, since we need to calculate their length. But the length we want is the printing size, not how many characters they have.

nojhan: How do we have access to this information? Shells only returns the number of characters when we ask for `${#string}`. I would find it easier to just let the user fill in the necessary blank spaces after a large unicode character (which is now possible, thanks to the multi-character filling support).

nojhan:
I finally setup for extending the function to multi-characters filling with colors.

Note that it seems the test framework does strip escaped sequence, so I'm not sure how to test.

tests/test_utils.sh
Rycieos: That is the $1M question. This SO post asks the same: https://unix.stackexchange.com/q/245013/421569

The answer is that there is not a great way to do it. Our options are:
1. Use `wc -L`: `length=$(printf '%s' "${foo}" | wc -L)`. Being a subshell, this will be very slow. Also, not guaranteed to be accurate. 
2. Print the string to the terminal, and record the position before and after. This is the most likely to be accurate. Likely would have a few issues:
    * Reading from the terminal emulator can be buggy, and might hang and lock up.
    * Printing text to the terminal is not ideal. If we reset the cursor position after, when PS1 is printed it should write over these characters, meaning we don't need to worry about it.
    * If *anything* else printed to the terminal at the same time we did, it could give a terribly wrong answer.
    * Portability issues?
3. Use an `awk` script. Doesn't require a subshell, but does require a binary call.
4. Just don't. Multi-width or zero width characters are somewhat rare, so take the easy route and ignore the problem.

I'm not sure what the right answer is, but probably we should try a few of them and see how well they work.

Note that this is something I've been thinking about for a while, as it's needed to solve #599 as well.

liquidprompt
```diff
@@ -991,6 +991,71 @@ _lp_grep_fields() {
...
+    local left_as_text="$ret"
+    __lp_strip_escapes "$right"
+    local right_as_text="$ret"
+    local gap_width=$((${COLUMNS:-80}-${#left_as_text}-${#right_as_text}))
+    if [[ $gap_width -lt 0 ]] ; then
+        # FIXME this would be larger than the actual width.
+        # Maybe think of an option to reduce either the right or left side?
+        # Or insert a newline?
+        lp_fill="${left}${right}"
+        return
```
Rycieos, line 1012: I have a solution: `return 1` and leave it up to the theme to decide what to do.

The theme can then either:
* Throw away the return string, and try to shorten one side or the other and try again, or
* Ignore the return value, and use the return string even though it is too long.

tests/test_utils.sh
```diff
@@ -694,6 +694,72 @@ function test_path_format_last_dir() {
...
+    local lp_fill
+    echo $COLUMNS
+    _lp_fill "Left" "Right" " "
+    assertEquals "full width" ${COLUMNS} ${#lp_fill}
```
Rycieos, line 701: This test is failing in the runners because `$COLUMNS` is not set in environments without a terminal.

Just set `$COLUMNS` to something sane, like 60.

Rycieos (changes requested):

tests/test_utils.sh
nojhan: I would go for 4, given that it's almost just a templating issue, at the end. If you configure your prompt to use long chars, you will see the results pretty easily. And on all the terminal emulator I have on my box, long characters are just overwritten on top of each others, so that I just have to add spaces, if I want to match their width. The only issue remaining would actually be the very end of line, indeed, but that's just *one* (long) character, and probably a very rare case.

tests/test_utils.sh
```diff
@@ -694,6 +694,73 @@ function test_path_format_last_dir() {
...
+
+    _lp_fill "Le" "Ri" "${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123" 0
+    __lp_strip_escapes "$lp_fill"
+    # assertEquals "multi-fill 11 with escape and no split" "Le${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123123 Ri" "$ret"
+    assertEquals "multi-fill 11 with escape and no split" "Le123123 Ri" "$ret"
+
+    _lp_fill "Le" "Ri" "${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123" 1
+    __lp_strip_escapes "$lp_fill"
+    # assertEquals "multi-fill 11 with escape and split" "Le${_LP_OPEN_ESC}${_LP_CLOSE_ESC}1231231Ri" "$ret"
+    assertEquals "multi-fill 11 with escape and split" "Le1231231Ri" "$ret"
```
Rycieos, line 757: The unit testing library does not strip escape sequences. This didn't work correctly because your code does not modify the fill strings that it doesn't need to split. The correct expected strings according to the docs would be 
```
"Le${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123 Ri"
```
and
```
"Le${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123${_LP_OPEN_ESC}${_LP_CLOSE_ESC}1231Ri"
```

tests/test_utils.sh
```diff
@@ -694,6 +694,73 @@ function test_path_format_last_dir() {
...
+
+    COLUMNS=32
+    _lp_fill "Left part·" "·right part" "⣀⠔⠉⠢" 1
+    assertEquals "beautiful fill" "Left part·⣀⠔⠉⠢⣀⠔⠉⠢⣀⠔⠉·right part" "$lp_fill"
```
Rycieos, line 761: This test (and only this test) is failing on Windows. I am not sure why...

Rycieos (changes requested):

Rycieos:
If you could rebase onto the latest `master`, the docs linting work will validate your changes as well.

tests/test_utils.sh
nojhan: FTR:
```
shunit2:ERROR test_lp_fill() returned non-zero return code.
ASSERT:beautiful fill expected:<Left part·⣀⠔⠉⠢⣀⠔⠉⠢⣀⠔⠉·right part> but was:<Left part·⣀⠔⠉·right part>
```

I can't easily debug, not having Windows at hand, but this is strange indeed.

Rycieos: I do. I have never run the tests locally on Windows before, I always let Github Actions run them. Wow, they are so slow.

That test passes for me. This makes me think it is an encoding issue on the runner. I can't find any docs that suggest that the encoding of the Windows image is not UTF-8, but I don't know what else it would be.

I see a few possible solutions:
* Try to force bash to UTF-8 encoding.
* Try to force the runner image to UTF-8 encoding.
* Give up and tell the tests to skip this test if it is on a Windows VM. Put this block before the test:
```bash
# The Windows runners have issues with these Unicode characters.
if [[ ${RUNNER_OS-} == "Windows" ]]; then
  startSkipping
fi
```

nojhan: I let you decide, I have absolutely no clue, and I did not touched such things under Windows for literally decades.

nojhan:
I fixed the typos detected by the new doc linter and rebased.

nojhan:
I'm wondering whether I should add some config variable like `LP_MARK_FILL` to configure the default of `_lp_fill`, so that any basic use of the function may be configurable. Or should we prefer each use to have its own configuration variable? Both?

docs/functions/util.rst
```diff
@@ -26,6 +26,41 @@ These functions are designed to be used by themes.
...
+   so as to make the resulting string the same width as the current terminal.
+
+   If *fillstring* is omitted, it defaults to one space.
+
+   If *fillstring* is a string with several characters
+   and *splitends* is 1 (the default),
+   then the final occurrence of *fillstring*
+   will have its end cut, so as to fit the terminal width.
+
+   If *fillstring* have multiple characters and *splitends* is 0,
```
Rycieos, line 41:
```suggestion
   If *fillstring* has multiple characters and *splitends* is 0,
```

tests/test_utils.sh
Rycieos: Go for the last option, it's not worth the headache.

I would add those 4 lines on line 749.

Rycieos (changes requested):
> I'm wondering whether I should add some config variable like LP_MARK_FILL to configure the default of _lp_fill, so that any basic use of the function may be configurable. Or should we prefer each use to have its own configuration variable? Both?

I would say no global config variable. Some themes will want multi-character fills, and some might want fills on multiple lines, which might want different options anyway.

Fix these two issues and rebase, and I'll merge.

tests/test_utils.sh
```diff
@@ -694,6 +694,67 @@ function test_path_format_last_dir() {
...
+
+    _lp_fill "Le" "Ri" "${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123" 1
+    assertEquals "multi-fill 11 with escape and split" "Le${_LP_OPEN_ESC}${_LP_CLOSE_ESC}123${_LP_OPEN_ESC}${_LP_CLOSE_ESC}1231Ri" "$lp_fill"
+
+    # The Windows runners have issues with these Unicode characters.
+    if [[ ${RUNNER_OS-} == "Windows" ]]; then
+        COLUMNS=32
+        _lp_fill "Left part·" "·right part" "⣀⠔⠉⠢" 1
+        assertEquals "beautiful fill" "Left part·⣀⠔⠉⠢⣀⠔⠉⠢⣀⠔⠉·right part" "$lp_fill"
+    fi
```
Rycieos, line 755: This makes the tests run _only_ on Windows, not the other way around. Change to this:
```suggestion
    # The Windows runners have issues with these Unicode characters.
    # Skip all following tests.
    if [[ ${RUNNER_OS-} == "Windows" ]]; then
      startSkipping
    fi

    COLUMNS=32
    _lp_fill "Left part·" "·right part" "⣀⠔⠉⠢" 1
    assertEquals "beautiful fill" "Left part·⣀⠔⠉⠢⣀⠔⠉⠢⣀⠔⠉·right part" "$lp_fill"
```

Rycieos (changes requested):
Test is still failing on the Windows runners, looks like I didn't explain what I wanted well enough.

tests/test_utils.sh
nojhan: :facepalm: I should not have done that in a hurry with the kids running around asking questions… sorry for that.

PR #744, opened 2022-10-13
fix(tools): do not indent theme preview

nojhan:
Do not indent theme previews, or else the first line of multi-line prompt will not be aligned with the following ones.


Rycieos (approved):
I don't see a good way to indent following lines, so yeah, I guess this is the right way to fix this.

Issue #729, opened 2022-08-16
Give names/icons to standard exit codes


nojhan:
Display a more readable name along with/instead of the exit code, when it is a standard one.

### Description

Standard exit codes are best known through their names, optionally display them next to the error code.
Optionally shorten those names as single characters/icons.

See: https://stackoverflow.com/questions/1101957/are-there-any-standard-exit-status-codes-in-linux

### How will this be useful?

Better understanding of what's going on by the user.

Rycieos:
Interesting idea. I would find this useful.

I would be cautious though about attributing specific codes to specific strings. As many of the linked answers and comments have pointed out, there is no standard around error codes, only some repeated patterns.

The one thing that is for sure the same between Bash and Zsh is that they will show the exit code of a process killed by a signal as 128 + the signal code (probably because the kernel sets the exit code). However, you can make any program return that exit code as well (`bash -c 'exit 135'`, would look like `SIGBUS`), which could be quite confusing.

We would also need to look into how portable these signals are, and if all OSes set the exit code on exit from a signal.

I would at least be open to displaying the 31 basic signals if they turn out to be portable. Other codes, I would need to see some patterns shown in common programs. I would rather show no information for a code than the wrong one.

nojhan:
Done in #745

PR #745, opened 2022-10-16
Feat+error meaning

nojhan:
Basically displays the (guessed) meaning of the last error code (feature request #729).

Separated in two commits, so that you may allow a more or less granular level of details.

First commit: adds the core feature.
- Use a set of reasonably common error codes on POSIX systems
  (but is LP used out of a POSIX env?).
- Use the color than lp_error (namely LP_COLOR_ERR).
- Use a separated LP_ENABLE_ERROR_MEANING.
- Put in the default theme right after the error code.
- Code would be very much simpler/configurable with Bash 4's associative arrays,
  but here use a case switch with non-configurable error meanings.

Second commit: let the user decide the level of guess for error meaning.
- Adds LP_ENABLE_ERROR_MEANING_EXTENDED to allow for a larger set of error codes.


docs/config.rst
```diff
@@ -415,6 +415,37 @@ Features
 
    .. versionadded:: 2.0
 
+.. attribute:: LP_ENABLE_ERROR_MEANING
```
Rycieos, line 418: This config variable is never checked, so the feature is always enabled.

docs/config.rst
```diff
@@ -415,6 +415,37 @@ Features
...
+
+   Extends the set of interpreted error codes to a larger set of (POSIX) codes.
+
+   .. note:: This use a reasonable set of error codes
+             that are common on POSIX systems on x86 or ARM architectures
+             (most notably from ``sysexit.h`` and ``signal.h``).
+             But any software may use its own set of codes,
+             and thus the guess may be wrong.
+
+   See also: :attr:`LP_COLOR_ERR`.
```
Rycieos, line 445: This need a mention that it will have no effect if `LP_ENABLE_ERROR_MEANING` is disabled.

liquidprompt
```diff
@@ -1258,6 +1261,200 @@ _lp_error_color() {
     lp_error_color="${LP_COLOR_ERR}${lp_error}${NO_COL}"
 }
 
+_lp_error_meaning() {
+    # Reasonable codes for a POSIX env.
+
+    _lp_error || return "$?"
```
Rycieos, line 1267: The issue with checking for the error code in this way is that if `LP_ENABLE_ERROR` is false, this function will be disabled as well. I think it makes sense to allow users to enable error meanings without enabling the printing of the raw code.

I would change this check to this:
```suggestion
    (( LP_ENABLE_ERROR_MEANING )) || return 2
    (( lp_error != 0 )) || return 1
```

Rycieos (changes requested):
Looks good in general, just a few issues.

nojhan:
Do I squash all commits into one ? (i.e. do you want to keep the extended error meaning setup?)

Rycieos (changes requested):
Looks good now.

I missed that question the first time, sorry. I do not have much of an opinion. Clearly you intended for both bits to be part of the feature, so it would be fine in once commit. But if you want it as separate, I'm fine with that.

But no matter what, squash away the extra commits from the fixes and docs.

nojhan:
The two commits were only there to make it easy to just have a single config, if you're OK with the "extended" config, I'll squash everything.

liquidprompt
```diff
@@ -1258,6 +1261,202 @@ _lp_error_color() {
     lp_error_color="${LP_COLOR_ERR}${lp_error}${NO_COL}"
 }
 
+_lp_error_meaning() {
+    # Reasonable codes for a POSIX env.
+
+    (( LP_ENABLE_ERROR_MEANING )) || return 2
+
+    (( _lp_error != 0 )) || return 1
```
Rycieos, line 1269: > ./liquidprompt: line 1331: _lp_error: unbound variable

```suggestion
    (( lp_error != 0 )) || return 1
```

Rycieos (changes requested):
Yeah, squash down to 1.

Rycieos (approved):

PR #746, opened 2022-10-16
feat(theme): adds the unfold theme

nojhan:
The Unfold theme, very similar to the default theme, but spread on two lines and a right-aligned section.


nojhan:
This depends on PR #738 (for having `lp_fill` to right-align).

nojhan:
I should probably consider adding a `LP_UNFOLD_MARK_FILL="·"` config. Or would it be too much? The right alignment with spaces can be difficult to parse visually on large terminals.

themes/unfold/unfold.theme
```diff
@@ -0,0 +1,44 @@
...
+    _lp_default_theme_directory
+}
+
+_lp_unfold_theme_prompt() {
+    _lp_default_theme_prompt_data
+
+    if [[ -f "${LP_PS1_FILE-}" ]]; then
+        # shellcheck source=liquid.ps1
+        source "$LP_PS1_FILE"
+    fi
```
Rycieos, line 16: Why support this? If a user has this set, this theme is no different from the default theme. It might even be confusing for a user to enable this theme and see nothing change.

Rycieos (changes requested):
Nice and simple, I like it.

Clearly, this could be a template instead, aka a version 1 theme. Any thoughts on whether this needs to be a (new type) theme, or could just be a template file?

themes/unfold/unfold.theme
nojhan: For both questions: I don't really know. I just used the "default" theme functions and build up from there.

Rycieos: Fair enough 😄 

For this question, I say remove this code. No need to support old style themes in this new thing.

For the other question, a new style theme is probably the right choice. So I guess forget I asked.

nojhan: ✅ done.

themes/unfold/unfold.theme
```diff
@@ -0,0 +1,39 @@
...
+}
+
+_lp_unfold_theme_directory() {
+    _lp_default_theme_directory
+}
+
+_lp_unfold_theme_prompt() {
+    _lp_default_theme_prompt_data
+
+    if [[ -z "${LP_PS1-}" ]]; then
```
Rycieos, line 13: Get rid of this check as well. Best case it will never happen, since we are not importing a template file to set it, worst case the user switched from the default theme, `$LP_PS1` is still set, and their prompt will never change.

The `else` block below can disappear then obviously.

Rycieos (changes requested):
Looks really good.

One thing I would want: a way to dynamically switch between the default theme and this theme when the prompt gets long. That should probably be its own theme, and maybe I will make that.

nojhan:
Automatic theme switching may also be a fix for issue #599.

Rycieos (changes requested):
One last thing I forgot: the new theme file needs to be added to the `shellcheck.sh` script.

Rycieos (approved):
Looks good! I'm glad we have a theme that uses the new `_lp_fill()` function.

PR #748, opened 2022-10-22
fix(mux): typo in returned variable of _lp_multiplexer

nojhan:
Just two characters changed.


Rycieos (changes requested):
The variable is also typoed in the docs, so it needs updating there as well.

Note you will want to rebase, as the failed docs-lint was caused by something else, which I just pushed a commit to fix.

Rycieos (approved):

PR #750, opened 2022-10-24
Doc+disabled

nojhan:
- Adds a list of features that are disabled by default.
- Fix comments about competitors.


Rycieos (changes requested):
Single lint issue.

Rycieos (approved):

Issue #722, opened 2022-08-16
Generic environment variables tags


nojhan:
Display a set of tags showing whether some environment variables are set or not.

### Description

Configuration would be an associative array of environment variable name => character.
By default, if the environment variable is set, the character is displayed.
Optionally, characters are always displayed, but with different colors if the variable is set or not.

Rycieos:
Interesting idea. Could cut down on tool support requests, assuming the tools expose things as environment variables.

One catch is that Associative arrays were added to Bash in 4.0. As of right now, all Liquidprompt features are supported Bash 3.2+ and Zsh 5.0+. Thoughts on having some features that have a stricter set of requirements?

nojhan:
Technically, an associative array is just an array of pair with fancy access. So we can configure this with an array of strings figuring a pair:
```bash
LP_ENV_TAGS=(
    "HTTP_PROXY ↥" # We may refactor proxy as just an entry here.
    "VERBOSE ⋁"   # May be multiple characters.
    "EDITOR %"     # '%' can be a placeholder for rendering the actual content of the variable.
    "DESKTOP_SESSION ∢"
)
LP_ENV_MARK_OPEN="{"
LP_ENV_SEP=" "
LP_ENV_MARK_CLOSE="}"
```
Would be rendered as: "{↥ ⋁ vi ∢}"

nojhan:
FTR, I did not refactored the proxy in the generic code, as I felt it may considered as important by the user, and thus displayed apart from the section built by the generic code.

Rycieos:
I agree, we should leave it, but add a section to the proxy config docs explaining how you can add a line to the env vars config to do the same thing.

nojhan:
I realized it is not really possible to implement the proxy management with the env var section, because the proxy feature does in fact manage all the possible env var (lower/upper case, SSL, "all"). However, the env var section can only manage one variable at a time.

Rycieos:
Ah, that is a fair point. Good to note that _usually_ they are lowercase (https://unix.stackexchange.com/a/212972/421569), but not always, and there are multiple that we want to check. So yes, probably best to not deprecate the proxy feature.

That will be an unusual case, since usually environment vars are either defined by POSIX, or defined by a specific program/project, and will not have different forms.

nojhan:
Done in #754

PR #754, opened 2022-11-02
feat(env_vars): adds a section to watch user-defined environment variables

nojhan:
- Implements feature request #722.
- Incidentally adds test_proxy, which was missing.

Example:
```
LP_ENV_VARS=("VERBOSE V" "EDITOR ed:%s" "DESKTOP_SESSION %s nos")
08:54:10 [:~/code/liquidprompt](ed:kak notion) feat+env+* ± 
```

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -447,6 +447,25 @@ Features
 
    .. versionadded:: 2.2
 
+.. attribute:: LP_ENABLE_ENV_VARS
+   :type: bool
+   :value: 1
+
+   Display a user-defined set of environment variables.
+   May show if the variables are unset, set, or their actual content.
+
+   Watched variables should be added to the :attr:`LP_ENV_VARS` array.
+
+   The resulting prompt section is configured by:
+
+   - :attr:`LP_MARK_ENV_VARS_OPEN`
+   - :attr:`LP_MARK_ENV_VARS_SEP`
+   - :attr:`LP_MARK_ENV_VARS_CLOSE`
+   - :attr:`LP_COLOR_ENV_VARS_SET`
+   - :attr:`LP_COLOR_ENV_VARS_UNSET`
+
+   .. versionadded:: 2.2
+
 .. attribute:: LP_ENABLE_FOSSIL
    :type: bool
    :value: 1
@@ -773,6 +792,43 @@ Features
 
    .. versionadded:: 2.1
 
+.. attribute:: LP_ENV_VARS
+   :type: array<string>
+   :value: ()
+
+   The set of environment variables that the user wants to watch.
+
+   Items should be a string with three space-separated elements
+   of the form `"<name> <set>[ <unset>]"`, containing:
+
+   - the name of the variable to watch,
+   - the string to display if the variable is set,
+   - (optionally) the string to display if the variable is not set.
+
+   The string used when the variable is set may contain the ``%s`` mark,
+   which is replaced by the actual content of the variable.
+
+   For example::
+
+    LP_ENV_VARS=(
+        # Display "V" if VERBOSE is set, nothing if it's unset.
+        "VERBOSE V"
+        # Display the name of the desktop session, if set, T if unset.
+        "DESKTOP_SESSION %s T"
+        # Display "ed:" followed the name of the default editor, nothing if unset.
+        "EDITOR ed:%s"
+    )
+
+   See also :attr:`LP_ENABLE_ENV_VARS`.
+
+   The resulting prompt section is configured by:
+
+   -  :attr:`LP_MARK_ENV_VARS_OPEN`
+   -  :attr:`LP_MARK_ENV_VARS_SEP`
+   -  :attr:`LP_MARK_ENV_VARS_CLOSE`
+   -  :attr:`LP_COLOR_ENV_VARS_SET`
+   -  :attr:`LP_COLOR_ENV_VARS_UNSET`
+
 .. attribute:: LP_HG_COMMAND
    :type: string
    :value: "hg"
@@ -1050,6 +1106,58 @@ Marks
 
    Mark used instead of :attr:`LP_MARK_DEFAULT` to indicate that the current
    directory is disabled for VCS display through :attr:`LP_DISABLED_VCS_PATHS`.
+
+.. attribute:: LP_MARK_ENV_VARS_OPEN
+   :type: string
+   :value: "("
+
+   Mark used to start the user-defined environment variables watch list.
+
+   See also:
+
+   - :attr:`LP_ENABLE_ENV_VARS`
+   - :attr:`LP_ENV_VARS`
+   - :attr:`LP_MARK_ENV_VARS_SEP`
+   - :attr:`LP_MARK_ENV_VARS_CLOSE`
+   - :attr:`LP_COLOR_ENV_VARS_SET`
+   - :attr:`LP_COLOR_ENV_VARS_UNSET`
+
+   .. versionadded:: 2.2
+
+.. attribute:: LP_MARK_ENV_VARS_SEP
+   :type: string
+   :value: " "
+
+   Mark used to separate items of the user-defined
+   environment variables watch list.
+
+   See also:
+
+   - :attr:`LP_ENABLE_ENV_VARS`
+   - :attr:`LP_ENV_VARS`
+   - :attr:`LP_MARK_ENV_VARS_OPEN`
+   - :attr:`LP_MARK_ENV_VARS_CLOSE`
+   - :attr:`LP_COLOR_ENV_VARS_SET`
+   - :attr:`LP_COLOR_ENV_VARS_UNSET`
+
+   .. versionadded:: 2.2
+
+.. attribute:: LP_MARK_ENV_VARS_CLOSE
+   :type: string
+   :value: ")"
+
+   Mark used to end the user-defined environment variables watch list.
+
+   See also:
+
+   - :attr:`LP_ENABLE_ENV_VARS`
+   - :attr:`LP_ENV_VARS`
+   - :attr:`LP_MARK_ENV_VARS_OPEN`
+   - :attr:`LP_MARK_ENV_VARS_SEP`
+   - :attr:`LP_COLOR_ENV_VARS_SET`
+   - :attr:`LP_COLOR_ENV_VARS_UNSET`
+
+   .. versionadded:: 2.2
 
 .. attribute:: LP_MARK_FOSSIL
    :type: string
@@ -1416,6 +1524,42 @@ Valid preset color variables are:
    Color used to indicate the last command exited with a non-zero return code.
 
    See also: :attr:`LP_ENABLE_ERROR`.
+
+.. attribute:: LP_COLOR_ENV_VARS_SET
+   :type: string
+   :value: $BOLD_BLUE
+
+   Color of the environment variables that are set,
+   in the user-defined watch list.
+
+   See also:
+
+   - :attr:`LP_ENABLE_ENV_VARS`
+   - :attr:`LP_ENV_VARS`
+   - :attr:`LP_COLOR_ENV_VARS_UNSET`
+   - :attr:`LP_MARK_ENV_VARS_OPEN`
+   - :attr:`LP_MARK_ENV_VARS_SEP`
+   - :attr:`LP_MARK_ENV_VARS_CLOSE`
+
+   .. versionadded:: 2.2
+
+.. attribute:: LP_COLOR_ENV_VARS_UNSET
+   :type: string
+   :value: $BLUE
+
+   Color of the environment variables that are unset,
+   in the user-defined watch list.
+
+   See also:
+
+   - :attr:`LP_ENABLE_ENV_VARS`
+   - :attr:`LP_ENV_VARS`
+   - :attr:`LP_COLOR_ENV_VARS_SET`
+   - :attr:`LP_MARK_ENV_VARS_OPEN`
+   - :attr:`LP_MARK_ENV_VARS_SEP`
+   - :attr:`LP_MARK_ENV_VARS_CLOSE`
+
+   .. versionadded:: 2.2
 
 .. attribute:: LP_COLOR_HOST
    :type: string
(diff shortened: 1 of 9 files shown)


nojhan:
The `LP_ENABLE_ENV_VARS` entry in the doc is not well formatted, but I can't understand why.

docs/config.rst
```diff
@@ -1417,6 +1525,42 @@ Valid preset color variables are:
...
+   - :attr:`LP_MARK_ENV_VARS_SEP`
+   - :attr:`LP_MARK_ENV_VARS_CLOSE`
+
+   .. versionadded:: 2.2
+
+.. attribute:: LP_COLOR_ENV_VARS_UNSET
+   :type: string
+   :value: $BLUE
+
+   Color of the environment variables that are set,
```
Rycieos, line 1550:
```suggestion
   Color of the environment variables that are unset,
```

liquidprompt
```diff
@@ -1550,6 +1556,70 @@ _lp_http_proxy_color() {
...
+    local var_rep var evar fmt def
+
+    lp_env_vars=()
+    # For all user-defined setup.
+    for var_rep in ${LP_ENV_VARS[@]+"${LP_ENV_VARS[@]}"}; do
+        IFS=' ' read -r var fmt def <<<"$var_rep"
+        # Variable name and format has to be set, but default is optional.
+        if [[ -n "${var}" && -n "${fmt}" ]]; then
+            # A way to test for undefined ${!var}
+            if declare -xp "${var}" &>/dev/null; then
```
Rycieos, line 1583: A better way to test this (that we normally use) that doesn't require a redirect is `[[ -n ${var+x} ]]`, which returns true if the variable is set. It is also much faster:

```bash
$ test() { for i in {1..100000}; do [[ -n ${var+x} ]]; done }
$ time test
real   	0m0.257s
user   	0m0.247s
sys    	0m0.008s

$ test() { for i in {1..100000}; do declare -xp "${var}" &>/dev/null; done }
$ time test
real   	0m1.746s
user   	0m1.083s
sys    	0m0.635s
```

liquidprompt
```diff
@@ -1550,6 +1556,70 @@ _lp_http_proxy_color() {
...
+                    # "Some Zsh specific parameter expansions like ${(q)value} trigger this warning,
+                    # but ShellCheck does not support Zsh."
+                    # shellcheck disable=SC2296
+                    evar="${(P)var}"
+                else
+                    # NOTE: indirection expansion are deprecated starting at bash 4.3, should use nameref.
+                    evar="${!var}"
+                fi
+                # Print tag/content or default.
+                if [[ -n "${evar}" && "${evar}" != "0" ]]; then
```
Rycieos, line 1596: Why skip this if the variable contents are `0`? Is that not a valid value that a user might want to see?

I see you have test coverage for it, but I need to know the reasoning, and so does the documentation.

liquidprompt
```diff
@@ -1550,6 +1556,70 @@ _lp_http_proxy_color() {
...
+                    lp_env_vars+=( "${color_set}${fmt/\%s/${evar}}${color_no}" )
```
Rycieos, line 1597: Probably should be nested quoted to be safe
```suggestion
                    lp_env_vars+=( "${color_set}${fmt/\%s/"${evar}"}${color_no}" )
```

liquidprompt
```diff
@@ -1550,6 +1556,70 @@ _lp_http_proxy_color() {
...
+                elif [[ -n "$def" ]]; then
+                    lp_env_vars+=( "${color_unset}${def}${color_no}" )
+                fi
```
Rycieos, line 1600: This will treat set but empty variables as "unset". I think that it is not correct behavior.

I see you have test coverage for it, but I need to know the reasoning, and so does the documentation.

liquidprompt
```diff
@@ -1550,6 +1556,70 @@ _lp_http_proxy_color() {
...
+        return 1
+    fi
+}
+
+_lp_env_vars_color() {
+    if _lp_env_vars "${LP_COLOR_ENV_VARS_SET}" "${LP_COLOR_ENV_VARS_UNSET}"; then
+        _lp_join "${LP_MARK_ENV_VARS_SEP}" "${lp_env_vars[@]}"
+        lp_env_vars_color="${LP_MARK_ENV_VARS_OPEN}${lp_join}${LP_MARK_ENV_VARS_CLOSE}"
+    else
+        return $?
```
Rycieos, line 1618: Needs quoting
```suggestion
        return "$?"
```

liquidprompt
```diff
@@ -1550,6 +1556,70 @@ _lp_http_proxy_color() {
...
+                # Expands the underlying variable name.
+                if (( _LP_SHELL_zsh )); then
+                    # From https://www.shellcheck.net/wiki/SC2296 :
+                    # "Some Zsh specific parameter expansions like ${(q)value} trigger this warning,
+                    # but ShellCheck does not support Zsh."
+                    # shellcheck disable=SC2296
+                    evar="${(P)var}"
+                else
+                    # NOTE: indirection expansion are deprecated starting at bash 4.3, should use nameref.
+                    evar="${!var}"
```
Rycieos, line 1593: Could we save some time by checking if the format string includes `%s` before we try to get the variable contents? It might make the logic a lot more complicated, so might not be worth it.

liquidprompt
```diff
@@ -1550,6 +1556,70 @@ _lp_http_proxy_color() {
...
+    if [[ -z "$color_set" && -z "$color_unset" ]]; then
+        color_no=""
+    fi
+
+    local var_rep var evar fmt def
+
+    lp_env_vars=()
+    # For all user-defined setup.
+    for var_rep in ${LP_ENV_VARS[@]+"${LP_ENV_VARS[@]}"}; do
+        IFS=' ' read -r var fmt def <<<"$var_rep"
```
Rycieos, line 1579: These variable names are confusing.
```suggestion
        IFS=' ' read -r var set_fmt unset_fmt <<<"$var_rep"
```

Rycieos (changes requested):
Looks good in general.

liquidprompt
nojhan: But this does not test if the indirected variable is set, it tests if the primary variable is set:
```sh
$ var="UNSET" ; if [[ -n ${var+x} ]]; then echo "SET"; else echo "UNSET"; fi
SET
$ var="UNSET" ; if declare -xp "${var}" &>/dev/null; then echo "SET"; else echo "UNSET"; fi
UNSET
```
And given that the *user* gives the names to watch, we need the second option. Bash >4.3 has nameref that would help, but before that, the only legit option seems to be indirection expansion, which does not work here.

liquidprompt
nojhan: Usually, if, say, `VERBOSE=0`, then the feature is OFF, and I would not want to see it. But maybe it would be better to show the zero, indeed. The user can still `unset`.

liquidprompt
Rycieos: Ah, true. But the indirection expansion does work for me:
```bash
 $ foo=bar
$ var=foo
$ printf '%s\n' "${!var+x}"
x
$ unset foo
$ printf '%s\n' "${!var+x}"


```
So `[[ -n ${!var+x} ]]` should work.

nojhan: Indeed. I would have bet I tried that…

liquidprompt
nojhan: I realized the test for unset does not allow access to the value anyway, so we need to do one more indirection expansion, and then it make sense to test for the subst mark. GG

liquidprompt
```diff
@@ -1550,6 +1556,78 @@ _lp_http_proxy_color() {
...
+        color_no=""
+    fi
+
+    local var_rep var evar fmt_if_set fmt_if_unset
+
+    lp_env_vars=()
+    # For all user-defined setup.
+    for var_rep in ${LP_ENV_VARS[@]+"${LP_ENV_VARS[@]}"}; do
+        IFS=' ' read -r var fmt_if_set fmt_if_unset <<<"$var_rep"
+        # Variable name and format has to be set, but default is optional.
```
Rycieos, line 1580: Default?
```suggestion
        # Variable name and set format has to be set, but unset format is optional.
```

liquidprompt
```diff
@@ -1550,6 +1556,78 @@ _lp_http_proxy_color() {
...
+                # "Some Zsh specific parameter expansions like ${(q)value} trigger this warning,
+                # but ShellCheck does not support Zsh."
+                # shellcheck disable=SC2296
+                var_is_set="${(P)var+__LP_IS_SET__}"
+            else
+                # NOTE: indirection expansion are deprecated starting at bash 4.3, should use nameref.
+                var_is_set="${!var+__LP_IS_SET__}"
+            fi
+
+            if [[ "$var_is_set" == "__LP_IS_SET__" ]]; then
```
Rycieos, line 1595: This seems overly complicated to me, why not
```suggestion
                var_is_set="${(P)var+x}"
            else
                # NOTE: indirection expansion are deprecated starting at bash 4.3, should use nameref.
                var_is_set="${!var+x}"
            fi

            if [[ -n $var_is_set ]]; then
```

nojhan:
I don't get why the tests show double quotes on MacOS…

liquidprompt
```diff
@@ -1550,6 +1556,78 @@ _lp_http_proxy_color() {
...
+                if [[ "${fmt_if_set}" == *"%s"* ]]; then
+                    local evar=
+                    if (( _LP_SHELL_zsh )); then
+                        # shellcheck disable=SC2296
+                        evar="${(P)var}"
+                    else
+                        evar="${!var}"
+                    fi
+                    # Print content.
+                    lp_env_vars+=( "${color_set}${fmt_if_set/\%s/"${evar}"}${color_no}" )
```
Rycieos, line 1605: I was wrong, Bash 3.2 evaluates the quotes as part of this.
```bash
$ fmt="foo%s"
$ evar=bar
$ printf '%s\n' "${fmt/\%s/"${evar}"}"
foo"bar"
```
```suggestion
                    lp_env_vars+=( "${color_set}${fmt_if_set/\%s/${evar}}${color_no}" )
```

docs/config.rst
```diff
@@ -447,6 +447,25 @@ Features
...
+
+   The resulting prompt section is configured by:
+
+   - :attr:`LP_MARK_ENV_VARS_OPEN`
+   - :attr:`LP_MARK_ENV_VARS_SEP`
+   - :attr:`LP_MARK_ENV_VARS_CLOSE`
+   - :attr:`LP_COLOR_ENV_VARS_SET`
+   - :attr:`LP_COLOR_ENV_VARS_UNSET`
+
+  .. versionadded:: 2.2
```
Rycieos, line 467: I bet this line is causing it. It has only 2 spaces of indentation.

Rycieos (changes requested):
The MacOS tests are very handing for catching Bash 3.2 bugs.

liquidprompt
nojhan: I find the syntax unreadable (as often with Bash), but at first I misunderstood (I believed we would have the content of var if it was set, hence using a unique placeholder). As least I would prefer to see `${!var+IS_SET}` which is more clear to me.

docs/config.rst
nojhan: :facepalm:

Rycieos: I think the linter missed it completely.

liquidprompt
```diff
@@ -1550,6 +1556,78 @@ _lp_http_proxy_color() {
     lp_http_proxy_color="${LP_COLOR_PROXY}${LP_MARK_PROXY}${NO_COL}"
 }
 
+_lp_env_vars() { # [color_set, [color_unset]]
+    # Expects LP_ENV_VARS to be an array containing items of the form:
+    # "<ENV_VAR_NAME> <string if set>[ <string if unset>]"
+    # Strings may be `%s`, which will be replaced by the variable's actual content.
+
+    (( LP_ENABLE_ENV_VARS )) || return 2
+    [[ -n "${LP_ENV_VARS-}" && ${#LP_ENV_VARS} -gt 0 ]] || return 1
```
Rycieos, line 1565: `$LP_ENV_VARS` needs a default value up top, like
```bash
LP_ENV_VARS=( ${LP_ENV_VARS[@]+"${LP_ENV_VARS[@]}"} )
```
Then you shouldn't need the `-n "${LP_ENV_VARS-}"` guard check.

liquidprompt
Rycieos: Fair enough. The big one I wanted is the `-n` check instead of `==`, which you have changed.

Rycieos (changes requested):

liquidprompt
```diff
@@ -1550,6 +1557,77 @@ _lp_http_proxy_color() {
     lp_http_proxy_color="${LP_COLOR_PROXY}${LP_MARK_PROXY}${NO_COL}"
 }
 
+_lp_env_vars() { # [color_set, [color_unset]]
```
Rycieos, line 1560: This function needs docs.

I think I've been forgetting to check that recently, you might want to do a pass on recently merged PRs...

liquidprompt
```diff
@@ -1550,6 +1557,77 @@ _lp_http_proxy_color() {
...
+                lp_env_vars+=( "${color_unset}${fmt_if_unset}${color_no}" )
+            fi
+        fi
+    done
+    if [[ -z "${lp_env_vars-}" ]]; then
+        return 1
+    fi
+}
+
+_lp_env_vars_color() {
```
Rycieos, line 1621: This function needs docs.

liquidprompt
```diff
@@ -3735,6 +3813,12 @@ _lp_default_theme_prompt_data() {
         LP_PROXY=
     fi
 
+    if _lp_env_vars_color; then
+        LP_ENVVARS="$lp_env_vars_color"
```
Rycieos, line 3817: This template variable needs docs.

Rycieos (changes requested):

liquidprompt
nojhan: I checked:
 - error meaning — OK
 - LP_MARK_JOBS_SEPARATOR — OK
 - CMAKE — OK
 - _lp_join — OK

Rycieos (changes requested):
Looks good, just needs to-long-line lint errors fixed.

nojhan:
Fixed and squashed.

Issue #724, opened 2022-08-16
OS name and version


nojhan:
Display the OS name and version.

### Description

Display somewhere the operating system name, optionally along with its version.
Probably use `lsb_release`.

Not sure this can be done in a portable way, especially for Linux.

Rycieos:
> Not sure this can be done in a portable way, especially for Linux.

That is going to be the trick. For the `external-tool-tester.sh`, I just dump as much info as possible, and let the human reader figure it out:

https://github.com/nojhan/liquidprompt/blob/6f0dc9f7dda99fe98c6dce26be5e461692782c6d/tools/external-tool-tester.sh#L24-L34

Portability is not likely without a bunch of testing and work put in. Might be easiest to just do `uname -o` or `uname -s` or even `uname` and return that as the OS. Open to ideas.

nojhan:
Here is [how Powerlevel10k is doing it](https://github.com/romkatv/powerlevel10k/blob/master/internal/p10k.zsh#L8416) in at most 3 subshell calls (MIT licensed):
```sh
  _p9k_uname="$(uname)"
  [[ $_p9k_uname == Linux ]] && _p9k_uname_o="$(uname -o 2>/dev/null)"
  _p9k_uname_m="$(uname -m)"

  if [[ $_p9k_uname == Linux && $_p9k_uname_o == Android ]]; then
    _p9k_set_os Android ANDROID_ICON
  else
    case $_p9k_uname in
      SunOS)                     _p9k_set_os Solaris SUNOS_ICON;;
      Darwin)                    _p9k_set_os OSX     APPLE_ICON;;
      CYGWIN*|MSYS*|MINGW*)      _p9k_set_os Windows WINDOWS_ICON;;
      FreeBSD|OpenBSD|DragonFly) _p9k_set_os BSD     FREEBSD_ICON;;
      Linux)
        _p9k_os='Linux'
        local os_release_id
        if [[ -r /etc/os-release ]]; then
          local lines=(${(f)"$(</etc/os-release)"})
          lines=(${(@M)lines:#ID=*})
          (( $#lines == 1 )) && os_release_id=${lines[1]#ID=}
        elif [[ -e /etc/artix-release ]]; then
          os_release_id=artix
        fi
        case $os_release_id in
          *arch*)                  _p9k_set_os Linux LINUX_ARCH_ICON;;
          *debian*)                _p9k_set_os Linux LINUX_DEBIAN_ICON;;
[...]

PR #756, opened 2022-11-03
feat(dev): adds OS data

nojhan:
- Implements #724
- Display arch, family, kernel, distrib, version codename.
- As a dev env item.
- With hash-colored items.
- Adds the `_lp_substitute` utility function.

LP_OS


liquidprompt
```diff
@@ -150,6 +150,129 @@ case "$(uname)" in
     *)         LP_OS=Linux   ;;
 esac
 
+_lp_os() {
+    # Fine-grained OS detection: family, kernel, distrib, version.
+    local uname_sm kernel
+    if ! uname_sm="$(uname -s -m)"; then
+        return 1
+    fi
+    read -r kernel lp_os_arch <<<"$uname_sm"
```
Rycieos, line 159: Need to set `IFS` here.

liquidprompt
```diff
@@ -150,6 +150,129 @@ case "$(uname)" in
...
+            fi
+            ;;
+    esac
+}
+
+_lp_os_color() {
+    # Substitute lp_os_* names with icons/colored strings (from LP_MARK_OS),
+    # or hash colored versions.
+    # Join everything with LP_MARK_OS_SEP.
+    (( LP_ENABLE_OS )) || return 2
```
Rycieos, line 223: This check needs to be at the top of `_lp_os()`, not here. This should be 
```suggestion
    _lp_os || return "$?"
```

liquidprompt
```diff
@@ -150,6 +150,129 @@ case "$(uname)" in
     *)         LP_OS=Linux   ;;
 esac
 
+_lp_os() {
+    # Fine-grained OS detection: family, kernel, distrib, version.
+    local uname_sm kernel
+    if ! uname_sm="$(uname -s -m)"; then
```
Rycieos, line 156: Add this specific call (`uname -s -m`) to the external-tool-tester script.

liquidprompt
```diff
@@ -150,6 +150,129 @@ case "$(uname)" in
...
+    # Join everything with LP_MARK_OS_SEP.
+    (( LP_ENABLE_OS )) || return 2
+    lp_os_color=
+    if _lp_os; then
+        local active=()
+        if (( LP_ENABLE_OS_ARCH )); then
+            if _lp_substitute "$lp_os_arch" "${LP_MARK_OS[@]}"; then
+                active+=("$lp_substitute")
+            else
+                _lp_hash_color "$lp_os_arch"
```
Rycieos, line 231: Using the hash color is clever, but should be noted in the documentation.

tests/test_utils.sh
```diff
@@ -847,6 +847,16 @@ function test_grep_fields {
     rm -f "$filename"
 }
 
+function test_substitute {
+    local sub=(
+        "When=NOK"
+        "What?=NOPE"
+        "What=OK"
+    )
+    _lp_substitute "What" ${sub[@]}
```
Rycieos, line 856: Pretty sure this should be quoted
```suggestion
    _lp_substitute "What" "${sub[@]}"
```

liquidprompt
```diff
@@ -1057,6 +1189,28 @@ _lp_fill() {
...
+    # FIXME If we ever support Bash>4.3, this would be better implemented with associative arrays.
+    lp_substitute=""
+    local string="${1}"
+    shift 1
+    local pairs=("$@")
+    [[ -n "${pairs[*]}" ]] || return 2
+
+    local key_rep key rep
+    for key_rep in "${pairs[@]}"; do
+        key="${key_rep%=*}"
```
Rycieos, line 1204: This will only work if the input strings have only a single `=`. If they have multiple, it will mangle. The easy fix is to change this to
```suggestion
        key="${key_rep%%=*}"
```
so that both expressions split on the first `=`. Obviously, if a user wants to replace a string containing a `=`, that will be impossible.

The other way to fix this is to have the pairs be passed in as separate arguments, like
`_lp_substitute "my long string" "long" "short"`

liquidprompt
```diff
@@ -150,6 +150,129 @@ case "$(uname)" in
...
+
+_lp_os_color() {
+    # Substitute lp_os_* names with icons/colored strings (from LP_MARK_OS),
+    # or hash colored versions.
+    # Join everything with LP_MARK_OS_SEP.
+    (( LP_ENABLE_OS )) || return 2
+    lp_os_color=
+    if _lp_os; then
+        local active=()
+        if (( LP_ENABLE_OS_ARCH )); then
```
Rycieos, line 227: These config flag checks really should be done in `_lp_os()`, where if the flag is false, the data is never even looked up in the first place.

There here (or any other theme function), you just check if the string is not empty.

Rycieos (changes requested):

liquidprompt
nojhan: I thought about the simple pair, but I failed to find an easy way to pass a whole list of replacements, instead of just one.

This would be so much simpler with associative arrays…

liquidprompt
nojhan: I was thinking that this may be used as a fine-grained tool for OS-dependent features. But at the end I decided not to touch `LP_OS`, which has a slightly different scheme. Should I try to get a single set of function for those two purposes (feature management VS data display), or should I keep them separated?

nojhan: Also, I would go for having the checks in both places, given that useless calls to `_lp_substitute`/`_lp_hash_color` may be a bit slow.

liquidprompt
Rycieos: I'm not sure why it would be harder with pairs of strings. Your example in the docs could be like
```bash
       LP_MARK_OS=(
           "x86_64" "${BLUE}x64${NO_COL}"
           "i386" "i3"
           "i686" "i6"
           ...
       )
```

The the for loop here would change like 
```diff
-    for key_rep in "${pairs[@]}"; do
-        key="${key_rep%=*}"
+    for (( i = 0; i < ${#pairs[@]}; I+=2 )); do
+        key=${pairs[_LP_FIRST_INDEX+i]}
         if [[ "$key" == "$string" ]]; then
-            rep="${key_rep#*=}"
+            rep=${pairs[_LP_FIRST_INDEX+i+1]-}
```

liquidprompt
```diff
@@ -150,6 +150,175 @@ case "$(uname)" in
...
+                    else
+                        return "$?"
+                    fi
+                fi
+                ;;
+        esac
+    fi
+}
+
+_lp_os_color() {
```
Rycieos, line 253: Function needs docs.

liquidprompt
```diff
@@ -150,6 +150,175 @@ case "$(uname)" in
...
+    # Substitute lp_os_* names with icons/colored strings (from LP_MARK_OS),
+    # or hash colored versions.
+    # Join everything with LP_MARK_OS_SEP.
+    _lp_os || return "$?"
+
+    lp_os_color=
+    local active=()
+    if (( LP_ENABLE_OS_ARCH )); then
```
Rycieos, line 261: Since these strings are not always going to be filled, why not just check if the string has contents?
```suggestion
    if [[ -n $lp_os_arch ]]; then
```
That removes a double check of the config option, as well as protecting against situations where the data was not found.

liquidprompt
Rycieos: I have a different suggestion, see other comment.

liquidprompt
```diff
@@ -150,6 +150,175 @@ case "$(uname)" in
...
+
+        case "$kernel" in
+            FreeBSD)
+                if (( LP_ENABLE_OS_FAMILY )); then lp_os_family=BSD; fi
+                if (( LP_ENABLE_OS_KERNEL )); then lp_os_kernel=FreeBSD; fi
+                ;;
+            DragonFly)
+                if (( LP_ENABLE_OS_FAMILY )); then lp_os_family=BSD; fi
+                if (( LP_ENABLE_OS_KERNEL )); then lp_os_kernel=FreeBSD; fi
+                ;;
```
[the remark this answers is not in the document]
nojhan: Is it considered two different distributions, then?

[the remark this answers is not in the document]
Rycieos: Yes, the `|` character is an `OR` in extended regex, which the context before the `)` is.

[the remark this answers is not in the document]
nojhan: My question was rather: should I set `lp_os_distribution="DragonFly"`?

liquidprompt
nojhan: Clever!

liquidprompt
nojhan: I should have thought about that. I like it better, indeed, it feels more natural than the separator.

liquidprompt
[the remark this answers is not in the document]
nojhan: Wikipedia states that "DragonFly has diverged significantly from FreeBSD". Maybe we should keep it as a separated kernel?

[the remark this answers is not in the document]
Rycieos: Makes sense to me.

docs/config.rst
```diff
@@ -559,6 +559,81 @@ Features
...
+
+   See :attr:`LP_ENABLE_OS`.
+
+   .. versionadded:: 2.2
+
+.. attribute:: LP_ENABLE_OS_KERNEL
+   :type: bool
+   :value: 1
+
+   Display the current kernel of the OS.
```
Rycieos, line 621: This could be clearer. Is it the kernel name, or version?

docs/config.rst
```diff
@@ -1138,6 +1213,68 @@ Marks
...
    .. versionadded:: 2.1
 
+.. attribute:: LP_MARK_OS
+   :type: array<string>
+   :value: ()
+
+   A list of pair of strings to be replaced by another string
+   when displaying information about the OS.
+
+   Each pair in the list figures the match, then the replacement string.
```
Rycieos, line 1223:
```suggestion
   Each pair in the list configures the match, then the replacement string.
```

liquidprompt
```diff
@@ -150,6 +150,167 @@ case "$(uname)" in
...
+                    fi # Linux
+                    lp_os_family=GNU
+                fi
+                # TODO parse HOME_URL and add an hyperlink.
+                if (( LP_ENABLE_OS_DISTRIB && LP_ENABLE_OS_VERSION )); then
+                    if _lp_grep_fields "/etc/os-release" "=" "ID" "VERSION_CODENAME"; then
+                        lp_os_distrib=${lp_grep_fields[_LP_FIRST_INDEX+0]-}
+                        lp_os_version=${lp_grep_fields[_LP_FIRST_INDEX+1]-}
+                    else
+                        return "$?"
```
Rycieos, line 233: Why does this need to be an error? Sure, the docs as written back this up:
> Returns ``1`` if some expected information was missing.

But why does this need to be the case? Usually, we do a best effort on data sources, returning as much information as we can get.

And even if that did make sense, why `return "$?"`? The docs suggest this case should be `return 1`.

I say the only case where returning 1 makes sense is if we have no data at all to return.

liquidprompt
```diff
@@ -150,6 +150,167 @@ case "$(uname)" in
...
+    # Substitute lp_os_* names with icons/colored strings (from LP_MARK_OS),
+    # or hash colored versions.
+    # Join everything with LP_MARK_OS_SEP.
+    _lp_os || return "$?"
+
+    lp_os_color=
+    local active=()
+    if [[ -n $lp_os_arch ]]; then
+        if _lp_substitute "$lp_os_arch" "${LP_MARK_OS[@]}"; then
+            active+=("$lp_substitute")
```
Rycieos, line 263: `$lp_substitute` and `$lp_hash_color` need to be declared local above.

Rycieos (changes requested):
Getting there, also some docs linting issues with too long lines.

liquidprompt
nojhan: Makes sense. But then I think we need to avoid Linux being the fallback, and just let `lp_os_kernel=""`.

This is then one reason to separate the OS detection for LP_OS and for the data display.

liquidprompt
nojhan: Good catch. Added `lp_join` as well.

Rycieos (approved):
Looks good, just needs squashing.

tests/test_utils.sh
```diff
@@ -847,6 +847,16 @@ function test_grep_fields {
     rm -f "$filename"
 }
 
+function test_substitute {
+    local sub=(
```
Rycieos, line 851: This breaks on Zsh 5.0:
```suggestion
    local sub
    sub=(
```

liquidprompt
```diff
@@ -1064,6 +1234,26 @@ _lp_fill() {
     lp_fill="${left}${filled}${right}"
 }
 
+_lp_substitute() { # string pairs_array
+    # Replace string with the "right" part of the first pair in the given array, if the "left" part matches the string.
+    # FIXME If we ever support Bash>4.3, this would be better implemented with associative arrays.
+    lp_substitute=""
+    local string="${1}"
+    shift 1
+    local pairs=("$@")
```
Rycieos, line 1243: This breaks on Zsh 5.0
```suggestion
    local pairs
    pairs=("$@")
```

liquidprompt
```diff
@@ -150,6 +150,167 @@ case "$(uname)" in
...
+    fi
+    if [[ -n "$lp_os_version" ]]; then
+        if _lp_substitute "$lp_os_version" "${LP_MARK_OS[@]}"; then
+            active+=("$lp_substitute")
+        elif _lp_hash_color "$lp_os_version"; then
+            active+=("$lp_hash_color")
+        else
+            active+=("$lp_os_version")
+        fi
+    fi
```
Rycieos, line 305: How about making this a for loop to reduce code duplication?

```suggestion
    for value in \
        "$lp_os_arch" "$lp_os_family" "$lp_os_kernel" "$lp_os_distrib" "$lp_os_version"; do
        if [[ -n "$value" ]]; then
            if _lp_substitute "$value" "${LP_MARK_OS[@]}"; then
                active+=("$lp_substitute")
            elif _lp_hash_color "$value"; then
                active+=("$lp_hash_color")
            else
                active+=("$value")
            fi
        fi
    done
```

tests/test_os.sh
```diff
@@ -0,0 +1,169 @@
...
+    LP_MARK_OS=("FreeBSD" "NOPE" "x86" "NOPE" "GNU" "G" "Linux" "L" "x86_64" "6" "ubuntu" "U" "focal" "F")
+
+    LP_MARK_OS_SEP="+"
+    LP_ENABLE_OS_ARCH=1
+    LP_ENABLE_OS_FAMILY=1
+    LP_ENABLE_OS_KERNEL=1
+    LP_ENABLE_OS_DISTRIB=1
+    LP_ENABLE_OS_VERSION=1
+    _lp_os_color
+    assertEquals "6+G+L+U+F" "$lp_os_color"
```
Rycieos, line 67: For some unknown reason this test errors with a stack recusion overflow on Zsh 5.0:
```
+test_os:60> _lp_os_color
+_lp_os_color:5> lp_os_color=''
+_lp_os_color:6> local lp_os_arch lp_os_family lp_os_kernel lp_os_distrib lp_os_version
+_lp_os_color:8> _lp_os
+_lp_os:2> ((  LP_ENABLE_OS  ))
+_lp_os:5> lp_os_arch=''
+_lp_os:6> lp_os_family=''
+_lp_os:7> lp_os_kernel=''
+_lp_os:8> lp_os_distrib=''
+_lp_os:9> lp_os_version=''
+_lp_os:11> local uname_sm kernel
+_lp_os:14> ((  LP_ENABLE_OS_ARCH  ))
+_lp_os:16> uname_sm=+_lp_os:16> uname -s -m
+uname:0> printf 'Linux x86_64'
+_lp_os:16> uname_sm='Linux x86_64'
+_lp_os:17> IFS=' ' +_lp_os:17> read -r kernel lp_os_arch
+_lp_os:22> ((  LP_ENABLE_OS_KERNEL
       || LP_ENABLE_OS_FAMILY
       || LP_ENABLE_OS_DISTRIB
       || LP_ENABLE_OS_VERSION  ))
+_lp_os:27> case Linux (FreeBSD)
+_lp_os:27> case Linux (DragonFly)
+_lp_os:27> case Linux (OpenBSD)
+_lp_os:27> case Linux (Darwin)
+_lp_os:27> case Linux (SunOS)
+_lp_os:27> case Linux (CYGWIN*)
+_lp_os:27> case Linux (MSYS*)
+_lp_os:27> case Linux (MINGW*)
+_lp_os:27> case Linux (Linux)
+_lp_os:61> ((  LP_ENABLE_OS_KERNEL  ))
+_lp_os:61> lp_os_kernel=Linux
+_lp_os:62> ((  LP_ENABLE_OS_FAMILY  ))
+_lp_os:63> [[ Linux == Linux ]]
+_lp_os:64> osfamily=+_lp_os:64> uname -o
+uname:0> printf 'Linux x86_64'
+_lp_os:64> osfamily='Linux x86_64'
+_lp_os:65> [[ 'Linux x86_64' == Android ]]
+_lp_os:70> lp_os_family=GNU
+_lp_os:73> ((  LP_ENABLE_OS_DISTRIB && LP_ENABLE_OS_VERSION  ))
+_lp_os:74> _lp_grep_fields /etc/os-release '=' ID VERSION_CODENAME
+_lp_grep_fields:0> lp_grep_fields=( ubuntu focal )
+_lp_os:75> lp_os_distrib=ubuntu
+_lp_os:76> lp_os_version=focal
+_lp_os:91> [[ -z x86_64GNULinuxubuntufocal ]]
+_lp_os_color:10> local lp_substitute lp_hash_color lp_join
+_lp_os_color:21> [[ -n GNU ]]
+_lp_os_color:22> _lp_substitute GNU FreeBSD NOPE x86 NOPE GNU G Linux L x86_64 6 ubuntu U focal F
+_lp_substitute:3> lp_substitute=''
+_lp_substitute:4> local 'string=GNU'
+local:1> [[ -n x86_64 ]]
+local:2> _lp_substitute x86_64 FreeBSD NOPE x86 NOPE GNU G Linux L x86_64 6 ubuntu U focal F
+_lp_substitute:3> lp_substitute=''
+_lp_substitute:4> local 'string=x86_64'
+local:1> [[ -n x86_64 ]]
+local:2> _lp_substitute x86_64 FreeBSD NOPE x86 NOPE GNU G Linux L x86_64 6 ubuntu U focal F
+_lp_substitute:3> lp_substitute=''
+_lp_substitute:4> local 'string=x86_64'
...
local:2: job table full or recursion limit exceeded
```

Rycieos (changes requested):
Found some weird Zsh 5.0 bugs while testing before merging.

liquidprompt
nojhan: Spotted another one line 260 `local active=()` — all fixed.

liquidprompt
Rycieos: I agree. And it looks like you have done that change already.

tests/test_os.sh
Rycieos: That other `active=()` you spotted (which I missed) was apparently the cause of this crash, because now it works.

Rycieos (changes requested):
Looks good now, just needs a final squash.

Rycieos (approved):

nojhan:
:partying_face:

PR #762, opened 2023-01-19
Features around presets: config-from-doc, several presets, more config var

### LINKED ISSUE #755: Up-to-date presets examples
I think that the `liquidpromptrc-dist` file is an important tool for feature discovery.
As of now, it does not contain all the new features and should be updated.

More generally, should we consider shipping a (small) subset of presets?
In the main repository, or should we have external presets, just as we have external themes?

nojhan:
Features:
- Adds the `config-from-doc.sh` script, to automatically generate a default "liquidpromptrc" config file from the documentation (fix #755).
- Adds presets for color schemes:
    - tritanopia (blue-yellow blindness),
    - protanopia-deuteranopia (red-green blindness),
    - 256 colors, dark and light schemes.
- Adds presets for "icons":
    - more-unicode,
    - nerd-font.
- Adds more configuration variables:
    - Adds `LP_COLOR_OS_*`.
    - Set `LP_MARK_OS_SEP` to the same separator than the CMake's one by default (more consistent).
    - Adds `LP_ENABLE_CMAKE_COLORS` for enabling auto color based on hash.
    - Adds `LP_COLOR_CMAKE_*` variables when one do not want hash colors.
    - Adds `LP_COLOR_SCLS`

Fix along the way:
- Load config at each prompt activation in `theme-preview.sh`, to allow preview of presets using public functions (i.e. `lp_terminal_format`).
- Reorder "dev env" sections to be more logical.


nojhan:
I was unsure whether such a large PR would be better squashed for review, so I let it be in several commits, thinking it will be easier to review. But, alas, some commits are fixing previous commits. So would you prefer a single-commit PR, I can squash it right away, just let me know.

tools/config-from-doc.sh
```diff
@@ -0,0 +1,73 @@
...
+                printf "WARNING: attribute %s default shows lp_terminal_format.\n" "$att" 1>&2
+            fi
+            IFS='' read -r # Empty line
+            IFS='' read -r line
+            if [[ -n "$line" ]]; then
+                if [[ "$line" != *"deprecated"* ]]; then 
+                    doc="$line"
+                    while [[ "$line" != "" ]]; do
+                        IFS='' read -r line
+                        doc="${doc}${line}"
```
Rycieos, line 49: I thought the description being on all one line was unreadable, so I think we should keep the original formatting:
```suggestion
                    doc=""
                    while [[ -n "$line" ]]; do
                        doc="${doc}"$'\n'"#${line}"
                        IFS='' read -r line
```

tools/config-from-doc.sh
```diff
@@ -0,0 +1,73 @@
...
+                    doc="${doc//\`\`/\`}" # Remove all double back ticks.
+                    if [[ "${doc:0-1}" != "." ]]; then
+                        printf "WARNING: doc for attribute %s do not end with a period.\n" "$att" 1>&2
+                    fi
+
+                    # Everything should be good here, we print.
+                    moreinfo="https://liquidprompt.readthedocs.io/en/stable/config.html#${att}"
+                    all="#${doc}\n# Type: ${atype}, more information: ${moreinfo}\n#${att}=${avalue}\n\n"
+                    all="${all//%/%%}" # Escape percents.
+                    printf "%s" "$all"
```
Rycieos, line 62: `printf` does not convert `\n` in the input strings, so you either need to imbed actual newlines, or just let `printf` do what it does best: formatting:
```suggestion
                    printf '%s\n# Type: %s\n# More information: %s\n%s\n' \
                      "$doc" "$atype" "$moreinfo" "$config_line"
```
Note this depends on my lp_terminal_format suggestion above.

tools/config-from-doc.sh
```diff
@@ -0,0 +1,73 @@
...
+                else
+                    printf "NOTE: attribute %s is deprecated, bypass.\n" "$att" 1>&2
```
Rycieos, line 64: Good to have useful messages, but your average user probably would not understand these. I think we should add a `--verbose` flag or similar that would enable displaying these.

tools/theme-preview.sh
```diff
@@ -21,10 +21,8 @@ else
   theme=$1
   shift
 
-  for file in "$@"; do
-    # shellcheck disable=SC1090
-    . "$file"
-  done
+  # We will source the given config files after each prompt activation,
+  # so as to allow presets using public functions (like `lp_terminal_format`).
```
Rycieos, line 25: I don't understand the need for this change. Loading a theme file doesn't run any code (or shouldn't, see the docs:
https://github.com/nojhan/liquidprompt/blame/master/docs/theme/custom.rst#L10-L13
> A theme should be contained in one file with a ``.theme`` file suffix. There should be no "top level" code in the file, or in other words, all code should be contained in functions. Sourcing the file should run no code, as a user sourcing the theme file might not want to activate it yet.

As long as we don't run the theme's activate function until everything is setup, there can't be any issues with loading them this early.

tools/config-from-doc.sh
```diff
@@ -0,0 +1,73 @@
...
+# even if the file does not ends with a \n.
+# This bypass a known behavior of the C standard, not fixed in POSIX.
+while IFS='' read -r line || [[ -n "$line" ]] ; do
+    if [[ "$line" == *".. attribute:: LP_"* ]]; then
+        att="${line##*' '}"
+        IFS=' ' read -r t atype
+        IFS=' ' read -r v avalue
+        if [[ "$t" == ":type:" && "$v" == ":value:" ]]; then
+            if [[ "$avalue" == *"lp_terminal_format"* ]]; then
+                printf "WARNING: attribute %s default shows lp_terminal_format.\n" "$att" 1>&2
```
Rycieos, line 40: I wrote a handler for this case:
```suggestion
                config_line="#$avalue"$'\n'"#${att}=\$lp_terminal_format"
            else
                config_line="#${att}=${avalue}"
```
Alternatively we could change the format in the documentation to more closely match what would be a config file, but multiple lines might be tricky.

docs/config.rst
```diff
@@ -1883,6 +2007,16 @@ Valid preset color variables are:
 
    .. versionadded:: 2.1
 
+.. attribute:: LP_COLOR_SCLS
+   :type: string
+   :value: $CYAN
```
Rycieos, line 2012: This isn't entirely true, as it actually defaults to `LP_COLOR_VIRTUALENV`. Somehow we should say that it defaults to `LP_COLOR_VIRTUALENV`, but also mention that that defaults to `CYAN`

contrib/presets/colors/protanopia-deuteranopia.conf
```diff
@@ -0,0 +1,105 @@
+# Preset for Red-green dichromacy color blindness.
+
+__lp_foreground_color 3
+YELLOW="${_LP_OPEN_ESC}${af_color}${_LP_CLOSE_ESC}"
+BOLD_YELLOW="${_LP_OPEN_ESC}${_LP_TI_BOLD-}${af_color}${_LP_CLOSE_ESC}"
```
Rycieos, line 5: Why use `_LP_TI_BOLD` when `lp_terminal_format` has that as a flag?

contrib/presets/colors/protanopia-deuteranopia.conf
```diff
@@ -0,0 +1,105 @@
...
+# LP_COLOR_PATH_ROOT=${BOLD_YELLOW}
+# LP_COLOR_PROXY=${BOLD_BLUE}
+# LP_COLOR_ENV_VARS_UNSET=${BLUE}
+# LP_COLOR_ENV_VARS_SET=${BOLD_BLUE}
+# LP_COLOR_JOB_D=${YELLOW}
+# LP_COLOR_JOB_R=${BOLD_YELLOW}
+# LP_COLOR_JOB_Z=${BOLD_YELLOW}
+# LP_COLOR_ERR=${PURPLE}
+# LP_COLOR_ERR_MEANING=${LP_COLOR_ERR}
+# LP_COLOR_MARK=${BOLD}
```
Rycieos, line 75: I'm not sure if this is something we can fix or not, but I loaded this preset after loading 256-colors-light.conf, and the mark became black. If I loaded this preset in a fresh env, it worked correctly.

contrib/presets/colors/protanopia-deuteranopia.conf
```diff
@@ -0,0 +1,105 @@
+# Preset for Red-green dichromacy color blindness.
```
Rycieos, line 1: At least for me, this didn't help much. Removing red entirely means you have less color options, meaning each color is less different. Which is the opposite of what I need.

Red and green look the same, but only if they are dark. The brighter they are, the easier they are to tell apart. A nice light green like `lp_terminal_format 84` and a bright red like `lp_terminal_format 196` are *very* easy to tell apart. It also helps that the relative intensities are not the same.

I struggle more with differences between light yellows and light greens, so the `LP_COLORMAP` here is very difficult to use. I think a spectrum of light to dark (as well as colors shifting, probably from green to blue to red?) would be much easier to use, say:
```bash
for num in 84 154 225 33 21 196 124 ; do
  lp_terminal_format $num
  printf '%s\n' "$lp_terminal_format foobar$(tput sgr0)"
done
```
Each value there is very clearly different to me.

But those are just my thoughts. And I have never been officially diagnosed, so I don't even know if I have Protanopia or Deuteranopia.

contrib/presets/icons/more-unicode.conf
```diff
@@ -0,0 +1,56 @@
...
+LP_MARK_BRACKET_OPEN="⦗"
+LP_MARK_BZR="⚯"
+LP_MARK_CMAKE="·"
+LP_MARK_DEV_CLOSE="❭"
+LP_MARK_DEV_MID="|"
+LP_MARK_DEV_OPEN="❬"
+LP_MARK_DIRSTACK="⚞"
+LP_MARK_DISABLED="⌀"
+LP_MARK_FOSSIL="⌘"
+LP_MARK_GIT="±"
```
Rycieos, line 19: Aren't these the defaults already? I'd rather see this file contain only things different than the default.

contrib/presets/icons/nerd-font.conf
```diff
@@ -0,0 +1,81 @@
+# This preset use symbols from Nerd fonts.
+# https://www.nerdfonts.com/
+# It is required that your terminal emulator use a patched font to see it correctly.
+
+LP_TIME_FORMAT="%Hₕ%Mₘ%Sₛ"
+LP_DELIMITER_KUBECONTEXT_PREFIX=""
+LP_DELIMITER_KUBECONTEXT_SUFFIX=""
```
Rycieos, line 7: These two do not load for me, where all the others do. Do you have extra extensions besides nerdfont?

Rycieos (changes requested):
In general good, though the `config-from-doc.sh` needs some work.

As for the commits, I'd like to keep the multiple commits for the different stuff, but squash the fix commits into the original feature commit. That can wait until ready to merge.

tools/config-from-doc.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
nojhan: I see that behavior implemented here and there and I'm not sure to like it. What if I want to grep some variables to make my own preset? Or sort the lines, automate the enabling of all features, ...?

Wouldn't it be a better compromise to print a warning message on stderr and let the script continue, just in case?

tools/config-from-doc.sh
```diff
@@ -0,0 +1,73 @@
...
+
+file="../docs/config.rst"
+
+printf "# This file shows the default presets configuration for Liquid Prompt.
+#
+# You may edit it and load it from your shell configuration file
+# to have your own configuration.
+#
+# For instance:
+# cp liquidprompt/liquidpromptrc-dist ~/.liquidpromptrc
```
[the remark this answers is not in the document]
nojhan: I think you are right. Issue #765 shows that this can be confusing.

tools/config-from-doc.sh
nojhan: My point with this warning was indeed to ask the question whether developers could come up with a simpler default than calling `lp_terminal_format`.

docs/config.rst
nojhan: I was wondering whether this was a good idea to have dependencies between segments configurations. As far as I recall, this is the only occurrence, so maybe we should consider set it to CYAN and be done with it, to stay consistent?

contrib/presets/colors/protanopia-deuteranopia.conf
nojhan: I've read some explanations since the PR and it seems indeed that the mild conditions are generally treated differently than the severe ones. So I guess we should probably have presets for both cases, with either color removal (severe case), or contrast increase (mild case).

But then I am wondering: can we really have more contrast than the 16 colors palette? Especially considering that most terminal emulator allow for fine-tuning those colors. In which case, the default theme may very well be the best we can do for mild cases.

tools/config-from-doc.sh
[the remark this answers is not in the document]
Rycieos: This test only checks if `/dev/stdout` is a terminal. If you pipe the output to another command, it will also fail:
```bash
$ ./tools/external-tool-tester.sh | less
---Uname:
...
```
So I disagree. You could add some sort of argument that could be passed if someone really wanted to print to the terminal, but it would be just as easy to pipe to `less`

contrib/presets/colors/protanopia-deuteranopia.conf
Rycieos: That's fair. It is not an easy problem to solve.

It's also possible that we can't even come up with solutions that work for everyone or even multiple people.

tools/config-from-doc.sh
```diff
@@ -0,0 +1,73 @@
...
+                    doc="${doc//   / }" # Remove all triple saces (indentations).
+                    doc="${doc//\`\`/\`}" # Remove all double back ticks.
+                    if [[ "${doc:0-1}" != "." ]]; then
+                        printf "WARNING: doc for attribute %s do not end with a period.\n" "$att" 1>&2
+                    fi
+
+                    # Everything should be good here, we print.
+                    moreinfo="https://liquidprompt.readthedocs.io/en/stable/config.html#${att}"
+                    all="#${doc}\n# Type: ${atype}, more information: ${moreinfo}\n#${att}=${avalue}\n\n"
+                    all="${all//%/%%}" # Escape percents.
```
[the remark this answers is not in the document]
nojhan: I don't recall, to be honest.

tools/config-from-doc.sh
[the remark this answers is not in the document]
nojhan: I see. This is a good-enough compromise, then.

contrib/presets/icons/nerd-font.conf
nojhan: I thought I only had nerdfont, but it's very possible I have more. I remember having some difficulties finding my default font. I see that the icon (first character of the prefix) is in nerdfont, so I just replaced the other ones.

contrib/presets/icons/more-unicode.conf
nojhan: Good catch, I don't know how you can see that ^^

contrib/presets/colors/protanopia-deuteranopia.conf
nojhan: I renamed the presets, so that the underlying claims are less bold (and maybe easier to understand, also).

tools/theme-preview.sh
nojhan: For some presets, I need to call `lp_terminal_format`, but it is not available when sourcing presets early on, so that was a way to bypass the limitation. Is there another way of doing that? Having a common interface to set up 8-bit colors is really a need for presets, IMHO.

contrib/presets/colors/protanopia-deuteranopia.conf
Rycieos: The right way to solve this is to reload defaults before loading a preset (or to have a way to unload a loaded preset somehow), which isn't possible with our current way of doing config. I would say probably not.

tools/config-from-doc.sh
```diff
@@ -0,0 +1,120 @@
...
+    ./tools/config-from-doc.sh > default_commented.conf
+
+It must be started from the root directory of the liquidprompt repository,
+so that it finds the right documentation file.
+    
+By default, every key=value are commented out.
+You must edit it manually to get a working configuration file.
+
+You may enable warning about the parsing with the '--verbose' flag.
+Warnings will be printed of stderr and will not appear in the output file.
```
Rycieos, line 21:
```suggestion
Warnings will be printed to stderr and will not appear in the output file.
```

tools/config-from-doc.sh
```diff
@@ -0,0 +1,120 @@
...
+
+The output is made so that it can be easily manipulated.
+For instance if you want to get rid of comments and enable every features:
```
Rycieos, line 24:
```suggestion
For instance if you want to get rid of comments and enable every feature:
```

tools/config-from-doc.sh
```diff
@@ -0,0 +1,120 @@
...
+    ./tools/config-from-doc.sh | grep '^#\S' | sed 's/=0$/=1/' > full_raw.conf
+"
+
+if [ -t 1 ]; then
+    # Error if the output is a terminal
+    printf 'ERROR: This script must be redirected to a file or piped.\n\n' 1>&2
+    printf "$usage" 1>&2
+    exit 2
+    fi
```
Rycieos, line 33:
```suggestion
fi
```

tools/config-from-doc.sh
```diff
@@ -0,0 +1,120 @@
...
+
+if [[ $# -gt 1 ]]; then
+    printf "ERROR: This script don't take more than one argument.\n\n" 1>&2
```
Rycieos, line 36:
```suggestion
    printf "ERROR: This script does not accept more than one argument.\n\n" 1>&2
```

tools/theme-preview.sh
Rycieos: Wait, are you using this script to test config presets? It was never designed to do that, as you have discovered. It was designed for themes.

Hm. I bet your hack here would work. I can't think of any reason that it wouldn't, but if the goal is to support loading both theme and config preset files, the docs on this script should be updated to mention that.

I don't think that you need to source the files multiple times though. Nothing should overwrite the config options later.

liquidpromptrc-dist
```diff
@@ -1,240 +0,0 @@
-
```
Rycieos, line 1: I agree that this file should be deleted, but there are many places that it is explicitly referenced:
```bash
$ grep -RniI liquidpromptrc-dist
docs/config.rst:29:Liquid Prompt ships with an example config file, ``liquidpromptrc-dist``. You
docs/config.rst:32:    cp ~/liquidprompt/liquidpromptrc-dist ~/.config/liquidpromptrc
liquidprompt:626:            # shellcheck source=liquidpromptrc-dist
contrib/dist/archlinux-aur/PKGBUILD:37:  install -Dm644 liquidpromptrc-dist "$pkgdir/etc/liquidpromptrc"
contrib/dist/debian/create_deb.sh:8:cp ../../liquidpromptrc-dist ./liquidprompt/etc/liquidpromptrc
```
These ideally would all be replaced with references to the new script, or in the case of the scripts in `contrib/dist/`, replaced with calls to the script.

The tricky one is going to be in `liquidprompt` itself, as we won't have a file laying around for `shellcheck` to use. Probably the simplest and best solution is to replace the line with `# shellcheck source=/dev/null`.

Rycieos (changes requested):

nojhan:
Fixed and rebased.
I notably updated the Debian package definition file.

docs/config.rst
```diff
@@ -1511,6 +1650,7 @@ Valid preset color variables are:
 
 .. attribute:: LP_COLORMAP
    :type: array<string>
+   :value: ( "" $GREEN $BOLD_GREEN $YELLOW $BOLD_YELLOW $RED $BOLD_RED $WARN_RED $CRIT_RED $DANGER_RED )
```
Rycieos, line 1653: >  Line too long (104/80) (line-too-long)

You can wrap the line and it still works:
```suggestion
   :value: ( "" $GREEN $BOLD_GREEN $YELLOW $BOLD_YELLOW $RED
             $BOLD_RED $WARN_RED $CRIT_RED $DANGER_RED )
```

There is also a few lines below a definition of the default as well; that should be removed.

contrib/dist/debian/liquidprompt/DEBIAN/control
```diff
@@ -1,11 +1,11 @@
 Package: liquidprompt
-Version: 1.3-0
+Version: 2.2.0
 Section: base
 Priority: optional
 Architecture: all
 Depends: bash
 Recommends: acpi
-Suggests: git, screen, tmux
+Suggests: git, svn, fossil, mercurial, bzr, cmake, screen, tmux, kubernetes
```
Rycieos, line 8: `kubernetes` is not a package usually, or if it is, it's the server.

What we depend on is `kubectl`, which is normally not installed via package management. But if it was, it would be called `kubectl`: https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#install-using-native-package-management

tools/config-from-doc.sh
```diff
@@ -0,0 +1,120 @@
...
+# For instance:
+# source ~/whatever/my_liquidprompt.conf
+# source ~/liquidprompt/liquidprompt
+#
+# Entries are commented out and show the default values.
+# There is no need to uncomment them if you do not change
+# their value from the default.
+#
+# Note that the file is automatically extracted from the documentation,
+# but do not show all the help for each entry.
```
Rycieos, line 66:
```suggestion
# but does not show all the help for each entry.
```

tools/config-from-doc.sh
```diff
@@ -0,0 +1,120 @@
...
+# You should refer to the full documentation to get more details:
+# https://liquidprompt.readthedocs.io/ (in the 'Config Options' section).
```
Rycieos, line 68: Why not just link to there?
```suggestion
# https://liquidprompt.readthedocs.io/en/stable/config.html
```

Rycieos (changes requested):
I think you did not rebase correctly, because you are adding things here that are already on `master`. I can see some commits added on this PR that are duplicates of ones merged in (b0031af vs a2c59d2). You'll need to fix that before I can truly review, since this inflates the diff.

Issue #765, opened 2023-02-18
Too difficult to set up


kiteloopdesign:
### Description

doc is very good at describing the low-level stuff for developers. However, for the final user, It took me some time to find how to do the simplest of things: loading a theme

From `liquidprompt-dist` file, the following info is given:

```
# If you want to use different themes and features,
# you can load the corresponding files here:
#source ~/.config/liquidprompt/nojhan.theme
#LP_PS1_FILE=~/.config/liquidprompt/nojhan.ps1
```
However these files do not exist on the liquidprompt installation folder...

This is also contradictory to what actually needs to be done. Basically this:

```
source themes/powerline/powerline.theme
lp_theme lp_theme powerline
```

Documentation in rdt is not clear on this, info on how to set the theme should be first line!

![image](https://user-images.githubusercontent.com/17339330/219899306-d270d829-d313-48ec-9ca8-7db6e1f1c8c2.png)

![image](https://user-images.githubusercontent.com/17339330/219899339-e3b00ccc-a5bd-4b62-9335-374967ab6867.png)

kiteloopdesign:
Also this doesn't seem to work:
```
source liquid.theme 
lp_theme liquid
```
Thanks

nojhan:
That's a good point! Thanks for having taken the time to report it, this is very useful to us.

The documentation of the next release (not online on RTD at the moment, but can be built from the source) is already a little better, and we will take special care to make the theme switch clearer.

Rycieos:
> The documentation of the next release (not online on RTD at the moment, but can be built from the source) 

This isn't exactly correct, as the `master` branch is always built and can be seen here: https://liquidprompt.readthedocs.io/en/latest/. But the default version does track the latest release.

However, you are correct that we can improve this section. Thanks for the suggestion!

nojhan:
On top of the PR #767, I also added a short installation instruction on [the related wiki page](https://github.com/nojhan/liquidprompt/wiki/Themes), hope that helps.

nojhan:
Fixed.

PR #767, opened 2023-02-23
fix(doc): simpler doc on how to switch themes

nojhan:
+ remove deprecated/confusing files

Refs: #765

Not sure how to handle the doc linter complaining about `lp` in a code block being a spell check error?


docs/theme.rst
```diff
@@ -28,7 +28,21 @@ If you want to create your own theme, see :doc:`theme/custom`.
...
 
 Switching Themes
-----------------
+================
+
+In a nutshell
+-------------
+
+To use a theme, just source its file in your shell configuration file,
+after having load the liquidprompt, and then call `lp_theme`:
```
Rycieos, line 37: This should be double backtics:
And, to make the following a code block (and therefore ignore spelling issues), you need two colons:
```suggestion
after having load the liquidprompt, and then call ``lp_theme``::
```

Rycieos (changes requested):
Looks good, just the fix and a rebase to `master` to fix Shellcheck issues.

liquidpromptrc-dist
```diff
@@ -1,240 +0,0 @@
-
```
Rycieos, line 1: You are already deleting this file in #762, let's not duplicate work.

liquid.ps1
```diff
@@ -1,29 +0,0 @@
-
```
Rycieos, line 1: This file has a few references, they will need to be cleaned up:
```bash
± grep -RniI liquid.ps1
liquidprompt:4210:        # shellcheck source=liquid.ps1
docs/theme/default.rst:70:An example template file is available: liquid.ps1_.
docs/theme/default.rst:72:.. _liquid.ps1: https://github.com/nojhan/liquidprompt/blob/master/liquid.ps1
```

Rycieos (changes requested):

liquid.ps1
nojhan: I added themes/minimal/minimal.ps1, which should be less confusing. It is just default minus a couple of sections.

nojhan:
Rebased on master.

liquidprompt
```diff
@@ -4207,7 +4275,7 @@ _lp_default_theme_prompt_data() {
 
 _lp_default_theme_prompt_template() {
     if [[ -f "${LP_PS1_FILE-}" ]]; then
-        # shellcheck source=liquid.ps1
+        # shellcheck source=themes/minimal/minimal.ps1
```
Rycieos, line 4278: >         source "$LP_PS1_FILE"
>                ^------------^ SC1091 (info): Not following: themes/minimal/minimal.ps1: openBinaryFile: does not exist (No such file or directory)

docs/theme/default.rst
```diff
@@ -67,9 +67,9 @@ For a template file to be loaded, its filepath must be set in
 A template file does nothing more than set :attr:`LP_PS1` to a value. The
 following sections are available to be used.
 
-An example template file is available: liquid.ps1_.
+An example template file is available: minimal.ps1_.
 
-.. _liquid.ps1: https://github.com/nojhan/liquidprompt/blob/master/liquid.ps1
+.. _minimal.ps1: https://github.com/nojhan/liquidprompt/blob/master/themes/minimal/minimal.ps1
```
Rycieos, line 72: This file does not exist.

tools/theme-preview.sh
```diff
@@ -126,6 +126,7 @@ _lp_hostname() {
 }
 _lp_connection() {
   lp_connection=ssh
+  SSH_CONNECTION="192.168.0.254 56314 192.168.0.1 22"
```
Rycieos, line 129: This change was already added in 8ea315e. Either drop the last commit, or rebase to latest `master`.

Rycieos (changes requested):

liquidprompt
nojhan: Sorry :man_facepalming:

themes/minimal/minimal.ps1
```diff
@@ -0,0 +1,15 @@
...
+PS1="${LP_PS1_PREFIX}"
+# add user, host, permissions colon, working directory, and dirstack
+PS1+="${LP_BRACKET_OPEN}${LP_USER}${LP_HOST}${LP_PERM}${LP_PWD}${LP_BRACKET_CLOSE}"
+
+# Add VCS infos
+# If root, the info has not been collected unless LP_ENABLE_VCS_ROOT
+# is set.
+PS1+="${LP_VCS}"
+
+# add return code and prompt mark
```
Rycieos, line 11: Most of these comments are incorrect:
> add title escape time, jobs, load and battery

None of that is added.
> dirstack

That isn't in there either.
> add return code and prompt mark

Runtime is also added here.

Rycieos (changes requested):

themes/minimal/minimal.ps1
nojhan: Fixed the related comments in `liquidprompt` as well.

liquidprompt
```diff
@@ -4293,7 +4293,7 @@ _lp_default_theme_prompt_template() {
         # is set.
         PS1+="${LP_VCS}"
 
-        # add return code and prompt mark
+        # Add last runtime, return code & meaning, prompt mar and user-defined postfix.
```
Rycieos, line 4296:
```suggestion
        # Add last runtime, return code & meaning, prompt mark and user-defined postfix.
```

Rycieos (changes requested):
Looks good, other than the typo.

nojhan:
Rebased on current master.

PR #769, opened 2023-02-24
fix(theme-preview): adds a fake SSH_CONNECTION

nojhan:
Without the env variable, theme preview ends on an "unset variable" error.


Rycieos (approved):
Nice catch. I had to unset my own `SSH_CONNECTION` to make this error of course, which is why I never caught this:
```bash
env -i TERM=xterm-256color ./tools/theme-preview.sh default
```

Issue #52, opened 2012-08-16
Memory alerts feature?


nojhan:
Should we have a memory meter (just like the battery and load ones)?

Emmanuel Rouat have suggested (some weeks ago) to also add a swap detector:

```
# Experimental: detect if machine is swapping
function swap_color()
{
    if [ $(vmstat | awk 'END{ print $7 }') -gt "0" ]; then  # swapin > 0
        echo -en ${ALERT}
    fi
}
```

Rycieos:
I came up with an implementation for Linux. It doesn't use any forks or calls, and only loads one file, `/proc/meminfo`.

```bash
mempresure() {  # return: percentage of memory free
  local id value total available
  while IFS=$' \t' read -r id value _ ; do
    if [[ $id == 'MemTotal:' ]]; then
      total=$value
    elif [[ $id == 'MemAvailable:' ]]; then
      available=$value
    fi
  done < /proc/meminfo
  printf '%s\n' "$(( available * 100 / total ))"
}
```

I just don't know how useful this would be. Is there still anyone who would actually want this?

nojhan:
Having myself regular memory problems, I would be interested. This may be disabled by default, of course.

Rycieos:
Merged in #770

Issue #730, opened 2022-08-17
Disk usage


nojhan:
Display disk usage, if near to full.

### Description

Display the remaining free space on selected file systems if it goes below a threshold, just like other kind of sensors.
Probably should pay special (default?) attention to file systems hosting `/`, `/home` and `/boot`.

### How will this be useful?

Full disk are a rare source of problem, but it may be quite difficult to spot when it happens (as the system does not fail, but behave quite erratically).
I witnessed users tempted to reboot in such case, which render the debugging quite difficult after, albeit it would have been easy to fix right away while connected.
This would raise attention to the problem before it become difficult to solve.

Rycieos:
I would not find this useful, but I'm sure some people would.

Biggest issue is gathering this data. I'm sure Linux probably has a sysfs interface, which would make this fast, but on other OSes we would need to do some hacky `df` parsing.

nojhan:
Using `df` would have the advantage of being quite portable.

What about just checking the filesystem mounted for the current directory?

Zsh can do it with a one-liner:
```zsh
echo ${${=${(f)"$(df -P . 2>/dev/null)"}[2]}[5]%%%}
```

But for Bash I only came up with this:
```bash
FS=$'\n' read -d "" -r -a arr <<< "$(df -P . 2>/dev/null)"
IFS=" " read -r fs bc ut ds capa mo <<< "${arr[1]}" ; echo ${capa%%%}
```

Rycieos:
> What about just checking the filesystem mounted for the current directory?

I like that idea a lot. I didn't know that `df` allows for a path argument.

> ```shell
> FS=$'\n' read -d "" -r -a arr <<< "$(df -P . 2>/dev/null)"
> IFS=" " read -r fs bc ut ds capa mo <<< "${arr[1]}" ; echo ${capa%%%}
> ```

That works, but I don't love needing two `read` calls, though it is only one external binary call.

nojhan:
What about this:
```bash
read -d'' -r _ _ _ _ _ _ _ _ _ _ _ fs _ _ dispo perc mnt <<<$(df -P . 2>/dev/null)
```

Rycieos:
That doesn't seem to work:
```bash
$ IFS=" " read -d'' -r _ _ _ _ _ _ _ _ _ _ _ _ _ used avail perc mnt <<<$(df -P . 2>/dev/null)
$ echo "$perc"

```
I think read doesn't work very well with multiple lines input...

[...]

PR #770, opened 2023-02-27
feat: adds a RAM sensor

nojhan:
Displays a (colored) mark (and a percentage) when the available of RAM goes below a threshold.

Refs: #52

# Technical checklist:

- code follows our [shell standards](https://github.com/nojhan/liquidprompt/wiki/Shell-standards):
    - [x] correct use of `IFS`
    - [x] careful quoting
    - [x] cautious array access
    - [x] ran `test.sh`

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -827,6 +827,33 @@ Features
 
    See also: :attr:`LP_COLOR_PROXY`.
 
+.. attribute:: LP_ENABLE_RAM
+   :type: bool
+   :value: 1
+
+   Display a :attr:`LP_MARK_RAM` mark when the available amount of
+   Random Access Memory goes below a threshold.
+
+   Thresholds can be stated either:
+
+   * as a percentage with :attr:`LP_RAM_THRESHOLD_PERC`,
+   * or an absolute number *of kilobytes* with :attr:`LP_RAM_THRESHOLD`.
+
+   Display will occur if one of the thresholds is met.
+
+   If :attr:`LP_ALWAYS_DISPLAY_VALUES` is enabled, the prompt will show the
+   available space along with :attr:`LP_MARK_RAM`, if disabled, it will show
+   only the mark.
+
+   The precision of the displayed available space can be configured with
+   :attr:`LP_RAM_PRECISION`.
+
+   If :attr:`LP_DISPLAY_VALUES_AS_PERCENTS` is enabled, it will show the
+   percentage, if it is disabled, it will show the absolute value in a
+   human-readable form (i.e. with metric prefixed units).
+
+   .. versionadded:: 2.2
+
 .. attribute:: LP_ENABLE_RUBY_VENV
    :type: bool
    :value: 1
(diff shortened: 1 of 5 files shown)


nojhan:
Looking back at #730, I am now wondering whether handling the percentage of available memory is the right approach. Should we rather display/configure values in human readable amount? (branch `feat/disk` have `__lp_bytes_to_human`, for instance).

liquidprompt
```diff
@@ -3738,6 +3743,142 @@ _lp_load_color() {
...
+        __lp_ram_bytes() {
+            lp_ram_bytes=0
+            lp_ram_total_bytes=0
+
+            # Call vm_stat.
+            local stat=$(vm_stat 2>/dev/null)
+
+            # Read first line.
+            local page_size _ second third fourth fifth
+            IFS=$' ' read -r _ _ _ _ _ _ _ page_size _ <<< "$stat"
```
Rycieos, line 3808: Don't need the $ without escape chars
```suggestion
            IFS=' ' read -r _ _ _ _ _ _ _ page_size _ <<< "$stat"
```

Rycieos (changes requested):
I think it makes sense to support both percentages and absolute values as display. And ideally we would collapse into MiB or KiB to prevent the segment from being too large.

liquidprompt
```diff
@@ -3738,6 +3743,142 @@ _lp_load_color() {
...
+                    lp_ram_bytes=$value
+                    # Note: this is the best conservative estimate from the kernel
+                    # of the memory than may be available for a process.
+                    # See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773
+                fi
+                if [[ -n "$lp_ram_total_bytes" && -n "$lp_ram_bytes" ]]; then
+                    break
+                fi
+            # Assumes file exists.
+            done < $_LP_LINUX_RAM_FILE
```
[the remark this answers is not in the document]
nojhan: I don't see this warning with shellcheck 0.7.0 :disappointed:

nojhan:
I am waiting for PR #771 to be finished to get back at this, so that RAM and HD management would be consistent.

nojhan:
Since PR #771 has been merged, I rebased this one on top of it, using a similar implementation.

Rycieos (changes requested):
I'm not sure what you did, but somehow you replicated all of the commits on master onto your branch. Which obviously has a lot of conflicts. Probably the way to fix this is something like
```bash
git checkout -b wip/mem f95a43f
git rebase origin/master
```
There will be a few conflicts (because of feat/disk), but they are easy to fix. I did exactly that and pushed the `wip/mem` branch, you can `git reset --hard origin/wip/mem` if you think it looks good.

I didn't look at the diff too hard (since reviewing that on GitHub is much easier), but I did notice and wanted to point out to remember to move the `if [ -n "${ZSH_VERSION-}" ] ...` block to the top of the test file after the `set -u` line.

nojhan:
Every time I try an acrobatic rebase, I get problems instead of solution…

This should be good now, here is what I did:
```sh
git co master
git co -b feat/mem2
git cherry-pick 49201b05da985173622ba364e8d79e322c6799da
# Merge & fix missing stuff.
git co feat/mem
git reset --hard origin/master
git cherry-pick efa6f8e6566ea0b0af679a6b5c2f7c1990719475
git push -f
```

docs/config.rst
```diff
@@ -827,6 +827,33 @@ Features
...
+   * as a percentage with :attr:`LP_RAM_THRESHOLD_PERC`,
+   * or an absolute number *of kilobytes* with :attr:`LP_RAM_THRESHOLD`.
+
+   Display will occur if one of the thresholds is met.
+
+   If :attr:`LP_ALWAYS_DISPLAY_VALUES` is enabled, the prompt will show the
+   available space along with :attr:`LP_MARK_RAM`, if disabled, it will show
+   only the mark.
+
+   The precision of the available space can be configured with
```
Rycieos, line 848:
```suggestion
   The precision of the displayed available space can be configured with
```

liquidprompt
```diff
@@ -3759,6 +3769,148 @@ _lp_load_color() {
...
+        }
+        ;;
+    Darwin)
+        # Parse vm_stat
+        __lp_ram_bytes() {
+            lp_ram_avail_bytes=0
+            lp_ram_total_bytes=0
+
+            # Call vm_stat.
+            local stat=$(vm_stat 2>/dev/null)
```
Rycieos, line 3832: Needs quoting to not fail on Zsh 5.0
```suggestion
            local stat="$(vm_stat 2>/dev/null)"
```

Rycieos (changes requested):
Looking real good. Only one Zsh 5.0 bug. And a few `IFS=$' '`s that could be `IFS=' '` instead.

And sorry for the delay. My development server's main drive failed, and I had to reinstall the OS.

nojhan:
No problem, there is no hurry :-)

How should we handle the Zsh 5 bug?

Rycieos:
You already fixed it. A squash and I'll approve.

Rycieos (approved):

nojhan:
Great!

PR #772, opened 2023-03-03
feat(presets): adds config-from-doc and several presets

nojhan:
Add two presets:
- more-unicode
- nerd-font

Add four color schemes:
- 256 colors for dark themes.
- 256 colors for light themes.
- red-green blindnesses.
- blue-yellow blindness.

Doc:
- mention presets in themes and config.
- fix doc about OS and CMake.

Misc:
- Also uses exit code 2 on tools scripts, which is the convention for usage errors.
- update debian pkg

# Technical checklist:

- code follows our [shell standards](https://github.com/nojhan/liquidprompt/wiki/Shell-standards):
    - [ ] correct use of `IFS`
    - [ ] careful quoting
    - [ ] cautious array access
    - [ ] functions/variable naming conventions
    - [ ] functions have local variables
    - [ ] ran `test.sh`


docs/config.rst
```diff
@@ -1677,6 +1679,36 @@ Valid preset color variables are:
...
+   :type: string
+   :value: $MAGENTA
+
+   Color used to display the C compiler in the CMake segment.
+
+   See :attr:`LP_ENABLE_CMAKE`.
+
+   .. versionadded:: 2.2
+
+.. attribute:: LP_COLOR_CMAKE_CXX
```
Rycieos, line 1702: Maybe we should consider making an array for cmake colors. That is tough though as long as we can't use associative arrays / dictionaries.

docs/theme.rst
```diff
@@ -8,6 +8,11 @@ The :doc:`theme/default` has a templating engine (previously called "themes" in
 Liquid Prompt version 1), that allows for custom prompt ordering in the default
 theme.
 
+If you just want to change icons and/or colors, you can just rely on "presets",
+which are just regular configuration files.
+Presets can generally be combined, and some themes may use
+—or be compatible with— different presets. See :doc:`config`.
```
Rycieos, line 14: Is this a French thing? English generally uses commas to break sentences like this:
> and some themes may use, or be compatible with, different presets.

or parens:
> and some themes may use (or be compatible with) different presets.

The Em dash is still in use, but spaces surrounding them should be removed:
> and some themes may use—or be compatible with—different presets.

I'm not sure how to remove the space when it falls on a line break.

liquidprompt
```diff
@@ -633,7 +633,7 @@ __lp_source_config() {
         done
     fi
 
-    # Delete this code in version 1.11
+    # FIXME Delete this code in version 1.11
```
Rycieos, line 636: I'd rather just remove this completely, since we are just keeping these shims to prevent breakage.

Rycieos (changes requested):
Looking pretty good mostly.

docs/config.rst
nojhan: It's been several time the associative arrays may have been a cleaner option. Should we consider bumping shell version requirements?

docs/theme.rst
nojhan: In both french and english (regardless of their national variants) the em dash may be used when a comma is too weak and parentheses too strong (which I felt was the case here).

As for typography, french use them opened (space dash thin-space word), while in english it is either closed (no space) or opened (space dash hair-space word). For instance, the AP style book sets it open with the smallest ("hair") space, which I render as semi-open, probably as a compromise with my inner french and the necessity of correct line breaks (which are usually not good with the closed version, whether the open one allows you to use non-breaking space).

As usual, Wikipedia has an article about that question: https://en.wikipedia.org/wiki/Dash#Em_dash

All in all, if it's too pompous in american english, we may very well get rid of it (and use parentheses) without me getting the slightest upset.

Rycieos: Nice summary. I probably prefer parentheses because of the logic/programming connotations. But you are correct, there is nothing wrong with the Em dash. I'm fine leaving it.

docs/config.rst
Rycieos: Considering it is a good idea. My gut reaction is that we should release v2.2 first, since it would be a breaking change and would fit better in a v3.0. But we should talk about it.

nojhan: So, should we merge this PR and aim for the release?

augmentedfourth:
This broke the Homebrew formula, which copies liquidprompt-dist from the root of the repo into `/usr/local/share`.

@Rycieos Would you prefer the Homebrew installation to generate the -dist file and provide it as an example (as it always has), or give users a link to the documentation here on GitHub? I can do the documentation link pretty easily, but it would take a while for me to find time to make Homebrew generate and deploy the default config file.

augmentedfourth:
Proposed Homebrew documentation change if I go with the latter (simpler) version:

```
-      A sample file you may copy and modify has been installed to
-        #{HOMEBREW_PREFIX}/share/liquidpromptrc-dist
+      Configuration documentation can be found on GitHub:
+        https://github.com/nojhan/liquidprompt/blob/master/docs/config.rst
```

augmentedfourth:
Maybe the appropriate documentation link should be [this one](https://liquidprompt.readthedocs.io/en/stable/config.html)?

Rycieos:
I would rather see links to documentation, as it will be the most helpful. You could also add a bit about the generation script and how to use it since some might find that useful.

Yes, that is the correct link for the `stable` release, but for those installing from the `master` branch, it would be nice to link the "latest" version instead: `https://liquidprompt.readthedocs.io/en/latest/config.html`. If you could have the documentation differ based on which version they install that would be great.

augmentedfourth:
OK. I've set up a commit where the regular install will output:
```
If you'd like to reconfigure options, you may do so in ~/.liquidpromptrc.
Configuration documentation can be found here:
  https://liquidprompt.readthedocs.io/en/stable/config.html
```

And `HEAD`/`master` install will output:
```
If you'd like to reconfigure options, you may do so in ~/.liquidpromptrc.
Configuration documentation (for the HEAD version) can be found here:
  https://liquidprompt.readthedocs.io/en/latest/config.html
```

I will include a message about generating a default config file once there's a release that includes the tool. Adding that message to just `master` and then adding it in all situation later was more complex than I'd like.

https://github.com/Homebrew/homebrew-core/pull/136761

augmentedfourth:
Argh. It looks like the current generation of Homebrew maintainers are trying to stop hosting other projects' documentation and only use their "caveats" section for details that are specific to the Homebrew distribution.

"most people do not read the caveats and even less people will if we start using it for all kinds of documentation."

They're asking me to remove that `readthedocs` link entirely. I might be able to re-add the generated sample config file in the future once the tool exists in an official release.

I'd prefer to do nothing and just wait for the tool to exist everywhere, but the lack of the sample config file is breaking Homebrew HEAD installation with the current formula.

Rycieos:
Yeah that's not ideal. Just remove the install of the configfile stub for now?

augmentedfourth:
> Yeah that's not ideal. Just remove the install of the configfile stub for now?

That's my plan. I just hope we don't have any problems reintroducing it later.

PR #775, opened 2023-03-27
Use Liquid Prompt as a status bar

nojhan:
This introduces scripts to use Liquid Prompt as a status bar.

First, adds the LP_ENABLE_PATH configuration (because PWD would be useless in a status bar). I am not sure how to test this feature, which is very high level, so there is no test at the moment (ideas welcomed).

Then adds a preset that disable anything tied up to the shell context (path, shell levels, VCSs, etc.) and enable other features (fancy datetime, OS, env vars, etc. Because we should have some room in the status).

Finally, provides scripts to make a status line:
- A script to print the prompt (copy/pasted from the theme preview tool, not sure how we could re-use the code?).
- A script to dock a window, setting the right X11 properties (so far tested with the i3 WM).

screen_lp-status


docs/config.rst
```diff
@@ -739,6 +739,14 @@ Features
 
    .. versionadded:: 2.2
 
+.. attribute:: LP_ENABLE_PATH
+   :type: bool
+   :value: 1
+
+   Display the current working directory.
+
+   .. versionadded:: 2.2
```
Rycieos, line 748: Maybe add a note on this that despite it being a new option, the feature has existed since the beginning?
```suggestion
   .. versionadded:: 2.2
      Before this version, this feature was always enabled.
```

liquidprompt
```diff
@@ -4081,22 +4086,27 @@ _lp_default_theme_activate() {
...
+            else
+                LP_PERM="${LP_COLOR_NOWRITE}${LP_MARK_PERM}${NO_COL}"
+            fi
         fi
-    fi
 
-    local lp_path_format
-    _lp_path_format "$LP_COLOR_PATH" "$LP_COLOR_PATH_LAST_DIR" "$LP_COLOR_PATH_VCS_ROOT" "$LP_COLOR_PATH_SHORTENED" "/" "$LP_COLOR_PATH_SEPARATOR"
+        local lp_path_format
+        _lp_path_format "$LP_COLOR_PATH" "$LP_COLOR_PATH_LAST_DIR" "$LP_COLOR_PATH_VCS_ROOT" "$LP_COLOR_PATH_SHORTENED" "/" "$LP_COLOR_PATH_SEPARATOR"
```
Rycieos, line 4103: I think the check for `LP_ENABLE_PATH` really should be part of `_lp_path_format()`. While it's fine to keep the check at the top of this function, so far the API has been to do config checking inside the data functions.

Rycieos (commented):
Looks fine to me. I can't comment on the i3 specific stuff, but I'm fine with the new enable config option. I only have small nit picks.

liquidprompt
nojhan: Good catch.

Rycieos (approved):

Issue #782, opened 2023-07-05
View the remote as well as the branch when using git


maxulysse:
### Description

I would love to be able to see the remote as well as the branch when developing with git.

### How will this be useful?

I pushed to the wrong master, and that could probably have been avoided if I checked on which remote I was beforehand.
Knowing that already in my prompt, would be amazing

### Example prompt

This is what I have

```bash
· > symbol ± > BRANCH >  ✝ > 
```

I would like:

```bash
· > symbol ± > REMOTE > BRANCH >  ✝ > 
```

Just knowing if `origin` or `upstream` would be tip top.

Rycieos:
Git doesn't have a concept of "current" or default remote, at least not repo wide. It does have that concept per branch, which can be set with `git branch --set-upstream-to` or `git push --set-upstream`. Is that the information you would want?

nojhan:
That would be feasible by parsing `.git/config`, thus being quite fast. Albeit this is a config file, and thus may need associative tables to be a clean code.

PR #784, opened 2023-07-26
feat(VCS): display the remote repository name

nojhan:
- Only if LP_ENABLE_VCS_REMOTE is enabled (disabled by default, as the function parses a file and the output takes some room).
- Introduces LP_MARK_VCS_REMOTE.
- Re-use LP_COLOR_COMMITS and LP_COLOR_COMMITS_BEHIND.

Refs: #782

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -992,6 +992,20 @@ Features
 
    .. versionadded:: 2.1
 
+.. attribute:: LP_ENABLE_VCS_REMOTE
+   :type: bool
+   :value: 0
+
+   Enable the display of the remote repository in the VCS state section.
+
+   If enabled, will display :attr:`LP_MARK_VCS_REMOTE`, followed by the remote
+   repository name.
+
+   If the remote repository has commits not pulled in the local one,
+   the mark will be showed in :attr:`LP_COLOR_COMMITS_BEHIND`.
+   If the local repository has commits not pushed to the remote one,
+   the remote name is shown in :attr:`LP_COLOR_COMMITS`.
+
 .. attribute:: LP_ENABLE_VCS_ROOT
    :type: bool
    :value: 0
@@ -1693,6 +1707,14 @@ Marks
    Mark used to indicate untracked or extra files exist in the current
    repository.
 
+.. attribute:: LP_MARK_VCS_REMOTE
+   :type: string
+   :value: "⭚"
+
+   Mark used to indicate the VCS remote repository name and status.
+
+   See :attr:`LP_ENABLE_VCS_REMOTE`.
+
 .. attribute:: LP_MARK_VCSH
    :type: string
    :value: "|"
@@ -1861,6 +1883,8 @@ Valid preset color variables are:
    Color used to indicate that the current repository has a remote tracking
    branch that has commits that the local branch does not.
 
+   May be used by :attr:`LP_ENABLE_VCS_REMOTE`.
+
 .. attribute:: LP_COLOR_COMMITS
    :type: string
    :value: $YELLOW
@@ -1868,7 +1892,7 @@ Valid preset color variables are:
    Color used to indicate that the current repository has commits on the local
    branch that the remote tracking branch does not.
 
-   Also used to color :attr:`LP_MARK_STASH`.
+   Also used to color :attr:`LP_MARK_STASH` and :attr:`LP_MARK_VCS_REMOTE`.
 
 .. attribute:: LP_COLOR_CONTAINER
    :type: string
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1006,6 +1006,8 @@ Features
    If the local repository has commits not pushed to the remote one,
    the remote name is shown in :attr:`LP_COLOR_COMMITS`.
 
+   .. versionadded:: 2.2
+
 .. attribute:: LP_ENABLE_VCS_ROOT
    :type: bool
    :value: 0
@@ -1023,7 +1025,7 @@ Features
 
    .. _Python: https://docs.python.org/tutorial/venv.html
    .. _Conda: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html
- 
+
    .. versionadded:: 2.
 
 .. attribute:: LP_ENABLE_WIFI_STRENGTH
@@ -1716,6 +1718,8 @@ Marks
    Mark used to indicate the VCS remote repository name and status.
 
    See :attr:`LP_ENABLE_VCS_REMOTE`.
+
+   .. versionadded:: 2.2
 
 .. attribute:: LP_MARK_VCSH
    :type: string
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1026,8 +1026,6 @@ Features
    .. _Python: https://docs.python.org/tutorial/venv.html
    .. _Conda: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html
 
-   .. versionadded:: 2.
-
 .. attribute:: LP_ENABLE_WIFI_STRENGTH
    :type: bool
    :value: 0
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1001,10 +1001,11 @@ Features
    If enabled, will display :attr:`LP_MARK_VCS_REMOTE`, followed by the remote
    repository name.
 
-   If the remote repository has commits not pulled in the local one,
-   the mark will be showed in :attr:`LP_COLOR_COMMITS_BEHIND`.
-   If the local repository has commits not pushed to the remote one,
-   the remote name is shown in :attr:`LP_COLOR_COMMITS`.
+   In the default theme, if the remote repository has commits not pulled in the
+   local branch, the mark will be showed in :attr:`LP_COLOR_COMMITS_BEHIND`. If
+   the local repository has commits not pushed to the remote branch, the remote
+   name is shown in :attr:`LP_COLOR_COMMITS`. If neither is the case, nothing
+   will be shown.
 
    .. versionadded:: 2.2
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1023,6 +1023,8 @@ Features
 
    .. _Python: https://docs.python.org/tutorial/venv.html
    .. _Conda: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html
+ 
+   .. versionadded:: 2.
 
 .. attribute:: LP_ENABLE_WIFI_STRENGTH
    :type: bool
--- a/liquidprompt
+++ b/liquidprompt
@@ -2782,6 +2782,7 @@ _lp_git_active() {
 
 # Get the branch name of the Git repo in the current directory.
 _lp_git_branch() {
+    lp_vcs_branch=""
     local branch ret
     # Recent versions of Git support the --short option for symbolic-ref, but
     # not 1.7.9 (Ubuntu 12.04)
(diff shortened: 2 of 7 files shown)


docs/config.rst
```diff
@@ -992,6 +992,20 @@ Features
...
+
+   Enable the display of the remote repository in the VCS state section.
+
+   If enabled, will display :attr:`LP_MARK_VCS_REMOTE`, followed by the remote
+   repository name.
+
+   If the remote repository has commits not pulled in the local one,
+   the mark will be showed in :attr:`LP_COLOR_COMMITS_BEHIND`.
+   If the local repository has commits not pushed to the remote one,
+   the remote name is shown in :attr:`LP_COLOR_COMMITS`.
```
Rycieos, line 1007: New blocks here are missing `.. versionadded:: 2.1`

liquidprompt
```diff
@@ -2971,6 +2993,48 @@ _lp_git_staged_lines() {
     lp_vcs_staged_d_lines=$lp_git_diff_shortstat_d_lines
 }
 
+# Get the remote name for the current branch.
+# NOTE: Assumes that `_lp_find_vcs` and `_lp_vcs_branch` (costly functions) have been called.
+_lp_git_remote() {
+    lp_vcs_remote=""
```
Rycieos, line 2999: Setting this first is good, so if you exit early it is reset.

The issue is that `_lp_git_branch()` does not do this (my bad). So switching branches will leave `$lp_vcs_branch` set, which means this function doesn't realize no branch is set.

This has never been a problem before since everything always checked the return value of `_lp_git_branch()`, but this cannot. So add a line on line 2791 to set `lp_vcs_branch=""`.

liquidprompt
```diff
@@ -3398,6 +3465,7 @@ _lp_fossil_unstaged_files() { return 2 ; }
 _lp_fossil_unstaged_lines() { return 2 ; }
 _lp_fossil_staged_files() { return 2 ; }
 _lp_fossil_staged_lines() { return 2 ; }
+_lp_fossil_remote() { return 2 ; }
```
Rycieos, line 3468: These Bazzar functions are all wrong. I messed it up when adding these not implemented functions back in 4adbbf2c. If you could fix them as part of this that would be great.
```suggestion
_lp_bzr_remote() { return 2 ; }
```

liquidprompt
```diff
@@ -2971,6 +2993,48 @@ _lp_git_staged_lines() {
...
+        elif [[ "$line" == *"remote${delimiter}"* ]] ; then
+            if [[ "$in_branch" ]]; then
+                # Remove first part until delimiter.
+                lp_vcs_remote="${line#*"${key}${delimiter}"}"
+                return 0
+            else
+                in_branch=0
+            fi
+        fi
+    done <"$file"
```
Rycieos, line 3032: This block clearly does work. But do we need it to? I timed it:
```bash
$ time git config branch.master.remote
origin

real   	0m0.004s
user   	0m0.003s
sys    	0m0.001s

$ time _lp_git_remote

real   	0m0.002s
user   	0m0.001s
sys    	0m0.001s
```
The speed difference isn't that significant. And this isn't a feature enabled by default.

If you want to keep this design, I think we should have some tests to check this nested config object parsing.

```suggestion
    if lp_vcs_remote="$(git config "branch.${lp_vcs_branch}.remote")";
        return 0
    fi
```

Rycieos (changes requested):
Basics look good.

liquidprompt
nojhan: Considering it is still two times faster, I'm leaning toward keeping the more complex implementation. I gave a try to a test in a42f518, but _lp_find_vcs doesn't seem to work in that context (if you have any clue…).

Rycieos: It never loads the directory because the list of enabled VCSs is never compiled. You need to run `lp_activate` to make that happen:
```diff
diff --git a/tests/test_git.sh b/tests/test_git.sh
index 337e59a..bea2155 100644
--- a/tests/test_git.sh
+++ b/tests/test_git.sh
@@ -18,7 +18,10 @@ function test_git {
     LP_ENABLE_VCS_ROOT=0
     LP_ENABLE_VCS_REMOTE=1

-    wd="${SHUNIT_TMPDIR}/test_remote/"
+    PS1=""
+    lp_activate
+
+    wd="${SHUNIT_TMPDIR}/test_remote"
     mkdir -p "$wd"
     cd "$wd"

@@ -46,7 +49,7 @@ function test_git {
     git remote add notsoremote git@nojhan.github:nojhan/liquidprompt.git

     _lp_git_remote
-    assetEquals "Remote is found." "notsoremote." "$lp_vcs_remote"
+    assertEquals "Remote is found." "notsoremote." "$lp_vcs_remote"
 }

 . ./shunit2
```

nojhan:
There was a bug in the function, and tests had to use a local remote on an existing branch.

docs/config.rst
```diff
@@ -1693,6 +1709,14 @@ Marks
...
    repository.
 
+.. attribute:: LP_MARK_VCS_REMOTE
+   :type: string
+   :value: "⭚"
+
+   Mark used to indicate the VCS remote repository name and status.
+
+   See :attr:`LP_ENABLE_VCS_REMOTE`.
+
```
Rycieos, line 1719: Missing version added block.

tests/test_git.sh
```diff
@@ -0,0 +1,59 @@
...
+    cd "$wd"
+
+    _lp_are_vcs_enabled
+    assertTrue "VCS are enabled here." "$?"
+
+    git init
+    # We need a commit to have a branch that can have a remote.
+    touch test
+    git add test
+    git commit -m "test"
```
Rycieos, line 35: This line is causing many problems. In CI, the `git` install is without config, so a commit throws an error because no author is set. Locally, I have commit signing enabled, so the tests halt to ask me for my key password.

Try this:
```suggestion
    git commit -m "test" --author="A U Thor <author@example.com>" --no-verify --no-gpg-sign
```

Rycieos (changes requested):

liquidprompt
```diff
@@ -2971,6 +2994,48 @@ _lp_git_staged_lines() {
...
+        return 1
+    fi
+
+    # If we don't have a VCS root…
+    if [[ -z "$lp_vcs_root" ]]; then
+        # … we cannot find the config file.
+        return 1
+    fi
+
+    local file="${lp_vcs_root}/.git/config"
```
Rycieos, line 3014: I just realized, this is a big problem.

I made this mistake before. Not all Git directories have a `.git/` directory. Some have a `.git` file. See https://git-scm.com/docs/git-worktree for details.

See commit fc44d9724127e6b3bd803be5afed45d64fa566e4 for the fix I made to add this support (after I broke it in the first place), and 1be52e0d048f68db9ab0ffd2f4d2a4212b685f93 for more details.

The point is `$lp_vcs_root` is not a safe way to get to the config file. What we should do instead is use `$lp_vcs_dir`, which is supposed to be the database directory.

I say "supposed to be", because it turns out the `GIT_DIR` for a worktree is actually the `.git/worktrees/<worktree_name>/`.

The problem is we cannot "fix" our detection of `$lp_vcs_dir` to always resolve to the `.git/` directory, because some functions like `_lp_git_head_status()` need to 
read files like `REBASE_HEAD` that exist in the worktree dir.

That said, this seems to be not the only bug of this sort:
https://github.com/liquidprompt/liquidprompt/blob/b0af1a5eff5da520fcb02842aea35caef7e890ad/liquidprompt#L2571
The Git SVN detection is likely broken in Git worktrees.

I think the right solution is to have two Git dirs, the "real" one, and the "working", or "specific" one. Then we can load files from which ever is correct. The trick will be how do we detect the "real" one from a worktree? It looks like a file named `gitdir` exists in the "specific" git dir that contains the path to the "real" one, so we can probably load that.

That was a lot. And this is pretty complicated (and out of scope for this PR), so I can implement it.

liquidprompt
```diff
@@ -2971,6 +2994,48 @@ _lp_git_staged_lines() {
...
+    local in_branch=0
+    # We test for $line in the loop is here to ensure that we read the last line,
+    # even if the file does not ends with a \n.
+    # This bypass a known behavior of the C standard, not fixed in POSIX.
+    while IFS='' read -r line || [[ -n "$line" ]] ; do
+        # If we are in the branch config section.
+        if [[ "$line" == "[branch \"${lp_vcs_branch}\"]" ]]; then
+            in_branch=1
+        elif [[ "$line" == *"remote${delimiter}"* ]] ; then
+            if [[ "$in_branch" ]]; then
```
Rycieos, line 3025: Apparently this is working, but I didn't know that this syntax worked for variables with either `1` or `0`.

liquidprompt
```diff
@@ -2749,6 +2766,11 @@ _lp_vcs_staged_lines() {
     "_lp_${lp_vcs_type}_staged_lines"
 }
 
+# Get the remote name for the current branch.
+_lp_vcs_remote() {
```
Rycieos, line 2770: This function needs docs.

Rycieos (changes requested):
I can't comment on lines that didn't change, so I'll say it here:
* You are missing an empty definition for `_lp_svn_remote()`.

This is pretty much done, other than the large issue of the `lp_vcs_dir`. But I can take care of that, and since the other stuff is pretty small, I can wrap it all up if you like.

liquidprompt
nojhan: What about going back at your first suggested option of just calling `git remote`? Letting Git managing its complex configuration seems a better option given all this.

Rycieos: Yeah, that is an option. Since there is already a bug in Git SVN detection though, I'm thinking that I need to fix that, and once I do that, this should be pretty simple to still do this way.

Rycieos:
@nojhan I pushed two commits to this PR to fix the issues I saw. If you could just take a quick look I can merge this.

nojhan:
PR looks good (I'm trying hard to find something you forgot, like I always do, but I fail miserably).

On the design side, I am now thinking that the remote name display should only occur when there are commits to be pushed or pulled, and not displayed (even with `LP_ENABLE_VCS_REMOTE=1`) when there is no commit. This would be more consistent with LP's general design of displaying information only when needed.

This would probably only require the last control flow section of `_lp_vcs_details_color` to be edited by removing the `else` subsections.

What do you think of the idea?

nojhan:
This would look like:
```bash
    if _lp_vcs_remote; then
        # Commits behind but not ahead.
        if   [[ "$lp_vcs_commit_behind" -ne "0" && "$lp_vcs_commit_ahead" -eq "0"  ]]; then
            lp_vcs_details_color+="${LP_COLOR_COMMITS_BEHIND}${LP_MARK_VCS_REMOTE}${lp_vcs_remote}${NO_COL}"

        # Commits ahead but not behind.
        elif [[ "$lp_vcs_commit_behind" -eq "0" && "$lp_vcs_commit_ahead" -ne "0" ]]; then
            lp_vcs_details_color+="${LP_COLOR_COMMITS}${LP_MARK_VCS_REMOTE}${lp_vcs_remote}${NO_COL}"

        # Commits both ahead and behind: the mark is colored differently.
        elif [[ "$lp_vcs_commit_behind" -eq "0" && "$lp_vcs_commit_ahead" -ne "0" ]]; then
            lp_vcs_details_color+="${LP_MARK_VCS_REMOTE}${LP_MARK_VCS_REMOTE}${LP_COLOR_COMMITS}${lp_vcs_remote}${NO_COL}"

        # else: display nothing.
        fi
    fi
```

Rycieos:
Yeah I like that much more actually. I added bb328314c33779faee0634b49a93b8358681098e with that change.

nojhan:
The related doc may be outdated after this change.

Rycieos:
Good call, I amended commit 98e861e40ba9942b48bd08ed6a6d8d03290c1dce with clarified docs.

liquidprompt
Rycieos: I should have trusted my gut. This is not correct syntax. `[[ $var ]]` is the same as `[[ -n $var ]]`, which is `true` for any nonempty string. So this check always returns `true`. Which means this function always returns the first origin in the config file. Of course both my manual tests and the unit tests did not catch this as we only have the one remote.

I modified the tests to add a second remote to catch this case, and fixed the check in commit 4c22aa97b89140d1678a4c51546f0640290cf0d7.

PR #788, opened 2023-09-01
Fix RAM display on non-Linux hosts

### LINKED ISSUE #787: Memory display on Darwin is inaccurate
With the most recent update to liquidprompt, the following was added to my prompt:
```
⏚ M0% [:~] %
```
Not recognizing `M0%`, I changed `LP_DISPLAY_VALUES_AS_PERCENTS` back to the default of `0`, and it then displayed:
```
⏚ M85.20GiB [:~] %
```
With a number that high, I initially assumed that it was indicating free disk space, but I currently have over 100GB free. After a searching around a bit more, I realized it was supposed to be a memory display (https://github.com/nojhan/liquidprompt/pull/770). 

The issue is that my Macbook Pro only has 16GB of memory. 

Shell: zsh 5.9

Operating system: Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 arm64

Liquid Prompt version: 9a6cff9981323b15bfcdbc7d49720441ea97ce0a

### Steps to Reproduce

1. Macbook Pro (M1), running Ventura 13.5.1
2. Use default liquidprompt confiig
3. No other steps required

### Expected Behavior

Accurate display of memory
### Current Behavior

Inaccurate display of memory
### Possible Solution

No suggestions, but I can hide it by changing the default of `LP_ENABLE_RAM=1` to `LP_ENABLE_RAM=0`. 😄

Rycieos:
A number of small bugs that worked together to make Linux RAM display correct, but MacOS and BSD off by a factor of 1024.

Fixes #787

Note that this was not caught by the tests, as the tests only check what `__lp_ram_bytes()` returns, not what `_lp_ram()` returns. Would be good to have that added @nojhan if you want to do that. Specifically the values of `$lp_ram` and `$lp_ram_human`.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -3795,6 +3795,9 @@ case "$LP_OS" in
                 fi
             # Assumes file exists.
             done < "$_LP_LINUX_RAM_FILE"
+            # Values in file are in KiB.
+            lp_ram_total_bytes=$(( lp_ram_total_bytes * 1024 ))
+            lp_ram_avail_bytes=$(( lp_ram_avail_bytes * 1024 ))
             # Used memory may be used in a theme.
             # shellcheck disable=SC2034
             lp_ram_used_bytes=$((lp_ram_total_bytes-lp_ram_avail_bytes))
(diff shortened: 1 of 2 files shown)


tomhoover:
I reset `LP_ENABLE_RAM=1` with your proposed fix. I'm now seeing `M32.00GiB` in my prompt, even though my MacBook Pro has only 16GB.

tomhoover:
More info:

Pressing `return` multiple times resulted in the following:
```
⏚ M225.17GiB [:~] 1 %
⏚ M47.56GiB [:~] 1 %
⏚ M62.39GiB [:~] 1 %
⏚ M256.98GiB [:~] 1 %
⏚ M218.29GiB [:~] 1 %
⏚ M198.43GiB [:~] 1 %
⏚ M149.60GiB [:~] 1 %
⏚ M134.96GiB [:~] 1 %
⏚ M128.87GiB [:~] 1 %
⏚ M122.40GiB [:~] 1 %
⏚ M110.14GiB [:~] 1 %
⏚ M114.84GiB [:~] 1 %
⏚ M111.43GiB [:~] 1 %
⏚ M108.31GiB [:~] 1 %
```
I'll be happy to run any tests you'd like.

nojhan:
I added tests on `lp_ram_*` and (hopefully) fixed the MacOS mem counter. On my mac it seems to be (more or less) consistent with the activity monitor.

Rycieos:
Looks like your fix was a step in the right direction. I found an issue with adding the two fields, which I fixed. I also fixed the test case. I added my own Mac as a test case, and fixed an issue I discovered on Cygwin.

nojhan (approved):
Seems good, works well on my setups.

PR #792, opened 2023-10-14
Add LP_THEME variable for controlling current theme

### LINKED ISSUE #721: Generic hook
Provide a generic hook that may be used by a (power) user to call configuration change/functions whenever the prompt is called.

For example: change the locale configuration whenever entering a Git directory, read a `.env` file when entering a directory, send notifications, etc.

### Description

The hook takes the form of a normalized function name that may be overloaded in the RC file, and which is called automatically by the prompt if `LP_HOOK` is set.
The hook would be located somewhere near the end of the prompt, to allow its user to display stuff with it.

### How will this be useful?

- change the locale configuration whenever entering some directory,
- more generally, change the environment config when entering a directory or whatever virtual env/container,
- send notifications to the system if such or such threshold is overcome,
- change Liquidprompt’s theme for some directories,
- test simple features before suggesting a PR to LP :-)

Rycieos:
LP_THEME is always set to the currently active theme when changed by lp_theme(). It can also be assigned a theme name to have lp_theme() called automatically. This is useful for interacting with tools like direnv that cannot call shell functions since they exist in a subshell.

Closes #721

### DIFF
--- a/docs/functions/public.rst
+++ b/docs/functions/public.rst
@@ -49,9 +49,9 @@ their config.
    The optional flag ``--list`` will instead list all currently loaded
    themes.
 
-   The variable :envvar:`LP_THEME` can be set to a theme name which functions
-   like calling this function. :envvar:`LP_THEME` will always hold the
-   currently active theme; it is updated by this function. If
+   The variable :envvar:`LP_THEME` can be set to a theme name, which is
+   essentially equivalent to calling this function. :envvar:`LP_THEME` will
+   always hold the currently active theme; it is updated by this function. If
    :envvar:`LP_THEME` is set to an invalid theme name, it will be reset to the
    previous value.
 
--- a/docs/functions/public.rst
+++ b/docs/functions/public.rst
@@ -49,9 +49,18 @@ their config.
    The optional flag ``--list`` will instead list all currently loaded
    themes.
 
+   The variable :envvar:`LP_THEME` can be set to a theme name which functions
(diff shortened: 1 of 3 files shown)


docs/functions/public.rst
```diff
@@ -49,10 +49,19 @@ their config.
    The optional flag ``--list`` will instead list all currently loaded
    themes.
 
+   The variable :envvar:`LP_THEME` can be set to a theme name which functions
+   like calling this function. :envvar:`LP_THEME` will always hold the
```
nojhan, line 53: The first sentence is a bit hard to grasp. What about: "The variable LP_THEME can be set to a theme name, which is essentially equivalent to calling this function."

nojhan (changes requested):
Awesome, works like a charm with direnv!

Shouldn't we have tests for checking if `LP_THEME` is correctly set after calling `lp_theme`, and conversely?

Rycieos:
Good call on tests. I added some, and reworded the docs with your suggestion.

Rycieos:
> I added an FAQ item to recall that: [How to change the configuration depending on the directory?](https://github.com/liquidprompt/liquidprompt/wiki/FAQ#how-to-change-the-configuration-depending-on-the-directory).

You probably don't need `export` in that line, since Liquid Prompt runs in the shell.

nojhan:
Actually, it does not work without `export`. The [direnv man](https://direnv.net/man/direnv.1.html) states that "all exported variables are then captured by direnv and then made available to your current shell".

PR #803, opened 2023-11-05
fix(_lp_fill): avoid costly strip_escape on filling sequence

nojhan:
The previous behavior was to allow escaped sequences in the *fillchars* argument of _lp_fill. This could lead to a very slow call to __lp_strip_escapes if the gap was large.

The new behavior is to remove escaped sequences from fillchars, but allow them in the new arguments *fillprefix* and *fillsuffix*.

This actually simplifies the _lp_fill function a bit.

This patch also removes fixed COLUMNS in theme-preview.sh and fix a spacing problem in the unfold theme.

Ref: [this thread on liquidprompt-powerline](https://github.com/liquidprompt/liquidprompt-powerline/pull/2#pullrequestreview-1711142969)

### DIFF
--- a/docs/functions/util.rst
+++ b/docs/functions/util.rst
@@ -56,7 +56,7 @@ These functions are designed to be used by themes.
 
    See also :attr:`LP_ENABLE_HYPERLINKS` and :func:`_lp_create_link`.
 
-.. function:: _lp_fill(left, right, [fillstring], [splitends]) -> var:lp_fill
+.. function:: _lp_fill(left, right, [fillstring], [splitends, [fillprefix, [fillsuffix]]]) -> var:lp_fill
 
    Adds as much *fillstring* (e.g. spaces) between *left* and *right*,
    so as to make the resulting string the same width as the current terminal.
@@ -72,12 +72,12 @@ These functions are designed to be used by themes.
    some spaces will be inserted after the last occurrence of *fillstring*,
    so as to match the exact width of the terminal.
 
-   *fillstring* may contains escaped sequences (such as colors).
+   *fillstring* may NOT contains escaped sequences (such as colors).
 
-   .. note:: If *fillstring* have multiple characters and *splitends* is 1,
-             then **the last occurrence** of *fillstring*
-             is stripped of any escaped sequences
-             and only its printable characters are inserted.
+  .. note:: Any escaped sequence in *fillstring* will be removed automatically.
+
+   *fillprefix* and *fillsuffix* may be used to add escaped sequences at the
+   beginning and, respectively, the end of the filling sequence.
 
    If the available number of columns in the terminal is smaller than
    the width of *left* and *right* combined, then
--- a/docs/functions/util.rst
+++ b/docs/functions/util.rst
@@ -56,7 +56,7 @@ These functions are designed to be used by themes.
 
    See also :attr:`LP_ENABLE_HYPERLINKS` and :func:`_lp_create_link`.
 
-.. function:: _lp_fill(left, right, [fillstring], [splitends, [fillprefix, [fillsuffix]]]) -> var:lp_fill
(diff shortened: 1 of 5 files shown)


docs/functions/util.rst
```diff
@@ -72,12 +72,12 @@ These functions are designed to be used by themes.
...
+   *fillstring* may NOT contains escaped sequences (such as colors).
 
-   .. note:: If *fillstring* have multiple characters and *splitends* is 1,
-             then **the last occurrence** of *fillstring*
-             is stripped of any escaped sequences
-             and only its printable characters are inserted.
+  .. note:: Any escaped sequence in *fillstring* will be removed automatically.
+
+   *fillprefix* and *fillsuffix* may be used to add escaped sequences at the
+   beginning and, respectively, the end of the filling sequence.
```
Rycieos, line 80: This is confusing. I can't put escape sequences in `fillstring`, but if I do, they will be removed?

I think this would make it more clear:
```rst
   Any escaped sequence in *fillstring* will be removed automatically.
   *fillprefix* and *fillsuffix* may be used to add escaped sequences at the
   beginning and end, respectively, of the filling sequence.
```

Rycieos (changes requested):
This does seem like an improvement. But I don't see the point of `fillprefix` and `fillsufix`. Every time you use them, you concat them together with `left` and `right` respectively. Why not leave off these new params, and simply document that `fillstring` will have any formatting removed, so if you want to format it, add that formatting at the end of `left` (and to reset formatting at the beginning of `right` if that is what you want).

nojhan:
> This does seem like an improvement. But I don't see the point of `fillprefix` and `fillsufix`.

That's definitely right, it seems obvious in retrospect.

tools/theme-preview.sh
```diff
@@ -2,8 +2,6 @@
 #shellcheck disable=SC2317
 set -u
 
-COLUMNS=160
-
```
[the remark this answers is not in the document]
nojhan: Done in PR #805

liquidprompt
```diff
@@ -1254,51 +1254,59 @@ _lp_grep_fields() {
     done <"$file"
 }
 
-# Adds as much $3 character (e.g. spaces) between $1 and $2,
+# Adds as much $3 (fillchars) character(s) (e.g. spaces) between $1 (left) and $2 (right),
 # so as to make the resulting string the same width as the current terminal.
+# If $4 (splitends) is true, then will cut $3 (fillchars) so as to perfectly fill the available space.
+# WARNING: this function may be slow if you put escaped characters in $3 (fillchars)
+#          and the width of the gap to fill is large. Thus, any escaped character is
+#          removed from $3 (fillchars) at the beginning.
```
Rycieos, line 1262: This does not make much sense. If escape sequences are removed, how can it be slow in that case?

liquidprompt
```diff
@@ -1254,51 +1254,59 @@ _lp_grep_fields() {
...
+
     # Compute the gap between left and right.
     local ret
-    __lp_strip_escapes "$left"
+    __lp_strip_escapes "${left}"
     local left_as_text="$ret"
-    __lp_strip_escapes "$right"
+    __lp_strip_escapes "${right}"
     local right_as_text="$ret"
+    # Width of the gap is width of the screen minus every printable characters.
```
Rycieos, line 1281:
```suggestion
    # Width of the gap is width of the screen minus every printable character.
```

Rycieos (changes requested):
Look good. Only a few comment issues.

nojhan:
> Edit: and the tests are failing.

Yes, I got interrupted while working on it (on a separate commit). Hence why I did not re-request for a new review right away :-)

Rycieos (approved):
Looks good.

PR #805, opened 2023-11-11
Add extra options to theme-preview.sh for config

### LINKED ISSUE #800: theme-preview fails when sourcing preset
The `theme-preview.sh` script works well for raw themes, but adding presets to the list of sourced files fails on unbound variables and functions. This does not depends on the order of sourced files.

Shell: Bash 5.1.16(1)-release
Operating system: Linux 6.1.0-1008-oem #8-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar  3 10:51:21 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
Liquid Prompt version: dce7453

### Steps to Reproduce

Run `./tools/theme-preview.sh <mytheme> <path/to/mytheme.theme> <any/preset>`

### Expected Behavior

It should possible to combine themes and presets.

### Current Behavior

Examples:
1. `./tools/theme-preview.sh default liquidprompt contrib/presets/more-unicode.conf` => "contrib/presets/more-unicode.conf: line 43: BLUE: unbound variable"
2. `./tools/theme-preview.sh default contrib/presets/more-unicode.conf` => same
3. `./tools/theme-preview.sh default contrib/presets/more-unicode.conf liquidprompt` => same
4. `./tools/theme-preview.sh unfold themes/unfold/unfold.theme contrib/presets/colors/256-colors-dark.conf` => "./liquidprompt: line 1103: LP_ENABLE_COLOR: unbound variable"
5. `./tools/theme-preview.sh default liquidprompt contrib/presets/colors/green-not-blue.conf` => "contrib/presets/colors/green-not-blue.conf: line 3: __lp_foreground_color: command not found
contrib/presets/colors/green-not-blue.conf: line 4: af_color: unbound variable"

Rycieos:
The original intent for theme-preview.sh was to have a reference for a static output of a theme. So all runs of the tool for a specific theme would always show the same output.

But a different, completely valid use is to see how a theme would look with your config, or to see how changing your config might change how a theme looks.

Add optional arguments that will allow for loading the user config, input files, or to return to the original functionality of locking the `COLUMNS` to a set value.

There is still no way to control the mocked data, and there might still be some weird quirks with the imbedded config in the tool conflicting with user config.

Fix the minimal theme using the wrong `PS1` variable.
Fix Zsh crash in unfold theme.

Fixes #800


tools/theme-preview.sh
```diff
@@ -2,31 +2,84 @@
...
-Example usage: %s powerline themes/powerline/powerline.theme
+Options:
+  --config-file=<filename> Load <filename> as an additional Liquid Prompt
+                           config file.
+  --user-config            Load the user, or if not found, the system config
+                           file.
+  --reproducible           Set the terminal size to a static value. Load no
+                           config files. Overrides any --config-file or
+                           --user-config arguments.
+  -h, --help               Print this help text.
```
nojhan, line 20: I would sort the options in the reverse order, which is going from the simplest to the more complex: first the _help_, then _reproducible_, then _user-config_, then _config-file_.

tools/theme-preview.sh
```diff
@@ -2,31 +2,84 @@
...
+
+Example usage: %s unfold themes/unfold/unfold.theme
```
nojhan, line 22: I think we need at least an example with _reproducible_, and one with config files, to show how to load presets.

nojhan:
```suggestion
Example usage:
%s alternate_vcs themes/alternate_vcs/alternate_vcs.theme
%s unfold themes/unfold/unfold.theme --reproducible
%s unfold themes/unfold/unfold.theme --user-config --config-file contrib/presets/colors/256-colors-dark.conf
%s 
```

nojhan (changes requested):
Awesome, that's going to be useful.

I tested with several themes and presets and it works well, except for the `minimal` theme, which ends on missing LP_PS1_PREFIX.

tools/theme-preview.sh
nojhan: And maybe indicate here that presets are config files?

Rycieos:
> it works well, except for the `minimal` theme, which ends on missing LP_PS1_PREFIX.

So the `minimal` theme is actually not a theme, but instead a "[default theme template](https://liquidprompt.readthedocs.io/en/stable/theme/default.html#templates)" (aka a Liquid Prompt v1 theme). To use such a template, you need to set [LP_PS1_FILE](https://liquidprompt.readthedocs.io/en/stable/theme/default.html#LP_PS1_FILE) to its path.

Likely it should not have been under `themes/`, since that is just confusing. Maybe we move it to a `templates/` directory?

It does leave the UX lacking though. You can load a template like this with an inline config file:
```bash
./tools/theme-preview.sh default --config-file <(echo LP_PS1_FILE=themes/minimal/minimal.ps1)
```
But maybe we add an argument, `--template-file`, that will do that substitution for you?

nojhan:
> But maybe we add an argument, `--template-file`, that will do that substitution for you?

I guess that would be the clearest option for a naive user, indeed.

Rycieos:
This is ready for review. The failed docs lint is because I moved the `minimal.ps1` file, and the docs are looking for the new location on the `master` branch.

tools/theme-preview.sh
```diff
@@ -2,30 +2,104 @@
...
+  -h, --help                 Print this help text.
+  --reproducible             Set the terminal size to a static value. Load no
+                             config files. Overrides any --config-file or
+                             --user-config arguments.
+  --user-config              Load the user, or if not found, the system config
+                             file.
+  --template-file=<filename> Load <filename> as a default theme template.
+                             Implies "default" theme.
+  --config-file=<filename>   Load <filename> as an additional Liquid Prompt
+                             config file.
```
nojhan, line 22: This is nitpicking, but I would add "[…] config file (e.g. presets)." to be crystal clear.

tools/theme-preview.sh
```diff
@@ -2,30 +2,104 @@
...
+
+Example usage:
+%s alternate_vcs themes/alternate_vcs/alternate_vcs.theme
+%s unfold themes/unfold/unfold.theme --reproducible
+%s unfold themes/unfold/unfold.theme --user-config --config-file contrib/presets/colors/256-colors-dark.conf
+%s default --template-file templates/minimal/minimal.ps1
+' "$0" "$0" "$0" "$0" "$0"
+}
```
nojhan, line 30: Shouldn't usage messages be output on stderr? I am always hesitating about that.

nojhan (approved):
Looks good, just two minor (and optional) questions.

nojhan (approved):

PR #815, opened 2023-12-04
Added tests for rvm, with my newly acquired knowledge from the Perl i…

jonasbn:
Hi,

I have added tests for rvm, with my newly acquired knowledge from the Perl implementation

Attempting to skip test if rvm and relevant tooling is not available.

Looking forward to feedback.

# Technical checklist:

    - [ ] correct use of `IFS`
    - [ ] careful quoting
    - [ ] cautious array access
    - [ ] functions/variable naming conventions
    - [ ] functions have local variables
    - [X] unit tests have been updated (see `tests/test_*.sh` files)
    - [X] ran `tests.sh`

### DIFF
--- a/tests/test_ruby.sh
+++ b/tests/test_ruby.sh
@@ -13,8 +13,9 @@ fi
 PS1="$ "
 
 LP_ENABLE_RUBY_VENV=1
+LP_RUBY_RVM_PROMPT_OPTIONS=("i" "v" "g" "s")
 
-typeset -a rbenv_outputs rbenv_return_strings
+typeset -a rbenv_outputs rbenv_return_strings rvm_outputs rvm_return_strings
 
 # Add test cases to these arrays like below
 
@@ -24,6 +25,14 @@ rbenv_outputs+=(
 )
 rbenv_return_strings+=(
 "1.9.3-p327"
+)
+
+# Example from rvm documentation
+rvm_outputs+=(
+"ruby-3.2.2"
+)
+rvm_return_strings+=(
+"ruby-3.2.2"
 )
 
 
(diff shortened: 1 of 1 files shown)


tests/test_ruby.sh
```diff
@@ -43,4 +52,24 @@ function test_rbenv {
   done
 }
 
+function test_rvm {
+
+  # Disable test if rvm (and rvm-prompt) is not installed
+  [ -z $(command -v rvm-prompt >/dev/null) ] && startSkipping
```
Rycieos, line 58: This is not particularly useful. Our CI test runs, and my own machine, do not have `rvm-prompt` installed, so this test will rarely be run. Instead of mocking the output of `rvm`, mock `rvm-prompt`, so this test can run without it.

Rycieos (changes requested):

tests/test_ruby.sh
jonasbn: The skipping strategy is often use in Perl distributions, so it was natural for me to go down that path. I have removed the conditional and added a basic mock, let me know what you think.

tests/test_ruby.sh
```diff
@@ -43,4 +52,25 @@ function test_rbenv {
...
 
+function test_rvm {
+
+  rvm-prompt() {
+    printf '%s\n' "ruby-3.2.2"
+  }
+
+  rvm() {
+    printf '%s\n' "$__rvm_output"
+  }
```
Rycieos, line 63: `liquidprompt` never calls `rvm`. So if your `rvm-prompt` mock doesn't either, then there is no point in mocking `rvm`.

Also, the `rvm-prompt` mock shouldn't be hard coded. We have the array of inputs, use that.

```suggestion
  rvm-prompt() {
    printf '%s\n' "$__rvm_output"
  }
```

tests/test_ruby.sh
```diff
@@ -43,4 +52,25 @@ function test_rbenv {
...
+
+  _LP_RUBY_VENV_PROGRAM=rvm
+
+  for (( index=0; index < ${#rvm_outputs[@]}; index++ )); do
+    __rvm_output=${rvm_outputs[$index]}
+
+    _lp_ruby_env
+    assertEquals "rvm returns at index ${index}" "${rvm_return_strings[$index]}" "$lp_ruby_env"
```
Rycieos, line 71: Might as well be clear
```suggestion
    assertEquals "rvm-prompt returns at index ${index}" "${rvm_return_strings[$index]}" "$lp_ruby_env"
```

Rycieos (changes requested):

tests/test_ruby.sh
jonasbn: Thanks @Rycieos

Rycieos (changes requested):
Looks good, if you wouldn't mind quashing to one commit I will merge.

jonasbn:
Hi @Rycieos 

Squash completed, thanks for the help

jonasbn:
Thank you @Rycieos for accepting my mediocre contributions. I learned a lot from your assistance.

PR #817, opened 2023-12-19
Do not scan through all PIDs to find parent PID command

Fil0sOFF:
There's one command that takes a long time after profiling bash startup. 
We currently scan though all processes just to find out $PPID command.
The time to do it grows linearly with a number of processes running.
Let's fix that.

Compare:
```bash
strace -e openat ps -o comm= -p $PPID
```
vs
```bash
strace -e openat ps -o comm= -q $PPID
```
# Technical checklist:

    - [x] correct use of `IFS`
    - [x] careful quoting
    - [x] cautious array access
    - [x] ran `tests.sh`

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1641,7 +1641,7 @@ _lp_connection() {
         return
     fi
 
-    local sess_parent="$(ps -o comm= -p "$PPID" 2> /dev/null)"
+    local sess_parent="$(ps -o comm= -q "$PPID" 2> /dev/null)"
     if [[ "$sess_parent" = "su" || "$sess_parent" = "sudo" ]]; then
         lp_connection=su   # su/sudo
     else
--- a/tools/external-tool-tester.sh
+++ b/tools/external-tool-tester.sh
@@ -69,7 +69,7 @@ test_tool tput AB 0
 test_tool tput AB 0 0 0
 
 test_tool who am i
-test_tool ps -o comm= -p "$PPID"
+test_tool ps -o comm= -q "$PPID"
 test_tool logname
 
 test_tool screen -ls


Rycieos (changes requested):
`ps` does not support the `-q` flag on MacOS. This can't be merged as is.

Can you post some timing info (`time ps -o comm= -p "$PPID"`) from the situations that you think this is slow?

If this is really a problem, we could try to have two different paths for those that support the `-q` flag vs those that don't.

Fil0sOFF:
Oops, `-q` seems to be available on only on linux,  since [2014](https://gitlab.com/procps-ng/procps/-/commit/e751606fccdeed7ab79fc4324938b3aa47989842)

Rycieos:
I'm probably missing something, but why was a separate option added? If `-q` only is allowed when no other filters are specified, why not just use the optimized search when `-p` is specified but nothing else is.

Fil0sOFF:
I'm not sure why `ps` works that way. I only found about `-q` from a manpage, which i opened after `strace ps -o comm= -p $PPID` showed a long list of /proc/xx files.

`-p` is not that slow most of the time, but it can be in some cases.
```
$ time ps -o comm= -p "$PPID"
wezterm-gui

real    0m0.054s
user    0m0.026s
sys     0m0.031s


$ time ps -o comm= -q "$PPID"
wezterm-gui

real    0m0.012s
user    0m0.012s
sys     0m0.002s

```

after running 1000 `sleep`s:

```
$ for i in `seq 1000`; do sleep 30 & done

$ time ps -o comm= -p "$PPID"
wezterm-gui

real    0m0.104s
user    0m0.058s
sys     0m0.048s

$ time ps -o comm= -q "$PPID"
wezterm-gui

real    0m0.012s
user    0m0.011s
sys     0m0.003s
```


If different code paths for different systems are possible, then `var=$(<file)` might be the fastest option for linux. No forking, no external command execution. Bash- and zsh-compatible

Rycieos:
> If different code paths for different systems are possible, then `var=$(<file)` might be the fastest option for linux. No forking, no external command execution. Bash- and zsh-compatible

What file are you proposing to read? `/proc/$PPID/comm`? That should work, and I agree, it would be faster.

If you want to try to implement it, here are a few tips:
* Read with a command like `IFS="" read -r sess_parent <"$_LP_LINUX_COMM_FILE"`. Make sure to set `local sess_parent` first.
* Use the var for the file path to allow for mocking it in tests. See https://github.com/liquidprompt/liquidprompt/blob/046d8300b2e65337fcecc0f76add818b5f461fcf/liquidprompt#L608-L609 for where those are defined.
* Likely split line 1644 into a function, like `__lp_sess_parent() { sess_parent="$(ps -o comm= -p "$PPID" 2> /dev/null)"; }`. Then you can put the above in a different function.
* Don't define both functions, define one based on the OS at load time. See https://github.com/liquidprompt/liquidprompt/blob/046d8300b2e65337fcecc0f76add818b5f461fcf/liquidprompt#L3589-L3591 for an example. Set the local before you call that function.
* Make sure to add reading from this file to the external tool tester, even though it likely won't be very useful.

cgzones:
This approach might be obsolete with the next version of ps, see https://gitlab.com/procps-ng/procps/-/commit/7f80362b9ff8cb2b699a01e06a70972a4ae9399f

Rycieos:
Nice! Thanks for making that PR, and then bringing it to our attention. Hopefully it will propagate to distros soon.

PR #824, opened 2024-03-04
Avoid running __lp_set_prompt multiple times

steelman:
Unlike lp_activate lp_theme hasn't cleand PROMPT_COMMAND. Thus, __lp_set_prompt was dded each time a theme was switched. In case of a single theme switch (e.g. from the default to dotmatrix) the performance hit due to running __lp_set_prompt twice wasn't noticable. However, the second time the function was called, it didn't show the last command runtime properly even if it was longer than LP_RUNTIME_THRESHOLD.

Fixes: 3553a65 ("Add support for Bash 5.1+ PROMPT_COMMAND array")

# Technical checklist:

    - the code has been copied from 3553a65
    - [x] ran `tests.sh`
    - no — this is a bugfix

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -4913,8 +4913,10 @@ prompt_on() {
 
             if (( ${BASH_VERSINFO[0]:-0} > 5 || ( ${BASH_VERSINFO[0]:-0} == 5 && ${BASH_VERSINFO[1]:-0} >= 1 ) )); then
                 # PROMPT_COMMAND is an array since bash 5.1
-                PROMPT_COMMAND+=( "$set_prompt_command" )
-
+                if ! ( __lp_array_contains __lp_set_prompt ${PROMPT_COMMAND[@]+"${PROMPT_COMMAND[@]}"} \
+                  || __lp_array_contains "time __lp_set_prompt" ${PROMPT_COMMAND[@]+"${PROMPT_COMMAND[@]}"} ); then
+                   PROMPT_COMMAND+=( "$set_prompt_command" )
+                fi
             else
                 if [[ -z ${LP_OLD_PROMPT_COMMAND+x} ]]; then
                     LP_OLD_PROMPT_COMMAND="${PROMPT_COMMAND-}"


liquidprompt
```diff
@@ -4907,8 +4907,13 @@ prompt_on() {
 
             if (( ${BASH_VERSINFO[0]:-0} > 5 || ( ${BASH_VERSINFO[0]:-0} == 5 && ${BASH_VERSINFO[1]:-0} >= 1 ) )); then
                 # PROMPT_COMMAND is an array since bash 5.1
+                for i in ${PROMPT_COMMAND[@]+"${!PROMPT_COMMAND[@]}"}; do
+                    local value="${PROMPT_COMMAND[i]}"
+                    if [[ $value == "__lp_set_prompt" || $value == "time __lp_set_prompt" ]]; then
+                        unset 'PROMPT_COMMAND[i]'
+                    fi
+                done
                 PROMPT_COMMAND+=( "$set_prompt_command" )
```
Rycieos, line 4916: I think this is a better way to fix it, as it follows the other behavior in this function instead of the behavior of `__lp_disable_hooks()`:
```suggestion
                if ! ( __lp_array_contains __lp_set_prompt ${PROMPT_COMMAND[@]+"${PROMPT_COMMAND[@]}"} \
                  || __lp_array_contains "time __lp_set_prompt" ${PROMPT_COMMAND[@]+"${PROMPT_COMMAND[@]}"} ); then
                    PROMPT_COMMAND+=( "$set_prompt_command" )
                fi
```

Rycieos (changes requested):
Thanks for the report and patch. I can confirm the bugged behavior. I made one suggestion.

liquidprompt
steelman: I am new here and I can only guess, but I believe  `__lp_set_prompt` should be the last (as "last" as possible) entry in the `PROMPT_COMMAND` array. If for nothing else than at least to measure the command runtime which IMHO should include anything and everything between the `[⤶]` key being pressed and the prompt being printed. The loop that removes `__lp_set_prompt` from `PROMPT_COMMAND` assures that behavior.

Rycieos: > but I believe `__lp_set_prompt` should be the last

Usually, yes. But in a `PROMPT_COMMAND` that is an array, Bash will treat each command as the same, passing each of them the same error code.

> `__lp_set_prompt` MUST be at the end of the `PROMPT_COMMAND` array. Otherwise it would report runtime improperly.

There actually is an argument for both the end and the beginning:
* At the beginning, you are timing _only_ the actual command you ran.
* At the end, you include any other commands ran as part of `PROMPT_COMMAND`.

I would think most people would want the former. But I see valid reasons for the later.

> (as "last" as possible, Maybe there should be a note somewhere in the `README` to place `lp_theme` call at the end of `.bashrc`)

We have such a warning in the [install section of the docs](https://liquidprompt.readthedocs.io/en/stable/install.html#shell-installation). Well, for the `source ~/liquidprompt/liquidprompt` call, not the `lp_theme` call. But I am arguing that the `lp_theme` call shouldn't change the `PROMPT_COMMAND` anyway.

All that to say, to keep things consistent, `lp_theme` and `prompt_on` should NOT change `PROMPT_COMMAND` if `__lp_set_prompt` is already in it.

steelman: > There actually is an argument for both the end and the beginning:

No, when `__lp_set_prompt` is at the beginning (I've checked only with two commands in `PROMPT_COMMAND`) the time it reports includes the time a user spent _typing_ the command since the previous one finished. It makes little sense to report it this way.

> I would think most people would want the former. But I see valid reasons for the later.

I see your point, but (if I haven't messed up anything) I am sure it can't work this way now anyway.

> `lp_theme` and `prompt_on` should NOT change `PROMPT_COMMAND` if `__lp_set_prompt` is already in it.

It makes perfect sense when you put it that way. I'll apply the above change and investigate the improper time reporting issue.

Rycieos: > No, when `__lp_set_prompt` is at the beginning (I've checked only with two commands in `PROMPT_COMMAND`) the time it reports includes the time a user spent _typing_ the command since the previous one finished.

I was inclined to disagree with you, but after testing it clearly does behave that way. That is a bug. Let me see if I can fix it, because relying on `__lp_set_prompt` to always be the last thing in the array is prone to other problems.

Rycieos: I just fixed that bug in commit 6f2ca7e62d0281356bd984957a9c2da41dd228d6. Thanks for reporting it.

Hopefully now it does not matter where in `PROMPT_COMMAND` `__lp_set_prompt()` is.

steelman: I have applied your suggestion.

BTW can you explain this idiom `${PROMPT_COMMAND[@]+"${PROMPT_COMMAND[@]}"`. I've seen it here and there and I quite understand why would something like _if the value is set use the value_ be any different than _use the value_. I expect there is a good reason I simply don't know.

Rycieos: See [this section of the wiki](https://github.com/liquidprompt/liquidprompt/wiki/Shell-standards#arrays). 

Basically it boils down to needing to support `set -u` mode as well as all Bash versions 3.2+.

Rycieos (approved):

PR #826, opened 2024-03-20
Fix wrong redirection for some error messages

slowpeek:
Evidently,  some `echo "error message" 2>&1` was supposed to be `.. >&2`


Rycieos (approved):
Nice catch, clearly I spaced out when working on v2.0.

PR #827, opened 2024-03-21
Avoid showing the 130 error code after hitting ^C

steelman:
When a user hits ^C on command line to cancel entering a lengthy command, $! in __lp_set_prompt is set to 130 (128 + SIGINT). Displaying this error on the input line (especially in warning colours) makes little sense. Therefore, liquidprompt should check if the command submitted is a real command or is it part of PROMPT_COMMAND in which case the error does not come from a user's command and should not be presented.

# Technical checklist:

    - [ ] correct use of `IFS`
    - [x] careful quoting
    - [x] cautious array access
    - [x] ran `tests.sh`

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -486,6 +486,7 @@ __lp_source_config() {
     LP_ENABLE_OS_VERSION=${LP_ENABLE_OS_VERSION:-1}
     LP_ENABLE_HYPERLINKS=${LP_ENABLE_HYPERLINKS:-0}
     LP_ENABLE_PATH=${LP_ENABLE_PATH:-1}
+    LP_HIDE_EMPTY_ERROR=${LP_HIDE_EMPTY_ERROR:-1}
 
     LP_MARK_DEFAULT="${LP_MARK_DEFAULT:-$_LP_MARK_SYMBOL}"
     LP_MARK_BATTERY="${LP_MARK_BATTERY:-"⌁"}"
@@ -834,6 +835,7 @@ lp_activate() {
         LP_ENABLE_RUNTIME \
         || LP_ENABLE_RUNTIME_BELL \
         || LP_ENABLE_TITLE_COMMAND \
+        || LP_HIDE_EMPTY_ERROR \
     ))
 
     # LP_ENABLE_RUBY_VENV depends either from rvm or rbenv. Thus we cannot
@@ -4810,6 +4812,11 @@ __lp_set_prompt() {
     # As this get the last returned code, it should be called first
     local -i lp_error="$?"
 
+    if (( LP_HIDE_EMPTY_ERROR )); then
+        (( ! ${_LP_REAL_COMMAND:-1} )) && lp_error=0
+        _LP_REAL_COMMAND=0
+    fi
+
     if (( LP_ENABLE_RUNTIME || LP_ENABLE_RUNTIME_BELL )); then
         __lp_runtime_after
     fi
@@ -4857,6 +4864,9 @@ __lp_preexec() {
     fi
     if (( LP_ENABLE_TITLE_COMMAND )); then
         __lp_print_title_command
+    fi
+    if (( LP_HIDE_EMPTY_ERROR )); then
+        _LP_REAL_COMMAND=1
     fi
 }
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1174,6 +1174,18 @@ Features
 
    .. versionadded:: 2.1
 
+.. attribute:: LP_HIDE_EMPTY_ERROR
+   :type: bool
+   :value: 1
+
+   Hide the error code returned by a command if the new prompt is displayed
+   after the user hits Ctrl-C or submits an empty command (i.e. empty string
+   or a comment).
+
+   See also: :attr:`LP_ENABLE_ERROR`.
+
+   .. versionadded:: 2.2
+
 .. attribute:: LP_HOSTNAME_ALWAYS
    :type: int
    :value: 0
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -1,5 +1,6 @@
 Archlinux
 CMake
+Ctrl
 Cygwin
 DragonFly
 FreeBSD


liquidprompt
```diff
@@ -4797,11 +4797,18 @@ _lp_default_theme_prompt() {
...
+_LP_NOT_A_REAL_COMMAND=0
 __lp_set_prompt() {
     # Display the return value of the last command, if different from zero
     # As this get the last returned code, it should be called first
     local -i lp_error="$?"
 
+    if (( _LP_NOT_A_REAL_COMMAND )); then
+        lp_error=0
+    fi
+    # zsh doesn't run __lp_before_command, so we set it here
```
Rycieos, line 4809: Bash also doesn't run `__lp_before_command()` if all the features that depend on it are disabled.

liquidprompt
```diff
@@ -4797,11 +4797,18 @@ _lp_default_theme_prompt() {
 # Construct the prompt #
 ########################
 
+_LP_NOT_A_REAL_COMMAND=0
 __lp_set_prompt() {
     # Display the return value of the last command, if different from zero
     # As this get the last returned code, it should be called first
     local -i lp_error="$?"
 
+    if (( _LP_NOT_A_REAL_COMMAND )); then
```
Rycieos, line 4806: Since it is possible for this to false flag on Bash (if the command that a user ran also existed in `PROMPT_COMMAND`), this should have an extra check. How about `&&` this with `$lp_error == 130`. There still is a chance for a false flag (if the command that a user ran also existed in `PROMPT_COMMAND`, then they `Ctrl^C`ed it), but it would be very unlikely.

Rycieos (changes requested):
Interesting idea. I can see why one would want this. And I likely will use this once we merge it. However, it needs to be configurable.

* Add a config option to control this, like `LP_HIDE_EMPTY_ERROR`. Don't forget to add docs for it as well.
* The tracking variable `_LP_NOT_A_REAL_COMMAND` is very confusing. Rename it to `_LP_REAL_COMMAND`, and invert the Boolean logic.

liquidprompt
steelman: There seem to be another way to detect a real command: check if `$BASH_COMMAND` match `$(builtin builtin history 1)` ( `__lp_get_last_command_line()` but without the fallback) rather than _not_ match `$PROMPT_COMMAND`. What do you think?

Rycieos: The problem is anything with `$( ... )` syntax creates a subshell, which will be a `fork()` system call. That is slow (relatively), and we want to avoid slowing down prompt generation as much as possible.

steelman: Indeed. I was trying to figure something out but I am afraid checking `PROMPT_COMMAND` is the best we can get.

Regarding the `$lp_error == 130` condition — it isn't desired because of the following scenario:

1. Run a real command that returns an error other than 130 (e.g. `/bin/false`)
2. The error indicator appears.
3. Press [Enter]

Actual result:
The error indicator keeps appearing.

Expected result:
The error indicator disappears.

Rycieos: I don't think that condition would happen if you `&&` it with `! _LP_REAL_COMMAND`. I will admit I did not test it.

steelman: I am checking this at the moment and it works as I predicted. Upon entering `__lp_set_prompt()`:
1. `$?` (and `lp_error`) is 1
2. `_LP_REAL_COMMAND_` is 0
3. `LP_HIDE_EMPTY_ERROR` is 1

Thus, the condition `(( ! _LP_REAL_COMMAND && LP_HIDE_EMPTY_ERROR && $lp_error == 130 ))` evaluates to false and `lp_error` is still 1 and shows up at the prompt.

I am pushing a commit without the `lp_error` check. You can add it and see for yourself.

Rycieos: Ah, I misread your previous comment. I see now what you mean.

That actually already happens. Run your test on the `master` branch and you will see that is nothing new. I think Bash and Zsh do not change the value of `$?` if you hit Enter on an empty prompt. If that is the case, there is not much we can do to fix it. And bonus, once this PR is merged, you can Ctrl^C to remove it if you want. If it is a problem in Liquid Prompt somehow, I'd be happy to fix it. But I'm pretty confident it's not; try this minimal PS1: `PS1='($?) \$ '`:

```
$ PS1='($?) \$ '
(0) $ false
(1) $
(1) $
```

So adding the ` && lp_error == 130` to the check is still the correct logic here.

steelman: > That actually already happens

Yes and as far as I am concerned I don't like this behavior (I consider it invalid UI-wise). I don't assume I am right, this is just my opinion, but this is also what I'd like to change in this commit. My idea is: the error indicator is visible iff the last typed _real_ command returned an error. Neither of the following is a _real_ command (and they don't set `_LP_REAL_COMMAND`) in my understanding:

- an empty line
- a comment (a command starting with `#`, which also preserves an old value of `$?`)
- Ctrl-C

Therefore a prompt displayed after any of these events IMHO should not indicate an error.

> there is not much we can do to fix it

Yes. we can. Without the `&& $lp_error == 130` condition and with `_LP_REAL_COMMAND` set to 0 in all three above cases we can hide the _stale_ error code from users by setting `lp_error`  back to 0.

```sh
    local -i lp_error="$?"

    if (( ! _LP_REAL_COMMAND && LP_HIDE_EMPTY_ERROR )); then
        lp_error=0
    fi
```
At this point `$?` ma still be non-zero but we use `lp_error` to construct the prompt.

> And bonus, once this PR is merged, you can Ctrl^C to remove it if you want

That is because
1. ^C changes `$?` to 130
2. With `$lp_error == 130` check `lp_error` is set to 0 (but `$?` is still 130)

> If it is a problem in Liquid Prompt somehow, I'd be happy to fix it

No, it is not a problem in Liquid Prompt and thanks to `lp_error` being used to build prompt we can (by avoiding the `$lp_error == 130` condition) hide errors if they are not a result of a _real_ command.

But once again, this is my opinion. I am trying hard to explain it and back it up, but it is still an opinion. If you say we hide _only_ 130, I will add the condition.

Rycieos: Ok, I think I understand you now. So you want two separate enhancements:
1. If [Enter] on empty or empty-with-comment line: clear previous error code.
2. If Ctrl^C on prompt (not on running command): hide error code.

They are related, obviously. And they are both valid. I think they are closely related enough that they can both go in the same PR.

The only remaining question is if both should be covered by the same config option. I think it is valid to do so, but the docs will need to be _very_ explicit in what scenarios it will hide the error code.

With all that in light, I'll do an updated review.

Rycieos (changes requested):
I was going to post a detailed review of how you missed a couple important things, but the more I dug and reread, I realized that you are fighting against a terrible codebase, which is my fault of course.

So I decided to fix things up as I saw them. Take a look at my new branch [rework/preexec](https://github.com/liquidprompt/liquidprompt/tree/rework/preexec), specifically the first two commits eb334b2e0293a3ed9670bc399f8395519d74f0fd and 436b1b58a44ed8bf4210b6b955d7da6cf707470f. This largely reworks and improves how preexecs are handled, no longer runs them on empty prompts ([Enter]), and also fixes a related issue where runtime display was not cleared after an empty prompt.

After those fixes, your change is super simple, so I did that too in commit 1be449efd380c998fa23a150e1c683371210c525. Please take a look at these 3 commits, and let me know if everything makes sense. If so, I'll merge my first 2 into `master`, and you can continue from there.

If you accept my (our) commit 1be449efd380c998fa23a150e1c683371210c525, then you pretty much are only missing docs. You will want to add a section for your new config option [here](https://github.com/liquidprompt/liquidprompt/blob/master/docs/config.rst?plain=1#L1176). Remember to detail both scenarios this will trigger.

steelman:
LGTM +1

Rycieos (changes requested):
Looks great, except for one "spelling error":

> config.rst:1182: : Spell check: **Ctrl**: after the user hits Ctrl-C or submits an empty command (i.e. empty string.

You will need to add `Ctrl` to the `docs/spelling_wordlist.txt` file.

Rycieos (approved):

rpatterson:
I think this PR may have broken `LP_ENABLE_ERROR=1` for all exit status codes, not just `130`, @steelman. After updating to `v2.2.0`, I discovered that commands that exit with non-zero status were not being reported in the subsequent prompt. It has the same behavior if I checkout [the commit that merged this PR](https://github.com/nojhan/liquidprompt/commit/1b87fbd0922c5938baa5ee2ff3c92361560b19dc). If I checkout [the commit immediately preceding the merge](https://github.com/liquidprompt/liquidprompt/commit/af0a69ac260ae3278e9b449acb5c4ebf7e5bfef4) of PR #827, it works as it did before, reporting both 130 and all other non-zero exit status codes.

LMK if you prefer I report this as a separate issue and I'll copy this over. Here are my technical details:

```
$ echo $BASH_VERSION
5.1.16(1)-release
$ uname -a
Linux ... 6.8.0-76060800daily20240311-generic #202403110203~1715181801~22.04~aba43ee SMP PREEMPT_DYNAMIC Wed M x86_64 x86_64 x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID: Pop
Description:    Pop!_OS 22.04 LTS
Release:        22.04
Codename:       jammy
$ git rev-parse HEAD
1b87fbd0922c5938baa5ee2ff3c92361560b19dc
```

rpatterson:
FWIW, I also preferred the previous behavior. IOW, I prefer reporting `130` even when I cancel a long command with an interrupt. It's useful when scrolling up and reviewing my command history because it can be easy to miss the `^C`.

Rycieos:
> LMK if you prefer I report this as a separate issue and I'll copy this over.

I would prefer that, please do open a new issue.

rpatterson:
> I would prefer that, please do open a new issue.

Done.

PR #829, opened 2024-04-15
Remove superfluous / duplicate squared brakets from AWS profile definition

milouse:
The upcoming 2.2.0 release introduces new mechanism to either join or display individual dev env variables. At one point one removed the various hardcoded enclosing squared brackets from the various colored definition of each dev env variable to allow to join them, but miss the AWS profile one.

This may come from the fact the AWS functions were a bit misplaced in the file. Thus the first commit of this MR move back the code (without modifying it) up in the file, alongside other dev env definition. Then the second commit remove the duplicate brackets.

# Technical checklist:

    - [X] correct use of `IFS`
    - [X] careful quoting
    - [X] cautious array access
    - [X] portable array indexing with `_LP_FIRST_INDEX`
    - [X] printing a "%" character is done with `_LP_PERCENT`
    - [X] functions/variable naming conventions
    - [X] functions have local variables
    - [X] data functions have optimization guards (early exits)
    - [X] subshells are avoided as much as possible
    - [X] string substitutions may be done differently in Bash and Zsh (use `_LP_SHELL_*`)
    - [X] unit tests have been updated (see `tests/test_*.sh` files)
    - [X] ran `tests.sh`
    - [X] ran `shellcheck.sh` (requires [shellcheck](https://github.com/koalaman/shellcheck#user-content-installing)).
    - [X] functions and attributes are documented in alphabetical order
    - [X] new sections are documented in the default theme
    - [X] tag `.. versionadded:: X.Y` or `.. versionchanged:: Y.Z`
    - [X] functions signatures have arguments, returned code, and set value(s)
    - [X] attributes have types and defaults
    - [X] ran `docs/docs-lint.sh` (requires Python 3 and `requirements.txt`

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -2151,7 +2151,6 @@ _lp_software_collections_color() {
     lp_software_collections_color="${LP_COLOR_VIRTUALENV}${lp_software_collections}${NO_COL}"
 }
 
-
 _lp_kubernetes_context() {
     (( LP_ENABLE_KUBECONTEXT )) || return 2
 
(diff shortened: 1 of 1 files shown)


Rycieos (approved):
Nice catch, thanks.

Issue #835, opened 2024-05-27
[contrib] Vim & Neovim assets to help editing colors in theme files


tkapias:
### Description

I would like to submit some files for the "contrib/colors/vim" folder.

When writing, reading or editing a theme file for Liquid Prompt, there is no out-of-the-box solution, in any editor, to get a visual hint on used colors. The format is hard to read: `lp_terminal_format 33 -1 0 0 14 -1`, and contain both fg and bg.

There is a nice plugin for Vim and Neovim that lets the user write custom palettes with matching patterns to highlight specific colors in the editor: [vim-hexokinase](https://github.com/RRethy/vim-hexokinase)

To get vim-hexokinase to highlight theme files I wrote 2 palettes files, for foreground and background, I put them and also some instruction for vim8+ in a Gist [here](https://gist.github.com/tkapias/a419be5b70722ea7021704a37310cfa7).

### How will this be useful?

Advanced users will probably be happy to get a visual help when editing themes.

### Example

Note like the last highlight at the bottom include both foreground and background colors. 

![image](https://github.com/liquidprompt/liquidprompt/assets/45816387/aac6d814-f8f0-4356-b352-f36af7f19b85)

Rycieos:
Neat! I use Vim, so I'd love to get this working, personally.

Feel free to make a PR putting this in `contrib/vim/colors`.

I'm not a fan of the thousands of lines of JSON files, but I'm guessing that is a limitation of the Vim plugin. Which is archived? And it doesn't support setting both the foreground and background color at the same time? That's 3 strikes against it. I would love to see a better plugin used, but I won't let perfect be the enemy of good. If this is the best we can currently have, I can merge it.

tkapias:
The huge JSON file for the background is ugly, but you guessed right, the plugin use pattern for matching but an explicit table to translate them, so I had to list all combinations. 

Fortunately, it does not slow down vim's startup or the plugin's operations.

Also, I did not find another plugin that is async, for vim/neovim, and that lets you use custom patterns. And even if [vim-hexokinase](https://github.com/RRethy/vim-hexokinase) is archived I had no issues with it until now.

I will make a PR later today.

PR #838, opened 2024-05-30
[contrib] vim/neovim color patterns for theme's bg/fg

tkapias:
- Related issue: #835
- README.md: yes

## Highlighting color values in theme/preset files

Theme files for Liquid Prompt use the function `lp_terminal_format` documented in "docs/functions/public.rst".

The issue with this format is that this function takes 6 arguments (fg, bg, bold, underline, fallback_fg, fallback_bg), including 4 which are integers representing the ANSI escape code for colors, there is no way to highlight those values in any vanilla code/text editor.

Being a Vim user, I found out that the plugin [vim-hexokinase](https://github.com/RRethy/vim-hexokinase) does provide a way to match custom palettes of colors with patterns in the text.

**This contrib provides instructions and 2 JSON files containing custom values for vim-hexokinase, able to highlight the `fg` and `bg` arguments in a `lp_terminal_format` function.**

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -2064,13 +2064,21 @@ _lp_perl_env() {
     if [[ "$_LP_PERL_VENV_PROGRAM" = "perlbrew" ]] ; then
         local perlbrew_perl_ver="$(perlbrew use)"
 
+        if [[ "$perlbrew_perl_ver" == "Currently using system perl" ]]; then
+            return 1
+        fi
+
         # Example: Currently using perl-5.38.0
         __lp_escape "${perlbrew_perl_ver##*perl-}"
         lp_perl_env=$ret
     elif [[ "$_LP_PERL_VENV_PROGRAM" = "plenv" ]] ; then
-        local plenv_ver="$(plenv version)"
-        # Only first word is necessary as for rbenv
-        __lp_escape "${plenv_ver%%" (set"*}"
+        local plenv_ver="$(plenv version-name)"
+
+        if [[ "$plenv_ver" == "system" ]]; then
+            return 1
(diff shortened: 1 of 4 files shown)


tkapias:
It's probably more Github related, but I don't understand why Github does not preview the [README.md](https://github.com/tkapias/liquidprompt/tree/contrib/colors/vim-hexokinase/contrib/colors/vim-hexokinase). I copied the raw content to the comment field to be sure and the preview works, but in the Github file explorer it does not.

Can you confirm if it's the case for other users and if someone see the issue?

Rycieos (changes requested):
Sorry for the delay in reviewing.

I can confirm I also see it not rendered. I am not sure why, but I have these ideas:
1. There are some very long lines. Maybe GitHub's Markdown parser gives up after a certain line length? Try wrapping all lines at 80 chars.
2. Your code blocks contain type hints (eg \`\`\` bash), but I usually don't include a space between the backticks and the lang. Maybe try without?
3. Maybe GitHub tries to be "smart" and does some detection to see the file type? That would explain why it seems to highlight keywords, as the file is more bash/vim lines than Markdown lines. Maybe try putting the examples in code files and linking to them if you can't find a way to shorten them.

Otherwise this looks good. We need the Markdown to render though, as that is how most people will view it.

tkapias:
I closed this one and opened a new PR at #852.

I found why the Readme was not formatted by Github, it's just that the last code block was also the last test in the page and it ended with `" vi: ft=vim` to help identify the content syntax.
But, Github took it for the document itself. I added more text after and it's fixed.

Issue #716, opened 2022-06-30
Use `LP_TITLE_OPEN=$'\E]2'` for tmux 


kmARC:
The OSC sequence for setting tmux pane title is `\E]2` (see https://linux.die.net/man/1/tmux)

Currently liquidprompt sets this as `\Ek` (see [liquidprompt#L613](https://github.com/nojhan/liquidprompt/blob/6f0dc9f7dda99fe98c6dce26be5e461692782c6d/liquidprompt))

Shell: `5.1.16(1)-release`

Operating system: `Darwin Laptop056 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64 arm Darwin`

Liquidprompt version: `/opt/homebrew/Cellar/liquidprompt/2.1.2 `

### Expected Behavior
Liquidprompt sets tmux pane titles

### Current Behavior
It doesn't

### Possible Solution
Changing [liquidprompt#L613](https://github.com/nojhan/liquidprompt/blob/6f0dc9f7dda99fe98c6dce26be5e461692782c6d/liquidprompt) to 
```sh
        LP_TITLE_OPEN=$'\E]2'
```
fixes the issue

Rycieos:
Thanks for the report, documentation, and suggested solution.

Despite what the docs suggest, I am not able to reproduce this.

I need to have `set -g allow-rename on` in my Tmux config, but then the window title is set correctly with no changes to Liquidprompt. I always assumed Tmux had compatibility with Screen.

Changing `LP_TITLE_OPEN` to `$'\E]2'` as you suggest, or `$'\E]2;'` as I think the docs actually meant both make the window title not change.

Maybe you have a different configuration in your Tmux config? Or a different Tmux version? I tested on Tmux 2.7 and 3.3a.

kmARC:
Ohh... So I thought this supposedly renames the titles of the **tmux panes** not the **tmux windows**. 

Example with  `LP_TITLE_OPEN=$'\Ek'`:
<img width="1239" alt="Screenshot 2022-07-01 at 09 07 14" src="https://user-images.githubusercontent.com/6640417/176843404-e3b6f96d-5fe2-4c0e-baba-6fac26f163a5.png">

Example with  `LP_TITLE_OPEN=$'\E]2'` (no `;` at the end):
[...]

PR #839, opened 2024-06-18
Fix wrong LP_TITLE_OPEN for tmux

tianyuanhao:
Add a `;` after `$'\E]2'`, which doesn't work.

Refs: #716


Rycieos (approved):
https://github.com/tmux/tmux/blob/c07e856d244d07ab2b65e72328fb9fe20747794b/CHANGES#L3030
You appear to be correct. Thanks!

PR #840, opened 2024-06-18
Add editorconfig configuration

steelman:
The values are based on current state of files as well as existing Vim modeline values.

http://editorconfig.org/

# Technical checklist:

No shell code in this PR, the checklist does not apply.


Rycieos (approved):
Cool, thanks! I added the section for `*.rst` and a few other keys at the root myself. I removed the TODO for `*.txt` because there is only one file and it doesn't need formatting.

PR #841, opened 2024-06-21
Disk Space fixes

dex6:
Hi!
I ❤️ the disk space feature, simple thing but very useful. This PR tries to fix two small issues I've encountered with it. Both commits include unit tests updates to cover new cases.

#### 1. LP with disk space enabled breaks on pseudo filesystems like procfs and sysfs:

```
21:36:50 [mg@sirius:~] $ df -k -P /sys
Filesystem     1024-blocks  Used Available Capacity Mounted on
sysfs                    0     0         0        - /sys
21:36:55 [mg@sirius:~] $ cd /sys
bash: 100 -  : syntax error: operand expected (error token is "-  ")
21:36:55 [mg@sirius:~] $ pwd
/sys
# cd command worked; so expecting following prompt instead:
# 21:36:58 [mg@sirius:/sys] $
```
Note how Linux df reports `-` instead of `<value>%`, which breaks `lp_disk_perc=$(( 100 - ${capa:0:${#capa}-1} ))` calculation, cause the whole ${capa...} expands to empty value and triggers the "syntax error" as seen above.

At first, I've added explicit check for capa == "-", but then I've started my FreeBSD VM and saw it reports pseudo filesystems like normal ones (no dashes), but very small (total=1kiB devfs, or 8kiB procfs). Because of that, I've decided to check "total" column and reject all suspiciously small (<1 MiB) filesystems.

#### 2. I'm using following settings and I often work with relatively small (1-4GiB) filesystem images, which when mounted leads to not-so-useful mark in the prompt:

```
LP_ALWAYS_DISPLAY_VALUES=1
LP_DISPLAY_VALUES_AS_PERCENTS=1
LP_DISK_THRESHOLD=5000000   # ~5GB
LP_DISK_THRESHOLD_PERC=5

21:51:41 [mg@sirius:~] $ cd /mnt/image
21:51:45 🖴 88% [mg@sirius:/mnt/image] $ df -k -P .
Filesystem     1024-blocks   Used Available Capacity Mounted on
/dev/loop0         1977080 204824   1653472      12% /mnt/image
```
Note only 12% is used, but LP shows 88% free space mark which is way over the 5% I've set. This happens because the whole FS is smaller than my LP_DISK_THRESHOLD so it always matches the `lp_disk <= LP_DISK_THRESHOLD` condition.

To counter that, I've extended the condition to check LP_DISK_THRESHOLD only when FS is larger than it. However, I'm not sure whether this is the right thing to do. While it fixes my problem, maybe somebody with LP_DISPLAY_VALUES_AS_PERCENTS=0 would still like to have the value printed...

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1320,7 +1320,8 @@ Thresholds
    :value: 100000
 
    Display something if the available space on the hard drive hosting the
-   current directory goes below this absolute threshold *in kilobytes*.
+   current directory goes below this absolute threshold *in kilobytes*. If the
+   filesystem is smaller than :attr:`LP_DISK_THRESHOLD`, this check is skipped.
 
    The threshold for disk can also be set with :attr:`LP_DISK_THRESHOLD_PERC`,
    the first one to be reached triggering the display.
--- a/liquidprompt
+++ b/liquidprompt
@@ -4287,7 +4287,8 @@ _lp_disk() {
     [[ -z ${lp_disk} || -z ${lp_disk_perc} ]] && return 1
 
     # Comparison in KiB.
-    if (( lp_disk <= LP_DISK_THRESHOLD || lp_disk_perc <= LP_DISK_THRESHOLD_PERC )); then
+    # Ignore absolute threshold if whole FS is smaller than it; otherwise we would always show disk mark
+    if (( ( total >= LP_DISK_THRESHOLD && lp_disk <= LP_DISK_THRESHOLD ) || lp_disk_perc <= LP_DISK_THRESHOLD_PERC )); then
         # df outputs in KiB, but it's more robust to implement lp_human in bytes,
         # so we need to convert b in KiB here.
         local lp_bytes lp_bytes_units
--- a/liquidprompt
+++ b/liquidprompt
@@ -4261,20 +4261,25 @@ _lp_disk() {
     # Example output:
     # Filesystem     1024-blocks      Used Available Capacity Mounted on
     # /dev/nvme0n1p2   479608528 425188220  30011332      94% /
+    # For pseudo filesystems (like procfs or sysfs) output can be:
+    # Filesystem     1024-blocks  Used Available Capacity Mounted on
+    # sysfs                    0     0         0        - /sys
 
-    # Placeholder and capacity.
-    local _ capa
+    # Placeholder, total and capacity.
+    local _ total capa
     # Compute location of the targeted column on the safe line.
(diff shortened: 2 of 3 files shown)


Rycieos (changes requested):
Thanks for catching these and submitting a patch. These fixes look good.

I have only a few points:
1. I ran your updated tests against `liquidprompt` from `master`, and only the FreeBSD tests failed; the "Linux special FS" ones did not. This suggests the test coverage for that case is not working.
2. I think your logic of "only show if total is greater than absolute threshold" is a good idea, but it needs to be added to the docs for the `LP_DISK_THRESHOLD` config option. Something like:
> If the filesystem is smaller than ``LP_DISK_THRESHOLD``, this check is skipped.

dex6:
ad 2. Amended the second commit with documentation update as suggested.

ad 1. I've investigated this a bit and it looks like:
- in normal usage, whole subshell(?) executing `PROMPT_COMMAND` immediately exits with returncode 1 upon encountering syntax error in arithmetic expansion of `lp_disk_perc` (and does not continue remaining checks nor sets PS1), ultimately leading to the bug
- in unit tests, the problem causes subshell used by shunit assertFalse for executing `_lp_disk` to exit with returncode 1, which is indistinguishable from _lp_disk itself doing `return 1`. This is why the "Linux special FS" seem to "pass" on current master branch, especially because shunit does `2>/dev/null` and hides the syntax error produced by the subshell

The simplest solution would be to change return code from _lp_disk when detected pseudo FS from 1 to something else, but it seems like codes 0/1/2 are used consistently in LiquidPrompt's internals, so I think it's better to stay with 1.

Other possible solution is to instead:
```
assertFalse "Parsing of \"${names[i]}\" skipped" _lp_disk
```
call the function explicitly, avoiding subshell used by shunit:
```
_lp_disk
assertFalse "Parsing of \"${names[i]}\" skipped" "$?"
```
However (when executed on current master liquidprompt) this has a side effect that whole `test_disk` function crashes (with error) on "Linux Special FS" test case, and does not continue the loop checking next cases (FreeBSD, Cygwin). It's probably still better than not noticing the problem at all.

Please let me know what you think. If that's acceptable, I'll amend the first commit with such change.

dex6:
Here you are. Also changed test case order (added new ones at the end). This way it would crash after catching earlier failures.

Rycieos (approved):
Sorry for the delay, was on vacation last week.

Looks great, thanks for the fixes!

PR #846, opened 2024-09-10
Fix `systemd-nspawn` container detection

okulev:
When in `systemd-nspawn` container, `/proc/self/cgroup` is always present: https://systemd.io/CONTAINER_INTERFACE/ . So `elif [[ -r /run/host/container-manager ]]` branch cannot ever be reached.

The PR moves this `elif` inside the branch that processes `/proc/self/cgroup`.

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -4649,14 +4649,14 @@ _lp_container() {
             lp_container="Docker"
         elif [[ "$cgroup" == *"lxc"* ]]; then
             lp_container="LXC"
+        elif [[ -r /run/host/container-manager ]]; then
+            local ret
+            IFS='' read -r ret < /run/host/container-manager
+            __lp_escape "${ret#systemd-}"
+            lp_container="$ret"
         else
             return 1
         fi
-    elif [[ -r /run/host/container-manager ]]; then
-        local ret
-        IFS='' read -r ret < /run/host/container-manager
-        __lp_escape "${ret#systemd-}"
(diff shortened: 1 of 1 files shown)


Rycieos (commented):
This looks correct. Thanks for the docs link.

I am hesitant to merge, as I like to test changes manually. However, I am unable to get any of this to work. This is scary, as containers are supposed to be very reproduceable.

No mater what machine I run a Docker container on, `/proc/self/cgroup` always contains `0::/`. And `/run/host/container-manager` never exists.

If you are able to test this, could you share what steps work for you?

okulev:
This works on my Debian, and it should work on all Debian-based distros like Ubuntu too:

```sh
cat /proc/self/cgroup # Outside container and inside container the structure of cgroup is similar, as shown in `systemctl status`.

apt install debootstrap systemd-container
debootstrap --include=dbus,dbus-user-session,systemd-resolved stable /var/lib/machines/lp # This may take a while.
machinectl start lp
machinectl shell lp # Spawn shell in the just-started container.
cat /proc/self/cgroup
cat /run/host/container-manager

systemctl enable --now systemd-networkd # Hopefully network will work but I've seen cases where it doesn't.
apt install git
```

Rycieos:
Sorry for the delay, had to find time to sit down and hack this out.

Thanks for the repro steps, they worked perfectly. I have confirmed that your initial claim is correct.

PR #851, opened 2024-12-02
update to nerdfont v3.3.0

shadowwa:
Some glyphs had been updated with version 3.x of nerdfonts.
I've replaced the symbols using nerdfont cheatsheet based on name and glyph code.


Rycieos:
Looks great, thanks!

PR #856, opened 2024-12-29
feat(theme): explain

nojhan:
Adds a theme that displays the name of each of the prompt sections, along with a clickable hyperlink to the corresponding documentation.


Rycieos (changes requested):
I apologize for the delay. I hope to be more present now.

This is looking great. My biggest concern is the screenshots not matching the other themes.

PR #857, opened 2025-01-11
fix(_lp_strip_escape): remove Unicode empty character

nojhan:
Allows the user to remove default spaces in some parameters.

Remove any Unicode empty character
because it may be used as a placeholder
to overwrite a space character
used as default.
Passing a null string would trigger
the shell's default, so a solution
is to pass an empty character instead.
But then it would mess with functions counting
the characters, which often use
_lp_strip_escapes.

# Technical checklist:

    - [x] careful quoting
    - [ ] ran `tests.sh`


Rycieos:
This seems to me to be over-complicating the problem. If the problem is the inability to set certain config values to the empty string, then there is a much easier way to solve it.

From the [Bash docs](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html):
> When not performing substring expansion, using the form described below (e.g., ‘:-’), Bash tests for a parameter that is unset or null [empty string]. Omitting the colon results in a test only for a parameter that is unset. Put another way, if the colon is included, the operator tests for both *parameter*’s existence and that its value is not null; if the colon is omitted, the operator tests only for existence.

**All** of our config options are checked with `${LP_VAR:-"default"}` syntax, and so nothing is allowed to be set to the empty string (other than those that default to the empty string).

So, it seems to me the much easier solution is to change the assignment of those config variables from `:-` to `-`.

Not all would need changing. Any config option that controls how Liquid Prompt works needs to never be null. But config options that are meant to be used as strings to display to the user should be changed. So I think these variables would be appropriate for that treatment:
https://github.com/liquidprompt/liquidprompt/blob/b4e1d3481ec5c6072ae8e5985f6b73db89c61132/liquidprompt#L647-L749

Note that `LP_MARK_ENV_VAR_SEP` is not a valid config option. Is this an option from one of your themes?


Now, I imagine that you might still want the removal of the empty character. (By the way, which character are you using? `U+200B ZERO WIDTH SPACE`? That should probably be in the comment pointing out the empty character.) After all, you might need to use it in a theme for some reason, or maybe somehow it ends up in a data string, and then length calculations would be incorrect. However, I can think of a good reason not to remove it: there are many Unicode characters that are more than one "character" wide, but a string length check shows only one character. This empty character could be used to "fix" those calculations. That sounds dangerously hacky, but something to consider.

And anyway, it seems you are only removing one certain character, and there are multiple zero width characters in Unicode.

PR #858, opened 2025-03-25
Add support for Zsh named directories

olethanh:
Named directory in Zsh allow setting "alias" for directories.

They are set via `hash -d name=/path` and can be used in paths, eg. `cd ~name` or `touch ~name/file`.

In Zsh built-in prompt they are then displayed as `~name`  or `name/subdir`. This pull-request make liquidprompt displays it the same way.

Notes:
- Be careful not to include a terminal '/' when setting the path in the hash command, zsh don't supports it.
- Named directory can be unset via `hash -d=`

Feature is zsh only

Example prompt
```
[:~] % hash -d prj=~/Projects
[:~] % cd ~prj/liquidprompt 
[:~prj/liquidprompt] named-directory-support* ± cd ..
[:~prj] % cd ..
```

# Technical checklist:

    - [x] correct use of `IFS`
    - [x] careful quoting
    - [x] cautious array access
    - [x] ran `tests.sh`

 `docs/docs-lint.sh` returned an error for `https://dn-works.com/ufas/` but that is not related to my changes

```
(        overview: line  208) broken    https://dn-works.com/ufas/ - HTTPSConnectionPool(host='dn-works.com', port=443): Max retries exceeded with url: /ufas/ (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1000)')))
```

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1314,10 +1314,8 @@ __lp_floating_scale() {
   # Return $PWD with $HOME at the start replaced by "~".
   if (( _LP_SHELL_zsh )); then
       # Zsh-specific version, use zsh print -D builtin function which support named directories
-      zsh_minor_version=${ZSH_VERSION#*.}
-      zsh_minor_version=${zsh_minor_version%%.*}
       # Print -v was introduced in ZSH 5.3
-      if (( zsh_minor_version >= 3 )); then
+      if _lp_zsh_version_greatereq 5 3; then
           __lp_pwd_tilde() {
               print -v lp_pwd_tilde -n -r -D "${1:-$PWD}"
           }
@@ -1327,7 +1325,7 @@ __lp_floating_scale() {
           }
       fi
   else
-      # Bash-compatible fallback
+      # Bash version
       __lp_pwd_tilde() {
           # Needs to be in a variable, as different versions of Bash treat '~' in a
           # substitution differently
--- a/liquidprompt
+++ b/liquidprompt
@@ -1313,10 +1313,15 @@ __lp_floating_scale() {
 
 # Return $PWD with $HOME at the start replaced by "~".
 __lp_pwd_tilde() {  # [path]
-    # Needs to be in a variable, as different versions of Bash treat '~' in a
-    # substitution differently
-    local _path="${1:-$PWD}" tilde="~"
-    lp_pwd_tilde="${_path/#$HOME/$tilde}"
+    if (( _LP_SHELL_zsh )); then
+        # Use zsh print -D builtin function which support named directories
+        print -v lp_pwd_tilde -n -r -D "${1:-$PWD}"
+    else
+        # Needs to be in a variable, as different versions of Bash treat '~' in a
+        # substitution differently
+        local _path="${1:-$PWD}" tilde="~"
+        lp_pwd_tilde="${_path/#$HOME/$tilde}"
+    fi
 }
 
 # insert a space on the right
--- a/liquidprompt
+++ b/liquidprompt
@@ -1315,7 +1315,13 @@ __lp_floating_scale() {
 __lp_pwd_tilde() {  # [path]
     if (( _LP_SHELL_zsh )); then
         # Use zsh print -D builtin function which support named directories
-        print -v lp_pwd_tilde -n -r -D "${1:-$PWD}"
+        local minor_version=${ZSH_VERSION#*.}
+        local minor_version=${minor_version%%.*}
+        if (( minor_version >= 3 )); then
+            print -v lp_pwd_tilde -n -r -D "${1:-$PWD}"
+        else
+            lp_pwd_tilde=$(print -n -r -D "${1:-$PWD}")
+        fi
     else
         # Needs to be in a variable, as different versions of Bash treat '~' in a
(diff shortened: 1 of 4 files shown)


olethanh:
I fixed the module import issue caused by Python 3.13 and contacted the owner of the dn-works.com domain to warn them that their certificate is expired.

liquidprompt
```diff
@@ -1270,10 +1270,15 @@ __lp_floating_scale() {
 
 # Return $PWD with $HOME at the start replaced by "~".
 __lp_pwd_tilde() {  # [path]
-    # Needs to be in a variable, as different versions of Bash treat '~' in a
-    # substitution differently
-    local _path="${1:-$PWD}" tilde="~"
-    lp_pwd_tilde="${_path/#$HOME/$tilde}"
+    if (( _LP_SHELL_zsh )); then
+        # Use zsh print -D builtin function which support named directories
+        print -v lp_pwd_tilde -n -r -D "${1:-$PWD}"
```
Rycieos, line 1275: `print/printf -v <var>` wasn't supported in Zsh until 5.3. Currently we support Bash 3.2+ and Zsh 5.0.8+. Probably we should better document that. Maybe a note in CONTRIBUTING.md would be helpful, or we need to add these older shells to the CI tests. Now that I think about it, having that based on Docker images should be easy to do... I'll work on that.

But this does need to be solved. I get that `print -v` is much faster than `var=$()`, but we need to support these older shells. Probably the right thing to do would be to branch based on the Zsh version:

```suggestion
        # The following line is psudo-code; I have not checked what var needs to be checked.
        if [[ $ZSH_MINOR_VERSION -gt 2 ]]; then
            print -v lp_pwd_tilde -n -r -D "${1:-$PWD}"
        else
            lp_pwd_tilde=$(print -n -r -D "${1:-$PWD}")
        fi
```

Rycieos (changes requested):
Nice! This is an obvious feature that I am surprised no one has requested before now.

Thanks for fixing the docs testing job issue.

I only see one problem, but it is unfortunately a big one that must be solved before merging.

olethanh:
Thanks for the prompt reply, sorry I missed the notification for it.

I rebased the branch and made a  modification to check the version.

olethanh:
Kind of wondering if running the check everytime isn't going to negate the benefit performance of using `print -v` and if it would be more optimized to  check and define the function when loading the script. It would look something like

```bash
# Return $PWD with $HOME at the start replaced by "~".
if (( _LP_SHELL_zsh )); then
    # Zsh-specific version, use zsh print -D builtin function which support named directories
    zsh_minor_version=${ZSH_VERSION#*.}
    zsh_minor_version=${zsh_minor_version%%.*}
    # Print -v was introduced in ZSH 5.3
    if (( zsh_minor_version >= 3 )); then
        __lp_pwd_tilde() {
            print -v lp_pwd_tilde -n -r -D "${1:-$PWD}"
        }
    else
        __lp_pwd_tilde() {
            lp_pwd_tilde=$(print -n -r -D "${1:-$PWD}")
        }
    fi
else
    # Bash-compatible fallback
    __lp_pwd_tilde() {
        # Needs to be in a variable, as different versions of Bash treat '~' in a
        # substitution differently
        local _path="${1:-$PWD}" tilde="~"
        lp_pwd_tilde="${_path/#$HOME/$tilde}"
    }
fi
```

I have tested the change locally and can push it in the PR if you are open to it.

Rycieos:
> Kind of wondering if running the check everytime isn't going to negate the benefit performance of using `print -v` and if it would be more optimized to check and define the function when loading the script.

I agree, the code there is better.

My only problem with it is how you are checking the Zsh version, as the major version needs to be checked as well. I realized that there should be a better way, so I made one. Commit 5a009fe43f57dd83792c6bd5b7ce83f58d3de6fa has two new functions for checking the shell version. You can see the docs here: https://liquidprompt.readthedocs.io/en/latest/functions/util.html#lp_zsh_version_greatereq

That would look something like this:
```bash
if (( _LP_SHELL_zsh )); then
    # Zsh-specific version, use zsh print -D builtin function which support named directories
    # Print -v was introduced in ZSH 5.3
    if _lp_zsh_version_greatereq 5 3; then
        __lp_pwd_tilde() {
            ...
```

So if you rebase you should be able to do that. I also fixed the docs issues that you had; hopefully rebasing and dropping your changes there will be easy.

olethanh:
I rebased, add the modifications to define at load and use your function to check the version.
I had pondered if I should test the major version too, you are right it is more future proof.   Zsh provide a is-at-least function to test the version but it is not bash compatible https://github.com/zsh-users/zsh/blob/master/Functions/Misc/is-at-least.

Rycieos (approved):
> FYI Zsh provide a is-at-least function to test the version but it is not bash compatible https://github.com/zsh-users/zsh/blob/master/Functions/Misc/is-at-least.

Of course Zsh has a function for that, I should have guessed that. Ah well, at least now LP has a unified interface for checking shell versions.

Thanks for fixing this up, this looks great.

PR #867, opened 2026-07-28
Feat: option to truncate from FQDN

eric-eisenhart:
Adds `LP_HOSTNAME_FQDN_TRIM` setting that can remove end of FQDN based on domain segments or a static string.

IOW, turn `foo.bar.example.net` into `foo.bar` with `LP_HOSTNAME_FQDN_TRIM=2` or `LP_HOSTNAME_FQDN_TRIM=example.net`.

At my work we set hostnames to things like `svc.site.example.net` or `b.svc.site.example.net`. The FQDN is unnecessarily long in a prompt, but just `svc` or `b` is not enough information. This is a way to make that usable for me, and I figured it's reasonably likely to be useful for others.

# Technical checklist:

    - [x] correct use of `IFS`
    - [x] careful quoting
    - [x] cautious array access
    - [x] ran `tests.sh`
    - [x] ran `shellcheck.sh` (requires [shellcheck](https://github.com/koalaman/shellcheck#user-content-installing))

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -608,6 +608,7 @@ __lp_source_config() {
     LP_PATH_VCS_ROOT=${LP_PATH_VCS_ROOT:-1}
     LP_HOSTNAME_ALWAYS=${LP_HOSTNAME_ALWAYS:-0}
     LP_HOSTNAME_METHOD=${LP_HOSTNAME_METHOD:-short}
+    LP_HOSTNAME_TRIM=${LP_HOSTNAME_TRIM:-2}
     LP_USER_ALWAYS=${LP_USER_ALWAYS:-1}
     LP_ALWAYS_DISPLAY_VALUES=${LP_ALWAYS_DISPLAY_VALUES:-1}
     LP_DISPLAY_VALUES_AS_PERCENTS=${LP_DISPLAY_VALUES_AS_PERCENTS:-0}
@@ -2731,7 +2732,7 @@ _lp_hostname() {
             return 1
         fi
 
-        if [[ $LP_HOSTNAME_METHOD == fqdn ]]; then
+        if [[ $LP_HOSTNAME_METHOD == fqdn || $LP_HOSTNAME_METHOD == fqdn_trimmed ]]; then
             lp_hostname=$(hostname -f 2>/dev/null)
         elif [[ $LP_HOSTNAME_METHOD == pretty ]]; then
             if [[ $LP_OS == Darwin ]]; then
@@ -2748,6 +2749,12 @@ _lp_hostname() {
         # Truncate to the first subdomain
         if [[ $LP_HOSTNAME_METHOD == short ]]; then
             lp_hostname=${lp_hostname%%.*}
+        # Truncate domain sections from the end
+        elif [[ $LP_HOSTNAME_METHOD == fqdn_trimmed ]]; then
+            local i
+            for (( i=0; i < LP_HOSTNAME_TRIM; i++ )); do
+                lp_hostname=${lp_hostname%.*}
+            done
         fi
 
         local ret
--- a/liquidprompt
+++ b/liquidprompt
@@ -2752,7 +2752,17 @@ _lp_hostname() {
         fi
 
         # Truncate domain sections from the end
-        if [[ $LP_HOSTNAME_TRIM == *[!0-9]* ]]; then
+        if [[ $LP_HOSTNAME_TRIM == -* ]]; then
+            # Negative: keep that many sections from the start instead
+            local i=$LP_HOSTNAME_TRIM domain=$lp_hostname
+            while (( i < 0 )) && [[ $domain == *.* ]]; do
+                domain=${domain#*.}
+                (( i++ ))
+            done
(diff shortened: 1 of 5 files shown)


Rycieos (changes requested):
This is great, definitely a useful feature. Nice work on the tests too. 

Only problem (and it's a big one) is I don't think it should be limited to the `fqdn` method. It makes sense that users might want to shorten the `full` method as well. I suppose a user might be stuck with a pretty name they can't change and might want to shorten it too, so why not keep it simple. 

Actually, one other thing. This should support negative numbers, which instead keeps sections from the left. So for example, `-1` with `foo.bar.example.com` would return `foo`. And `-3` would return `foo.bar.example`. That's what the `short` method does right now. 

Currently `$LP_HOSTNAME_METHOD` has 4 valid options: `full`, `short`, `fqdn`, and `pretty`. `short` is effectively `LP_HOSTNAME_FQDN_TRIM=-1`. This means we could simply deprecate it. But that leaves `full` as a very weirdly named option.

So here is what I'm thinking:
* Rename `LP_HOSTNAME_FQDN_TRIM` to `LP_HOSTNAME_TRIM`.
* Make `LP_HOSTNAME_TRIM` apply to all methods.
* Make `LP_HOSTNAME_METHOD=short` be an alias for `LP_HOSTNAME_METHOD=full` and `LP_HOSTNAME_TRIM=-1`.
  * Update the docs for `short` to mention this.

Feel free to push back if you think differently.

eric-eisenhart:
> This is great, definitely a useful feature. Nice work on the tests too.

Thanks!

> Only problem (and it's a big one) is I don't think it should be limited to the `fqdn` method.

That definitely makes sense to me and seems like should actually simplify the change.

> Actually, one other thing. This should support negative numbers, which instead keeps sections from the left. 

I hadn't thought of those use cases, but makes sense to me that it would match what some want.

> But that leaves `full` as a very weirdly named option.

FWIW, I don't have any systems where `full` behaves any differently than `short`, because there's really no way to set "domain name". (Otherwise it _sounds_ like what I would want that led me to implement this feature)

> So here is what I'm thinking:
> 
>     * Rename `LP_HOSTNAME_FQDN_TRIM` to `LP_HOSTNAME_TRIM`.
> 
>     * Make `LP_HOSTNAME_TRIM` apply to all methods.
> 
>     * Make `LP_HOSTNAME_METHOD=short` be an alias for `LP_HOSTNAME_METHOD=full` and `LP_HOSTNAME_TRIM=-1`.
>       
>       * Update the docs for `short` to mention this.

Should `short` use `full` or `fqdn`? At the least it'd make it easier to understand docs because we all know what an fqdn is.

Rycieos:
> FWIW, I don't have any systems where `full` behaves any differently than `short`, because there's really no way to set "domain name". (Otherwise it _sounds_ like what I would want that led me to implement this feature)

You can use `hostnamectl` to set your hostname to any string, like for example `foo.example.com`.

> Should `short` use `full` or `fqdn`? At the least it'd make it easier to understand docs because we all know what an fqdn is.

It currently uses the same source as `full`, so it should stay that way.

eric-eisenhart:
> > FWIW, I don't have any systems where `full` behaves any differently than `short`, because there's really no way to set "domain name". (Otherwise it _sounds_ like what I would want that led me to implement this feature)
> 
> You can use `hostnamectl` to set your hostname to any string, like for example `foo.example.com`.

Yeah, but doesn't really work for a `foo.bar` type thing the way I'd want... Our servers still really need their actual FQDN in there.

```
$ sudo hostnamectl hostname foo.bar.example.net
$ sudo hostnamectl hostname --pretty foo.bar

$ cat /etc/hostname
foo.bar.example.net

$ cat /etc/machine-info
PRETTY_HOSTNAME=foo.bar

$ hostnamectl
 Static hostname: foo.bar.example.net
 Pretty hostname: foo.bar

$ hostname --fqdn
foo.bar.example.com

$ hostname --short
foo

$ hostname --domain
bar.example.com
```

eric-eisenhart:
Anyway, I think I got the changes done you wanted, so ready for re-review.

Code changes: actually pretty small.

liquidprompt
```diff
@@ -2745,9 +2752,25 @@ _lp_hostname() {
             lp_hostname=${HOSTNAME:-${HOST-}}
         fi
 
-        # Truncate to the first subdomain
-        if [[ $LP_HOSTNAME_METHOD == short ]]; then
-            lp_hostname=${lp_hostname%%.*}
+        # Truncate domain sections from the end
+        if [[ $LP_HOSTNAME_TRIM == -* ]]; then
+            # Negative: keep that many sections from the start instead
+            local i=$LP_HOSTNAME_TRIM domain=$lp_hostname
```
Rycieos: These must be quoted due to bug in Zsh 5.0 with `local`.

```suggestion
            local i="$LP_HOSTNAME_TRIM" domain="$lp_hostname"
```

Rycieos (changes requested):
Small issue, the rest looks great.

Rycieos (approved):
Nice, thanks!

PR #870, opened 2026-08-16
Exclude current session from attached multiplexer sessions count

lestenholm:
## Description

When running inside an active multiplexer session (`tmux`, `screen`, `shpool`, or `herdr`), `_lp_multiplexer()` already indicates the active multiplexer context in the prompt. Displaying `1r` for the session currently occupied is redundant.

This PR makes `_lp_attached_sessions()` automatically subtract the current session context when `_lp_multiplexer` is active:
- **Outside multiplexer:** Displays all attached sessions on the host (`1r`, `2r`, etc.).
- **Inside multiplexer with 1 attached session (itself):** Hidden (`0r`).
- **Inside multiplexer with 2 attached sessions (itself + 1 other):** Displays `1r`.

# Technical checklist:

    - [x] correct use of `IFS`
    - [x] careful quoting
    - [x] cautious array access
    - [x] unit tests have been updated (`tests/test_attached_sessions.sh`)
    - [x] ran `tests.sh`
    - [x] ran `shellcheck.sh`
    - [x] tag `.. versionadded:: 2.4`
    - [x] ran `docs/docs-lint.sh` (0 errors)

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -451,6 +451,20 @@ Features
 
    .. versionadded:: 2.1
 
+.. attribute:: LP_ATTACHED_SESSIONS_EXCLUDE_CURRENT
+   :type: bool
+   :value: 0
+
+   Exclude the current session from the attached multiplexer session count.
+
+   When enabled (``1``) and running inside an active multiplexer session, the
+   current session is subtracted from the attached session count so that only
+   other attached sessions on the host are counted.
+
+   See also: :attr:`LP_ENABLE_ATTACHED_SESSIONS`.
+
+   .. versionadded:: 2.4
+
 .. attribute:: LP_ENABLE_ATTACHED_SESSIONS
    :type: bool
    :value: 0
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -451,25 +451,14 @@ Features
 
    .. versionadded:: 2.1
 
-.. attribute:: LP_ATTACHED_SESSIONS_EXCLUDE_CURRENT
-   :type: bool
-   :value: 0
-
-   Exclude the current session from the attached multiplexer session count.
-
-   When enabled (``1``) and running inside an active multiplexer session, the
-   current session is subtracted from the attached session count so that only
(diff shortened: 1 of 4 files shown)


Rycieos:
Yeah, this is obviously a good idea. I have two questions:
1. If this the correct functionality? As proposed, the method is to subtract one from the number always. I'm thinking I would want to see the number as is, unless it is `1`, then hide it. Maybe that should be a config option? Maybe I'm overthinking it...
2. Why is a new configuration option needed? My thought is that if `LP_ENABLE_MULTIPLEXER` is enabled, you want to see a mark if you are in a multiplexer. If not, then you don't. Like you said, it's redundant to display it twice, so I think if `LP_ENABLE_MULTIPLEXER` is enabled this should always happen, no need for a config option. Just mention this functionality clearly in the docs.

lestenholm:
> Yeah, this is obviously a good idea. I have two questions:
> 
>     1. If this the correct functionality? As proposed, the method is to subtract one from the number always. I'm thinking I would want to see the number as is, unless it is `1`, then hide it. Maybe that should be a config option? Maybe I'm overthinking it...

So my thought for subtracting 1 (count--) was that the primary use case for `r` is when logging into a machine from the outside (e.g. over SSH);  a reminder that an active session exists on the system. Once inside a multiplexer session, the user is already interacting with it (and liquidprompt highlights the prompt brackets). The attached count should therefore IMO only reflect additional / other sessions on the host. :)

Subtracting 1 also means if there is only 1 session (the current one), nothing is shown; if there are 2, `1r`  indicates 1 other session is running elsewhere on the host.

But maybe **I** am overthinking this?

>     2. Why is a new configuration option needed? 

If the above functionality makes sense, I can make that the actual and documented behavior? (That's what I would personally prefer. :)

Rycieos:
1. Fair enough, ok, keep it as is.
2. Yeah, I think it makes sense. Drop the config option.

Rycieos (approved):
Looks great, thanks!

PR #871, opened 2026-08-24
Add Zellij as detected terminal multiplexer

lestenholm:
## Description

This PR adds support for **Zellij** (https://zellij.dev) as a detected terminal multiplexer, alongside `screen`, `tmux`, `shpool`, and `herdr`.

### Changes Included:
1. **Environment Detection:** Added Zellij detection to `_lp_multiplexer()` via `$ZELLIJ` and `$ZELLIJ_SESSION_NAME`.
2. **Session Counting:** Added Zellij parsing to `_lp_detached_sessions()` and `_lp_attached_sessions()` using `zellij list-sessions -n`.
3. **Session Exclusion:** Excludes current session context by default when running inside Zellij.
4. **Init Guard Fix:** Fixed guard in `__lp_post_source_config()` to check `(( LP_ENABLE_DETACHED_SESSIONS || LP_ENABLE_ATTACHED_SESSIONS ))` so attached session tracking works even if detached tracking is disabled.
5. **Documentation & Tests:** Updated all docs, wordlists, and unit test suites.

# Technical checklist:

    - [x] correct use of `IFS`
    - [x] careful quoting
    - [x] cautious array access
    - [x] unit tests have been updated (`tests/test_detached_sessions.sh`, `tests/test_attached_sessions.sh`)
    - [x] ran `tests.sh`
    - [x] ran `shellcheck.sh`
    - [x] tag `.. versionadded:: 2.4` / `.. versionchanged:: 2.4`
    - [x] ran `docs/docs-lint.sh`

### DIFF
--- a/liquidprompt
+++ b/liquidprompt
@@ -1083,11 +1083,12 @@ lp_activate() {
         fi
     fi
 
-    if (( LP_ENABLE_DETACHED_SESSIONS )); then
+    if (( LP_ENABLE_DETACHED_SESSIONS || LP_ENABLE_ATTACHED_SESSIONS )); then
         command -v screen >/dev/null ; _LP_ENABLE_SCREEN=$(( ! $? ))
         command -v tmux >/dev/null   ; _LP_ENABLE_TMUX=$(( ! $? ))
         command -v shpool >/dev/null ; _LP_ENABLE_SHPOOL=$(( ! $? ))
         command -v herdr >/dev/null  ; _LP_ENABLE_HERDR=$(( ! $? ))
+        command -v zellij >/dev/null ; _LP_ENABLE_ZELLIJ=$(( ! $? ))
     fi
 
     # Use standard path symbols inside Midnight Commander
@@ -2205,6 +2206,9 @@ _lp_multiplexer() { # [--internal]
     elif [[ -n ${HERDR_SESSION-} || -n ${HERDR_SESSION_NAME-} || -n ${HERDR_ENV-} ]]; then
         lp_multiplexer=herdr
         return 0
+    elif [[ -n ${ZELLIJ-} || -n ${ZELLIJ_SESSION_NAME-} ]]; then
+        lp_multiplexer=zellij
+        return 0
     fi
     return 1
 }
@@ -2881,7 +2885,7 @@ _lp_shell_level_color() {
 # Related jobs #
 ################
 
-# Return the count of detached screen, tmux, shpool, and herdr sessions.
+# Return the count of detached screen, tmux, shpool, herdr, and zellij sessions.
 _lp_detached_sessions() {
     (( LP_ENABLE_DETACHED_SESSIONS )) || return 2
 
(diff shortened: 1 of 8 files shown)


liquidprompt
```diff
@@ -2890,12 +2894,13 @@ _lp_detached_sessions() {
     (( _LP_ENABLE_TMUX )) && count+=$(tmux list-sessions 2> /dev/null | GREP_OPTIONS='' \grep -cv 'attached')
     (( _LP_ENABLE_SHPOOL )) && count+=$(shpool list 2> /dev/null | GREP_OPTIONS='' \grep -c 'disconnected')
     (( _LP_ENABLE_HERDR )) && count+=$(herdr session list 2> /dev/null | GREP_OPTIONS='' \grep -c '[[:space:]]detached\([[:space:]]\|$\)')
+    (( _LP_ENABLE_ZELLIJ )) && count+=$(zellij list-sessions -n 2> /dev/null | GREP_OPTIONS='' \grep -v '(EXITED' | GREP_OPTIONS='' \grep -v '(current)' | GREP_OPTIONS='' \grep -c '[^[:space:]]')
```
Rycieos: Is there no way to make this one `grep` command instead of 3? Usage of `-v` makes it hard, but I think this should be the same:

```suggestion
    (( _LP_ENABLE_ZELLIJ )) && count+=$(zellij list-sessions -n 2> /dev/null | GREP_OPTIONS='' \grep -vce '(EXITED' -e '(current)' -e '^[[:space:]]*$')
```

liquidprompt
```diff
@@ -2904,6 +2909,7 @@ _lp_attached_sessions() {
     (( _LP_ENABLE_TMUX )) && count+=$(tmux list-sessions 2> /dev/null | GREP_OPTIONS='' \grep -c '(attached)$')
     (( _LP_ENABLE_SHPOOL )) && count+=$(shpool list 2> /dev/null | GREP_OPTIONS='' \grep -c '[[:space:]]attached$')
     (( _LP_ENABLE_HERDR )) && count+=$(herdr session list 2> /dev/null | GREP_OPTIONS='' \grep -c '[[:space:]]\(running\|attached\)\([[:space:]]\|$\)')
+    (( _LP_ENABLE_ZELLIJ )) && count+=$(zellij list-sessions -n 2> /dev/null | GREP_OPTIONS='' \grep -c '(current)')
```
Rycieos, line 2912: What does `-n` do? I don't see it in their online documentation.

tests/test_attached_sessions.sh
```diff
@@ -214,42 +270,53 @@ s2	attached
...
+    printf '%s' "z1 (current)
+z2 (current)
 "
   }
 
   # 1. Inside tmux (subtract 1)
   TMUX=1
   _lp_attached_sessions
-  assertEquals "Exclude current session inside tmux" "7" "$lp_attached_sessions"
+  assertEquals "Exclude current session inside tmux" "9" "$lp_attached_sessions"
```
Rycieos: Let's refactor this constant out to a global near the top of this function so every added platform doesn't need to change every single line here.

tests/test_attached_sessions.sh
```diff
@@ -214,42 +270,53 @@ s2	attached
...
+  assertEquals "Exclude current session inside zellij" "9" "$lp_attached_sessions"
 
-  # 6. Inside tmux but LP_ENABLE_MULTIPLEXER=0 (multiplexer detection disabled, do not subtract)
-  TMUX=1
+  # 6. Outside any multiplexer (do not subtract)
+  unset ZELLIJ ZELLIJ_SESSION_NAME
+  _lp_attached_sessions
+  assertEquals "Do not exclude current session when outside multiplexer" "10" "$lp_attached_sessions"
+
+  # 7. Inside zellij but LP_ENABLE_MULTIPLEXER=0 (multiplexer detection disabled, do not subtract)
```
Rycieos: Drop these numbers in the comments to reduce git diff when adding new platforms.

Rycieos (changes requested):
Mostly fine.

I have to ask though; are you actually using multiple multiplexers, or are you just adding support to Liquid Prompt while testing them out?

liquidprompt
lestenholm: Ok, applied the single `grep -vce` suggestion.

liquidprompt
lestenholm: `-n` is the short flag for `--no-formatting` in `zellij list-sessions` (listed in `zellij list-sessions --help`). It strips ANSI terminal color escape codes (`\e[32;1m`) from session names and status tags so text parsing isn't polluted by escape sequences. It works without, it's just an old habit.

tests/test_attached_sessions.sh
lestenholm: Done! Refactored the expected counts to `total_sessions` and `inside_sessions` variables at the top of the function and dropped the numbers from the section comments to keep future diffs minimal.

lestenholm:
Thanks for the review!

To answer your question: it all started with me evaluating `herdr` and testing `zellij`. Since I had just added `herdr` support to Liquid Prompt, I figured I might as well add `zellij` while I was in the multiplexer code. :) (I'm only using `herdr` at the moment.)

I've updated the PR with your suggestions!

Rycieos (approved):
Nice, thanks!

PR #873, opened 2026-08-30
Add LP_ENABLE_VCS_RESOLVE_SYMLINKS config option

lestenholm:
# Add `LP_ENABLE_VCS_RESOLVE_SYMLINKS` option to detect VCS repositories in symlinked subdirectories

## Summary

When navigating into a subdirectory of a VCS repository via a symbolic link (for example, `~/frontend -> /projects/my-app/src/components`), Liquid Prompt would fail to detect the repository because the upward directory traversal in `_lp_find_vcs()` only climbed the logical parent paths in `$PWD` (`/home/user/frontend` -> `/home/user` -> `/home` -> `/`).

This PR adds the configuration option **`LP_ENABLE_VCS_RESOLVE_SYMLINKS`** (default: `0` / disabled for backwards compatibility and zero performance regression) to optionally search the physical directory hierarchy when the logical search finds no repository.

---

## Key Design Principles

1. **Fast-path first (Zero overhead):**
   - The primary check always runs against logical `$PWD`.
   - If `LP_ENABLE_VCS_RESOLVE_SYMLINKS=0` (default) or when `$PWD` is already inside a VCS repository, symlink resolution is never invoked and no subshells or extra directory climbs occur.
2. **Shell-native path resolution (0 subshells on modern shells):**
   - In **Zsh**: Uses parameter expansion `${PWD:A}` to resolve the canonical physical path in memory (**0 forks / 0 subshells**).
   - In **Bash 5.3+**: Uses non-subshell in-process command substitution `${ pwd -P ; }` (**0 forks / 0 subshells**).
   - In **Bash 3.2 – 5.2**: Safely falls back to the shell builtin `$(pwd -P 2>/dev/null)`.
3. **Clean, bounded resolution structure:**
   - `_lp_find_vcs([directory])` accepts an optional directory argument and delegates to a strictly bounded 1-level recursive fallback (`_lp_find_vcs "$lp_physical_pwd"`), keeping the primary directory climbing loop untouched and un-nested.
4. **Multi-VCS compatibility:**
   - Implemented within the generic `_lp_find_vcs()` traversal loop, automatically benefiting Git, Subversion, Mercurial, Bazaar, and Fossil.
5. **Integration with `LP_DISABLED_VCS_PATHS`:**
   - Evaluates both the logical symlink path and the physical repository target, ensuring excluded paths return exit code `2` (`lp_vcs_type="disabled"`).
6. **Display context preserved:**
   - The prompt directory segment retains the logical path (`~/frontend`) while the VCS segment accurately displays branch, dirty status, ahead/behind counts, stashes, and marks from the physical repository.

---

## Technical Checklist

- [x] code follows our [shell standards](https://github.com/liquidprompt/liquidprompt/wiki/Shell-standards):
    - [x] correct use of `IFS`
    - [x] careful quoting
    - [x] cautious array access (`${_LP_ENABLED_VCSS[@]+"${_LP_ENABLED_VCSS[@]}"}`, `${LP_DISABLED_VCS_PATHS[@]+"${LP_DISABLED_VCS_PATHS[@]}"}`)
    - [x] subshells are avoided as much as possible (0 subshells on Zsh and Bash 5.3+)
    - [x] string substitutions handled per shell (`_LP_SHELL_*`)
- [x] tests and checks have been added, run, and their warnings fixed:
    - [x] unit tests updated in `tests/test_git.sh` (`test_git_symlinks`)
    - [x] ran `tests.sh` under Bash and Zsh
    - [x] tested in isolated Fedora 44 container environment with Git, SVN, and Mercurial
- [x] documentation has been updated accordingly:
    - [x] `LP_ENABLE_VCS_RESOLVE_SYMLINKS` documented in alphabetical order in `docs/config.rst`
    - [x] tags `.. versionadded:: 2.4` and `.. versionchanged:: 2.4` added
    - [x] verified with `tools/config-from-doc.sh`
    - [x] verified with Sphinx build & spellcheck (`docs/docs-lint.sh`)
    - [x] entry added to `CHANGELOG.md`

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1204,6 +1204,19 @@ Features
 
    .. versionadded:: 2.2
 
+.. attribute:: LP_ENABLE_VCS_RESOLVE_SYMLINKS
+   :type: bool
+   :value: 0
+
+   Resolve symbolic links when searching for VCS repositories.
+
+   If enabled, Liquid Prompt will search the physical directory hierarchy
+   if no VCS repository is found in the logical path hierarchy. This allows
+   VCS information (such as Git branch, dirty status, stashes) to be displayed
+   when navigating into symlinked subdirectories within a repository.
+
+   .. versionadded:: 2.3
+
 .. attribute:: LP_ENABLE_VCS_ROOT
    :type: bool
    :value: 0
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1213,9 +1213,9 @@ Features
    If enabled, Liquid Prompt will search the physical directory hierarchy
    if no VCS repository is found in the logical path hierarchy. This allows
    VCS information (such as Git branch, dirty status, stashes) to be displayed
-   when navigating into symlinked subdirectories within a repository.
+   when navigating into a symbolic link to a repository subdirectory.
 
-   .. versionadded:: 2.3
+   .. versionadded:: 2.4
 
 .. attribute:: LP_ENABLE_VCS_ROOT
    :type: bool
--- a/docs/functions/data/vcs.rst
+++ b/docs/functions/data/vcs.rst
@@ -65,6 +65,9 @@ this.
    .. versionchanged:: 2.2
       Added the *lp_vcs_specific_dir* return value.
 
+   .. versionchanged:: 2.3
+      Added support for resolving symbolic links via :attr:`LP_ENABLE_VCS_RESOLVE_SYMLINKS`.
+
 .. function:: _lp_are_vcs_enabled()
 
    Returns ``true`` if the current directory is not excluded by the config
--- a/docs/functions/data/vcs.rst
+++ b/docs/functions/data/vcs.rst
@@ -15,7 +15,7 @@ a theme can provide a common look for all VCS types.
 See the default theme function :func:`_lp_vcs_details_color` for an example of
 this.
 
-.. function:: _lp_find_vcs() -> var:lp_vcs_type, var:lp_vcs_root, \
+.. function:: _lp_find_vcs([directory]) -> var:lp_vcs_type, var:lp_vcs_root, \
    var:lp_vcs_dir, var:lp_vcs_specific_dir, var:lp_vcs_subtype
 
    Returns ``true`` if the current directory is part of a version control
@@ -65,10 +65,11 @@ this.
    .. versionchanged:: 2.2
       Added the *lp_vcs_specific_dir* return value.
 
-   .. versionchanged:: 2.3
-      Added support for resolving symbolic links via :attr:`LP_ENABLE_VCS_RESOLVE_SYMLINKS`.
+   .. versionchanged:: 2.4
+      Added support for resolving symbolic links via
+      :attr:`LP_ENABLE_VCS_RESOLVE_SYMLINKS`.
 
-.. function:: _lp_are_vcs_enabled()
+.. function:: _lp_are_vcs_enabled([directory])
 
    Returns ``true`` if the current directory is not excluded by the config
    option :attr:`LP_DISABLED_VCS_PATHS`.
--- a/liquidprompt
+++ b/liquidprompt
@@ -1352,9 +1352,13 @@ __lp_floating_scale() {
           lp_pwd_tilde="${_path/#$HOME/$tilde}"
       }
 
-      __lp_physical_pwd() {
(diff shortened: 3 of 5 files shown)


docs/functions/data/vcs.rst
```diff
@@ -65,6 +65,9 @@ this.
    .. versionchanged:: 2.2
       Added the *lp_vcs_specific_dir* return value.
 
+   .. versionchanged:: 2.3
+      Added support for resolving symbolic links via :attr:`LP_ENABLE_VCS_RESOLVE_SYMLINKS`.
```
Rycieos: > Line too long (92/80)

docs/functions/data/vcs.rst
```diff
@@ -65,6 +65,9 @@ this.
    .. versionchanged:: 2.2
       Added the *lp_vcs_specific_dir* return value.
 
+   .. versionchanged:: 2.3
```
Rycieos:
```suggestion
   .. versionchanged:: 2.4
```

docs/config.rst
```diff
@@ -1204,6 +1204,19 @@ Features
...
+   :value: 0
+
+   Resolve symbolic links when searching for VCS repositories.
+
+   If enabled, Liquid Prompt will search the physical directory hierarchy
+   if no VCS repository is found in the logical path hierarchy. This allows
+   VCS information (such as Git branch, dirty status, stashes) to be displayed
+   when navigating into symlinked subdirectories within a repository.
+
+   .. versionadded:: 2.3
```
Rycieos:
```suggestion
   .. versionadded:: 2.4
```

liquidprompt
```diff
@@ -2997,30 +3006,51 @@ _lp_find_vcs() {
...
+            if [[ -n "$lp_physical_pwd" && "$lp_physical_pwd" != "$search_path" ]]; then
+                if ! _lp_are_vcs_enabled "$lp_physical_pwd"; then
+                    lp_vcs_type="disabled"
+                    lp_vcs_root=""
+                    return 2
+                fi
+                search_path="$lp_physical_pwd"
+                checked_physical=1
+            else
+                break
```
Rycieos: This whole thing feels backwards to me. Rather than put this thing in a `while true` loop, add a block down here like you did that recursively calls this function:

```bash
# Add the $1 check to prevent double checking for symlinks.
if [[ -z "$lp_vcs_type" && -z "${1-}" ]] && (( LP_ENABLE_VCS_RESOLVE_SYMLINKS )); then
    local lp_physical_pwd
    __lp_physical_pwd
    if [[ -n "$lp_physical_pwd" && "$lp_physical_pwd" != "$search_path" ]]; then
        _lp_find_vcs "$lp_physical_pwd"
        return "$?"
    fi
fi
```

That means both `_lp_find_vcs()` and `_lp_are_vcs_enabled()` need to support a custom directory argument, and the first line of `_lp_find_vcs()` will pass that argument to `_lp_are_vcs_enabled()`.

Rycieos (changes requested):
Yeah this is a good idea. Too bad on Bash it requires a subshell, otherwise I would want this enabled by default.

liquidprompt
lestenholm: Great feedback, silly mistakes from me again. Anyway,  I've updated the PR with all requested changes. 👍 

## Fixes
- **Refactored `_lp_find_vcs([directory])`**: Added the optional directory argument and switched to the 1-level recursive fallback block as suggested, keeping the main climbing loop untouched.
- **bash subshell**:  Added a dynamic optimization to `__lp_physical_pwd`: for **bash 5.3+**, it uses non-subshell in-process command substitution (`${ pwd -P ; }`) while safely falling back to `$(pwd -P)` on older Bash releases.

Since recent shells are now with zero subshell overhead, let me know if you would like `LP_ENABLE_VCS_RESOLVE_SYMLINKS` enabled by default? :)

## Docs
- **Documentation**: Updated tags to `.. versionadded:: 2.4` and `.. versionchanged:: 2.4`, documented the optional `[directory]` parameter in function signatures, fixed the line wrapping (< 80 chars), and rephrased the description to pass the Sphinx spelling check.
- **Changelog**: Added the `([#873])` link to `CHANGELOG.md`.

Rycieos (approved):
Nice, thanks!

